PR #4, opened 2017-08-16
change permission for pem, conf, and fact files

SteveHNH:
Remove the executable permission from the following files:
* /etc/insights-client.conf
* /etc/*.pem
* /etc/insights-client/rpm.egg.asc
* /etc/ansible/facts.d/insights.fact
* /etc/ansible/facts.d/insights_machine_id.fact


kylape:
But I'd be happy to merge this now and throw my suggestions into a trello card.

SteveHNH:
@kylape I can lock it down further. Let me take another swing at it

PR #5, opened 2017-08-29
check for insights group and root user

brantleyr:


insights_client/__init__.py
```diff
@@ -30,13 +31,16 @@
 try:
     insights_uid = pwd.getpwnam("insights").pw_uid
     insights_gid = pwd.getpwnam("insights").pw_gid
+    insights_grpid = grp.getgrpname("insights").gr_gid
+    insights_grpusers = grp.getgrpname("insights").gr_mem
```
kylape: Is this used anywhere?

insights_client/__init__.py
```diff
@@ -30,13 +31,16 @@
 try:
     insights_uid = pwd.getpwnam("insights").pw_uid
     insights_gid = pwd.getpwnam("insights").pw_gid
+    insights_grpid = grp.getgrpname("insights").gr_gid
```
kylape: Is this always the same as `pwd.getpwnam("insights").pw_gid`?

insights_client/__init__.py
```diff
@@ -30,13 +31,16 @@
...
 except:
-    insights_uid, insights_gid = None, None
-    raise
+    sys.exit("User and group 'insights' not found. Exiting.")
 
 
 def demote(uid, gid, phase):
-    if uid and gid and phase != "collect":
+    user_is_root = os.geteuid() is 0
+    if not user_is_root:
```
kylape: Small nitpick - IMHO `if os.geteuid() == 0` is just as readable.

insights_client/__init__.py
```diff
@@ -97,20 +106,34 @@ def _main():
...
     """
 
+    # check for insights user/group
     if not (insights_uid or insights_gid):
         print("WARNING: 'insights' user not found.  Using root to run all phases")
 
+    # check if the user is in the insights group
+    # make sure they are not root
+    in_insights_group = insights_grpid in curr_user_grps
+    is_root = os.geteuid() is 0
```
kylape: Same nitpick as above

insights_client/__init__.py
brantleyr: I don't think so? I would feel safer if we explicitly got the group ID. But you might be correct.

insights_client/__init__.py
brantleyr: insights_grpusers is not used anywhere so I can remove that.

insights_client/__init__.py
brantleyr: Do you mean `if os.geteuid() != 0`?

insights_client/__init__.py
kylape: Yeah that's fine.

PR #10, opened 2017-09-08
dont require root to run cron

brantleyr:


kylape:
Wouldn't these changes only be useful if an unprivileged user somehow set up their own cron.daily folder and linked our cron file to that folder?

brantleyr:
I was just respecting the fact that we no longer require root anywhere to run the client.

PR #14, opened 2017-09-15
improve how cron links are created with new install

SteveHNH:
Only link the old cron job to the new one if there is an old one to start with. Otherwise, just create the `/etc/cron.daily/insights-client` symlink


SteveHNH:
closing because we've changed how we want to do cron scheduling

PR #15, opened 2017-09-18
Fix permissions on rpm.egg and remove cron installer

SteveHNH:
Remove the cron installation from the RPM so we can move it to the register steps. Also fix permissions on rpm.egg so they are world readable

### DIFF
--- a/etc/insights-client.spec
+++ b/etc/insights-client.spec
@@ -80,7 +80,7 @@ ln -sf %{_bindir}/insights-client %{_bindir}/redhat-access-insights
 if ! [ -d "/etc/redhat-access-insights" ]; then
 mkdir /etc/redhat-access-insights
 fi
-# Symlink new cron job if the old one exists. Remove the old one                                               
+# Symlink new cron job if the old one exists. Remove the old one
 if [ -f "/etc/cron.daily/redhat-access-insights" ]; then
 rm -f /etc/cron.daily/redhat-access-insights
 ln -sf /etc/insights-client/insights-client.cron /etc/cron.daily/insights-client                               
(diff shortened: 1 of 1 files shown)


kylape:
After checking out your branch, building and installing the RPM, I now also have group write perms on the egg:

```
$ sudo ls -l /etc/insights-client
total 548
-rw-rw----+ 1 root root   4586 Sep 13 10:59 cert-api.access.redhat.com.pem
-rw-rw----+ 1 root root   1144 Sep 13 10:59 insights-client.conf
-rwxrwxr-x+ 1 root root    456 Sep 13 10:59 insights-client.cron
-rw-rw----+ 1 root root   1147 Sep 13 10:59 redhattools.pub.gpg
-rw-rw-r--+ 1 root root 536275 Sep 19 10:10 rpm.egg
-rw-rw-r--+ 1 root root    811 Sep 15 12:50 rpm.egg.asc
```

SteveHNH:
@kylape weird. Take Two! Let me give it another shot.

SteveHNH:
@kylape should be good to go now. I moved some of the permissions stanzas around for readability. Goes from defaultattr down to individually set permissions per file.

etc/insights-client.spec
```diff
@@ -80,11 +80,13 @@ ln -sf %{_bindir}/insights-client %{_bindir}/redhat-access-insights
 if ! [ -d "/etc/redhat-access-insights" ]; then
 mkdir /etc/redhat-access-insights
 fi
+# Symlink new cron job if the old one exists. Remove the old one
```
kylape: Looks like there's some trailing whitespace in this if block.

kylape:
The permissions look good now!

SteveHNH:
@kylape whitespace removed

PR #17, opened 2017-09-19
change egg hierarchy, run phases dynamically

gravitypriest:


insights_client/__init__.py
```diff
@@ -15,11 +15,17 @@
 import logging.handlers
 from subprocess import PIPE
 
-# setup eggs
+__author__ = 'Richard Brantley <rbrantle@redhat.com>, Jeremy Crafts <jcrafts@redhat.com>, Dan Varga <dvarga@redhat.com>'
```
kylape: I thought we were removing \_\_author__ and STDOUT_PREFIX

brantleyr: Yea those can be :wastebasket:.

PR #20, opened 2017-09-20
Let parent process's stderr pass through to children

kylape:
This enables real time console logging instead of dumping it all at once
after the child process has completed.


kylape:
This has become superseded by https://github.com/RedHatInsights/insights-client/pull/17

PR #24, opened 2017-09-25
Give world read perms to site-package files

kylape:
I found I couldn't run `insights-client` as anyone but root.  @SteveHNH can you review these permission changes in case I opened something up I shouldn't have?

### DIFF
--- a/etc/insights-client.spec
+++ b/etc/insights-client.spec
@@ -175,7 +175,7 @@ test "x$RPM_BUILD_ROOT" != "x" && rm -rf $RPM_BUILD_ROOT
 %attr(644,root,root) /etc/insights-client/rpm.egg
 %attr(644,root,root) /etc/insights-client/rpm.egg.asc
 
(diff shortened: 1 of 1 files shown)


SteveHNH:
@kylape Permissions look fine to me. There are a couple more tweaks that need to be made. I'll get those done outside of this PR since they are in different spots.

etc/insights-client.spec
```diff
@@ -175,8 +175,9 @@ test "x$RPM_BUILD_ROOT" != "x" && rm -rf $RPM_BUILD_ROOT
 %attr(644,root,root) /etc/insights-client/rpm.egg
 %attr(644,root,root) /etc/insights-client/rpm.egg.asc
 
-%attr(-,root,root) %{python_sitelib}/insights_client*.egg-info
-%attr(-,root,root) %{python_sitelib}/insights_client/*.py*
+%attr(655,root,root) %dir %{python_sitelib}/insights_client*.egg-info
```
SteveHNH: Change this to 755 to give root execute permissions on the directory.

PR #25, opened 2017-09-26
remove legacy conf vars

brantleyr:


brantleyr:
If we are removing these vars from the conf/man documentation we should remove them from the shipped .conf file as well.

kylape:
Are `cert_verify` and `gpg` considered legacy?  I thought they were still being used.

brantleyr:
I don't think we wanted them to be known/user-facing anymore. We removed gpg and cert stuff from the flags and man pages, so I assumed it should be removed from the shipped configuration files as well. If not, then disregard the PR.

PR #26, opened 2017-09-26
make config env work

brantleyr:


kylape:
Is there a particular reason we can't simply clone `os.environ` and throw in the extra keys we need, i.e. `INSIGHTS_PHASE` and `PYTHONPATH`?  That seems simpler, especially since that boolify stuff should already be in the core.

brantleyr:
All done. We now pass through the entire environment, with our keys added.

PR #30, opened 2017-09-27
Load phase functions from new module

kylape:
This allows us to support multiple versions of the phase functions simultaneously in one egg.

Please merge in lockstep with https://github.com/RedHatInsights/insights-core/pull/505

### DIFF
--- a/insights_client/__init__.py
+++ b/insights_client/__init__.py
@@ -96,7 +96,8 @@ def _main():
     """
     try:
         # flake8 complains because these imports aren't at the top
-        from insights.client import InsightsClient, get_phases  # noqa E402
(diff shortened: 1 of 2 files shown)


brantleyr:
@kylape this makes each RPM release specific to a phase version correct? So at some point we might release an RPM that would import 'v2?'

kylape:
@brantleyr yes that's the idea.  We don't _have_ to switch to v2 on a new RPM release, but this gives us the option if we want to.

brantleyr:
@kylape also I don't have perms to merge stuff in 'insights-core' so you'll have to handle that lockstep merge.

PR #33, opened 2017-10-04
fix egg python path

brantleyr:


brantleyr:
This is an issue because every time that "run.py" runs and gets the "client" it does not have access to the other EGGs (last_stable, or newest, or ENV).

brantleyr:
Closing this PR. Not the issue.

PR #45, opened 2017-12-05
take out pyyaml from setup.py

gravitypriest:


kylape:
Sorry I missed this for so long.  We actually want to keep pyyaml, so I'm closing this in favor of #46

PR #47, opened 2018-01-22
Use systemd timer instead of cron on RHEL 7 and up

kylape:
One thing to note is that I'm doing a `sysemctl daemon-reload` when uninstalling on RHEL 7.  Not sure if this is okay to do, but otherwise I get `Access denied` errors when running `systemctl status insights-client` after uninstallation.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,6 +3,13 @@
 from setuptools import setup, find_packages
 import subprocess
 
+with open("/etc/redhat-release") as fp:
+    r = fp.read().strip()
+
+release = r.split("release")[1].strip()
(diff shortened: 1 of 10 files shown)


kylape:
Please take note of the system resource limitations defined in the service file.  These are definitely up for discussion.

setup.py
```diff
@@ -3,6 +3,13 @@
 from setuptools import setup, find_packages
 import subprocess
 
+with open("/etc/redhat-release") as fp:
```
brantleyr: Can this be imported from major_version.py?

kylape: Yeah I should be able to do that.

PR #57, opened 2018-06-07
BALEET symlinks, pass the SFR version to core instead

gravitypriest:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1570172

### DIFF
--- a/etc/insights-client.spec
+++ b/etc/insights-client.spec
@@ -157,10 +157,9 @@ fi
 mkdir -p /etc/redhat-access-insights
 ln -sf /etc/insights-client/insights-client.conf /etc/redhat-access-insights/redhat-access-insights.conf
 ln -sf /etc/insights-client/insights-client.cron /etc/redhat-access-insights/redhat-access-insights.cron
-ln -sf /etc/insights-client/.registered /etc/redhat-access-insights/.registered
-ln -sf /etc/insights-client/.unregistered /etc/redhat-access-insights/.unregistered
 ln -sf /etc/insights-client/.lastupload /etc/redhat-access-insights/.lastupload
 ln -sf /etc/insights-client/machine-id /etc/redhat-access-insights/machine-id
+
 if [ -f "/etc/insights-client/.lastupload" ]; then
     setfacl -m g:insights:rwx /etc/insights-client/.lastupload
 fi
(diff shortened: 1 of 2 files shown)


etc/insights-client.spec
```diff
@@ -157,10 +157,9 @@ fi
 mkdir -p /etc/redhat-access-insights
 ln -sf /etc/insights-client/insights-client.conf /etc/redhat-access-insights/redhat-access-insights.conf
 ln -sf /etc/insights-client/insights-client.cron /etc/redhat-access-insights/redhat-access-insights.cron
-ln -sf /etc/insights-client/.registered /etc/redhat-access-insights/.registered
-ln -sf /etc/insights-client/.unregistered /etc/redhat-access-insights/.unregistered
```
kylape: What if the symlinks are already there from a previous installation?

insights_client/__init__.py
```diff
@@ -82,7 +82,9 @@ def run_phase(phase, client):
         # setup the env
         insights_env = {
             "INSIGHTS_PHASE": str(phase['name']),
-            "PYTHONPATH": str(egg)
+            "PYTHONPATH": str(egg),
+            # get the binary name so we can use the appropriate conf dir for .registered files
+            "INSIGHTS_SFR": sys.argv[0].rsplit('/', 1)[1]
```
kylape: I think `os.path.basename(sys.argv[0])` might be cleaner here.

gravitypriest:
My assumption is that if they're still using `redhat-access-insights` they're using it wholesale but you make a good point. One alternative is keeping dotfiles in both locations, which I was trying to avoid.

etc/insights-client.spec
gravitypriest: In that case should we delete these files on install? Shouldn't hurt since if they have an existing registration the API should echo back an OK and we'll regenerate them.

kylape:
I definitely don't want to have to maintain the same thing in two places, but it sounds like that's the best/only option.  We are already keeping `/etc/redhat-access-insights` around (even if you never had `redhat-access-insights` previously installed), so it doesn't seem any worse to just make the files in there "real".

I imagine the code to write to both locations can be abstracted pretty easily so that it's not _that_ much more disgusting :).

gravitypriest:
image

gravitypriest:
Thankfully I already wrote that fix once before so hopefully I can find it somewhere on my machine

PR #63, opened 2018-07-18
Obfuscate passwords with special characters

Glutexo:
SED command used to obfuscate passwords kept ones containing special characters partially unobfuscated. Replaced the used `\w` matching pattern with a custom character class that matches ASCII latin letters, decimal digits and some special symbols.

`\w` is generally problematic, because it is heavily dependent on locale settings and in non-ASCII-only mode it matches any Unicode word-character. This includes matching more than one ASCII letter as a single character and other quirks. This can lead to security issues as well. `[a-zA-Z0-9_]` is still little tricky, since it depends on `LC_COLLATION` locale setting; however _insights-client_ runs the SED command with a `LC_ALL=C` environment variable set, which ensures byte-strict ordering. [man perlre](http://manpages.org/perlre)

After a discussion with @kylape we came to a conclusion that in this SED script we handle only ASCII word-characters, not considering Unicode at all. The list of allowed special chars is `!@#$%^&*()-_+=/`, based on one provided by @YakovLittle.

This PR improves obfuscation results for secure passwords. As a downside this can damage results if a password would appear e.g. in a URI query format: `password=abc&username=def` gets obfuscated to `password=********` because of the ampersand being considered a part of the password. See more examples in the [tests](
https://github.com/RedHatInsights/insights-client/compare/master...Glutexo:obfuscate_passwords_with_special_chars?expand=1#diff-9e8bdfbdceed901aae26c617f024bb5a).

This is more a band-aid solution, since obfuscating the output without any understanding of the format is generally not a good idea. Using SED with all its peculiarities to do that comes with many downsides too. This is going to be addressed in a separate issue.

Added `pytest` to development dependencies.

https://bugzilla.redhat.com/show_bug.cgi?id=1582144

Asking @gravitypriest for review.


Glutexo:
@eduardocerqueria Jenkins is failing on RHEL 7 with

> /bin/sh: rpmbuild: command not found

I guess this is not caused by my changes, but if you think it is, please let me know.

eduardocerqueira:
Hi Štěpán,

The error should not be related to your PR. I am in PTO Today but will see
if someone from my team can take a look at this otherwise I can fix it
tomorrow morning.

When we fix we can re-test your PR if not merged yet.

--Eduardo

eduardocerqueira:
test this please

eduardocerqueira:
test this please

eduardocerqueira:
test this please

eduardocerqueira:
All set now!

Glutexo:
@eduardocerqueira Does Jenkins run the new test? I added _pytest_ to the development dependencies, thus running `pytest` should be sufficient.

eduardocerqueira:
This pipeline only runs the tests from insights-core (for new egg) and
insights-tests ( functional automated tests for client ) if your test was
added to insights-client I can expand tje pipeline to start running these
tests too. Thx to ask and I will take a look.

eduardocerqueira:
Hi @Glutexo just let you know I added a new stage to run unit tests for insights-client in the CI pipeline testing PR for insights-client project. 

https://github.com/RedHatInsights/insights-tests/pull/247

PR #68, opened 2018-07-25
[WIP] Add user and group deletion to post-uninstall

Glutexo:
_insights_ user and _group_ are created when installing the client. They were however not removed on uninstall. Altered the post-uninstall RPM scriptlet in the spec.

Created a rather big [test file](https://github.com/Glutexo/insights-client/blob/e7a685f06d68a59cdf4b9d1b4c945ff37397bea8/insights_client/tests/test_rpm.py) for that. It builds the RPM and install that with all necessary dependencies in a sandbox folder. This test is heavy on disk and network resources, so it is not run by default. Use `TEST_RPM=True pytest insights_client/tests/test_rpm.py` to run it. It requires administrator privileges and will ask for sudo password on the first _yum_ [call](https://github.com/Glutexo/insights-client/blob/e7a685f06d68a59cdf4b9d1b4c945ff37397bea8/insights_client/tests/test_rpm.py#L122).

RPM building in the test is making [use](https://github.com/Glutexo/insights-client/blob/e7a685f06d68a59cdf4b9d1b4c945ff37397bea8/insights_client/tests/test_rpm.py#L108) of the [Makefile](https://github.com/Glutexo/insights-client/blob/e7a685f06d68a59cdf4b9d1b4c945ff37397bea8/Makefile). It however contains currently some bugs that don’t allow to simply run `make rpm`. I‘ll address those in a separate PR and then I’ll remove the unnecessary [duplicate logic](https://github.com/Glutexo/insights-client/blob/e7a685f06d68a59cdf4b9d1b4c945ff37397bea8/insights_client/tests/test_rpm.py#L110) in the test in one more PR.

To determine whether the user and group is present, I wrote a simple Python module in [_insights_client/tests/helpers/getent.py_](https://github.com/Glutexo/insights-client/blob/e7a685f06d68a59cdf4b9d1b4c945ff37397bea8/insights_client/tests/helpers/getent.py). The system’s _getent_ utility unfortunately does not allow specifying custom root and chrooting on the other hand requires unnecessary administrator access. To use this as a module in the tests, I have to turn the [_insights_client/tests_](https://github.com/Glutexo/insights-client/blob/e7a685f06d68a59cdf4b9d1b4c945ff37397bea8/insights_client/tests/__init__.py) and [_insights_client/tests/helpers_](https://github.com/Glutexo/insights-client/blob/e7a685f06d68a59cdf4b9d1b4c945ff37397bea8/insights_client/tests/helpers/__init__.py) folders into modules. To exclude them from the egg I added an appropriate configuration to [_setup.py_](https://github.com/Glutexo/insights-client/blob/e7a685f06d68a59cdf4b9d1b4c945ff37397bea8/setup.py#L73).

I’d like to ask @gravitypriest for a review in this state. Just please don’t merge it yet. As @kylape pointed out, it is still necessary to ensure that it was the _insights-client_ that created the user and group. I suggested storing the uid/gid in _/etc/insights-client_ on install and use in on uninstall. Can we do it this way? This approach wouldn’t however solve the bug for existing installations.

https://bugzilla.redhat.com/show_bug.cgi?id=1592996
https://projects.engineering.redhat.com/browse/RHIPLAT1-308

### DIFF
--- a/etc/insights-client.spec
+++ b/etc/insights-client.spec
@@ -202,6 +202,9 @@ rm -rf /etc/redhat-access-insights/
 rm -f /etc/cron.daily/redhat-access-insights
 rm -f /etc/cron.weekly/redhat-access-insights
 fi
+userdel -rf insights > /dev/null 2>&1
+groupdel insights > /dev/null 2>&1
+exit 0
 
 %clean
 test "x$RPM_BUILD_ROOT" != "x" && rm -rf $RPM_BUILD_ROOT
--- a/insights_client/tests/helpers/getent.py
+++ b/insights_client/tests/helpers/getent.py
@@ -1,0 +1,84 @@
+# -*- coding: utf-8 -*-
+
+from os.path import join
+
+PASSWD_PATH = join('etc', 'passwd')
+GROUP_PATH = join('etc', 'group')
+
+
+def passwd_exists(username, file):
+    """
+    Checks whether a user with a given name exists in the given passwd database.
+    """
+    def matcher(line):
+        """
+        Matches a line corresponding to a user with a given name.
+        """
+        return ColonSeparatedLine(line)[0] == username
+
+    database = Database(file)
+    return database.exists(matcher)
+
+
+def group_exists(groupname, file):
+    """
+    Checks whether a group with a given name exists in the given group database.
+    """
+    def matcher(line):
+        """
+        Matches a line corresponding to a group with a given name.
+        """
+        return ColonSeparatedLine(line)[0] == groupname
+
+    database = Database(file)
+    return database.exists(matcher)
+
+
+class ColonSeparatedLine:
+    """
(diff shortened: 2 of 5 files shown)


Glutexo:
I’d just like to point to the fact that the checks are passing now. However from what @kylape said elsewhere it looks like not only we don’t know how to decide whether to remove the user and group, but we are not sure whether we actually want to create them in the first place.

Glutexo:
@kylape suggested closing this as we’re going to abandon creating the user and group at all. We’ve already removed ACLs (#70). I agree with that, I just couldn’t find any filed issue/ticket describing this planned change. Will close this when we’ll have one.

kylape:
https://bugzilla.redhat.com/show_bug.cgi?id=1570867

esakaiev:
test this please

PR #69, opened 2018-07-27
Format CLI switches properly in Markdown

dLobatog:
Before this fix, the README shows all switches in the same line. 
This is meant to fix that and format it in a more readable way.


kylape:
add to whitelist

PR #74, opened 2018-08-31
Spec/service/cron for insights-compliance

dLobatog:
These are meant to be scheduled independently of regular insights-client
runs, as OpenSCAP runs have the potential to be slower than 10 minutes,
which is insights-client timeout

**DO NOT MERGE YET - `insights-client --compliance` will fail as it's not implemented**


esakaiev:
test this please

esakaiev:
test this please

m-horky:
Closing for the lack of activity. Please reopen it (probably with new implementation) if this is something Compliance needs.

PR #75, opened 2018-10-15
remove insights user

gravitypriest:


gravitypriest:
May also need to remove the user from the system in the specfile for existing installations.

PR #84, opened 2019-07-31
figure out whether rpm.egg or last_stable.egg is newer and use that one

gravitypriest:
- also change macros in specfile for RHEL 6 compat

### DIFF
--- a/insights_client/__init__.py
+++ b/insights_client/__init__.py
@@ -7,7 +7,7 @@ from __future__ import print_function
 import os
 import sys
 import subprocess
-from subprocess import PIPE
+from subprocess import Popen, PIPE
 import shlex
 import logging
 import logging.handlers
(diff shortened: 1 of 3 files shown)


insights_client/__init__.py
```diff
@@ -28,26 +27,62 @@ def log(msg):
...
+
+
+def sorted_eggs(eggs):
+    '''
+    Sort eggs to go into sys.path by highest version
+    '''
+    if len(eggs) < 2:
+        # nothing to sort
+        return eggs
+    if egg_version(eggs[0]) > egg_version(eggs[1]):
```
kylape: Perhaps you should use a version parsing library to do the compare?  e.g. `packaging.version` or `distutils.version.LooseVersion` (if `packaging.version` isn't in rhel 6)

gravitypriest:
@kylape It was suggested a customer might have have auto-update enabled, then one day disable it and stop receiving newer core versions. Then when they upgrade the RPM they'd want to have the newest version, but the way it's set up currently, it'll always load the /var/lib/insights eggs before the RPM-bundled one.

insights_client/__init__.py
gravitypriest: Will investigate

PR #85, opened 2019-08-26
Display a message in motd Insights has not yet been used

stefwalter:
Insights represents key functionality of the RHEL operating system. Customers are missing out on Value of Subscription when not enabled. Thus we display a message in motd when Insights is installed but has never been used.

We use a ```/etc/motd.d/``` drop in file to do this. Once insights-client has been
used (and a new egg has been retrieved) we drop this message.

The message can be edited or removed in the configuration file:

    /etc/insights-client/insights-client.motd

This motd message seems to represent part of the insights-client command-line
RPM packaged wrapper of InsightsClient and is less suited to the core functionality.

### DIFF
--- a/etc/insights-client.motd
+++ b/etc/insights-client.motd
@@ -1,0 +1,3 @@
+This system is not registered to Red Hat Insights. See https://cloud.redhat.com/
+To register this system, run: insights-client --register
+
--- a/docs/insights-client.8
+++ b/docs/insights-client.8
@@ -63,6 +63,9 @@ Check this host's registration status with Red Hat Insights.
 .IP "--net-debug"
 Log network activity.
 
+.SH "MOTD"
+A message will be displayed on login about \fBinsights\-client\fP if installed but has not yet been used at least once. To change or remove this message, edit the \f8/etc/insights-client/insights-client.motd\fP file.
+
 .SH "SEE ALSO"
 .BR insights-client.conf (5)
 
--- a/etc/insights-client.spec
+++ b/etc/insights-client.spec
@@ -74,6 +74,9 @@ pathfix.py -pni "%{__python3}" %{buildroot}%{_bindir}/redhat-access-insights
 %else
 %{__python} setup.py install --root=${RPM_BUILD_ROOT} $PREFIX
 %endif
+#Link motd into place. setup.py does not support symlinks
+mkdir -p ${RPM_BUILD_ROOT}/etc/motd.d
+ln -snf /etc/insights-client/insights-client.motd ${RPM_BUILD_ROOT}/etc/motd.d/insights-client
 
 %post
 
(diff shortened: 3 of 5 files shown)


etc/insights-client.motd
```diff
@@ -0,0 +1,2 @@
+Missing out on Insights benefits. To register run: insights-client --register
```
andreasn: Needs to communicate the system status more clearly, and also the benefits of Insights.
https://www.redhat.com/en/technologies/management/insights is probably the place that has that written up best.

What about something like:
```
This system is not registered to Red Hat Insights.
https://www.redhat.com/en/technologies/management/insights

To register this system, run: insights-client --register
```

andreasn: Possibly https://cloud.redhat.com is a better (and for sure much more stable URL).
This is what we went for in https://github.com/candlepin/subscription-manager/pull/2089

stefwalter: Thanks. Updated the message.

stefwalter:
I've pushed new changes:

- Updated the message as suggested by @andreasn
- Added documentation to ```man insights-client``` on how to edit/remove the message

gravitypriest:
Hey @stefwalter, real quick -- is motd.d available in 8.0 or is this strictly an 8.1 and up feature?

stefwalter:
I see ```/etc/motd.d``` in both both RHEL 7.7 and RHEL 8.0 ... not sure about 6.x, but my guess would be that it's not available there.

gravitypriest:
No worries, all I needed to know. We're just building RHEL 8 for now so this'll be fine. Thanks for the PR.

gravitypriest:
Upon install and registration, running `rpm -V insights-client` complains that the motd file is missing from `/etc/motd.d`. Is this something we're okay with?

stefwalter:
```/etc/motd/insights-client``` is marked as a config file. @martinpitt do you think it's okay if a config file is marked as missing during ```rpm -V```?

If we'd prefer, we could replace it with a link to /dev/null instead of removing it (once insights-client has been used) and have the same affect.

gravitypriest:
That'd be preferable, I think. We just had to deal with some rpm -V errors in a previous release and don't want to start having more of them. In the interest of time I currently moved the symlink to post-install and took it out of the RPM-managed files, but that's not the most ideal solution.

martinpitt:
> do you think it's okay if a config file is marked as missing during `rpm -V`?

I'm afraid I'm not enough of an rpm packaging expert to give a definitive answer here.

> If we'd prefer, we could replace it with a link to /dev/null instead 

I do know that rpm has an irritating (for me, as I grew up with dpkg) tendency to put back deleted configuration files. Possibly the `%config(noreplace)` avoids that, but indeed having an explicit /dev/null symlink is more explicit and easier to understand IMHO. Then it's a changed configuration file instead of a removed one.

stefwalter:
Thinking about this further, an approach related to ```%post``` like @gravitypriest has taken may also be necessary. This is because of upgrades. We don't want this motd file to just appear on upgrades.

stefwalter:
@gravitypriest What do you think about #88

PR #88, opened 2019-08-30
Point the motd message at /dev/null instead of removing

stefwalter:
Once insights-client has been used, we remove the message in
motd recommending use of Insights. Instead of removing the symlink
we instead point it at /dev/null.

We also need to be careful for upgrades. Upgrading the
insights-client package on a system where Insights has already
been used should not result in a motd message.

### DIFF
--- a/etc/insights-client.spec
+++ b/etc/insights-client.spec
@@ -73,7 +73,6 @@ pathfix.py -pni "%{__python3}" %{buildroot}%{_bindir}/redhat-access-insights
 %else
 %{__python2} setup.py install --root=${RPM_BUILD_ROOT} $PREFIX
 %endif
-#Link motd into place. setup.py does not support symlinks
 
 %post
 
@@ -136,8 +135,11 @@ if ! [ -d "/var/lib/insights" ]; then
 mkdir -m 644 /var/lib/insights
 fi
 
-# symlink the motd file
-ln -snf /etc/insights-client/insights-client.motd /etc/motd.d/insights-client
+#Link motd into place. Only do so if motd file doesn't exist
+if ! [ -L /etc/motd.d/insights-client ]; then
+    mkdir -p /etc/motd.d
+    ln -snf /etc/insights-client/insights-client.motd /etc/motd.d/insights-client
+fi
 
 # always perform legacy symlinks
 %posttrans
--- a/insights_client/__init__.py
+++ b/insights_client/__init__.py
@@ -21,6 +21,7 @@ ENV_EGG = os.environ.get("EGG")
 NEW_EGG = "/var/lib/insights/newest.egg"
 STABLE_EGG = "/var/lib/insights/last_stable.egg"
 RPM_EGG = "/etc/insights-client/rpm.egg"
+MOTD_FILE = "/etc/motd.d/insights-client"
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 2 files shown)


stefwalter:
The Travis failure seems unrelated:

```
$ flake8
./insights_client/__init__.py:42:23: F841 local variable 'e' is assigned to but never used
The command "flake8" exited with 1.
```

etc/insights-client.spec
```diff
@@ -136,8 +135,11 @@ if ! [ -d "/var/lib/insights" ]; then
 mkdir -m 644 /var/lib/insights
 fi
 
-# symlink the motd file
-ln -snf /etc/insights-client/insights-client.motd /etc/motd.d/insights-client
+#Link motd into place. Only do so if insights-client new install
+if  [ $1 -eq 1  ]; then
+    mkdir -p /etc/motd.d
+    ln -snf /etc/insights-client/insights-client.motd /etc/motd.d/insights-client
+fi
```
stefwalter: @gravitypriest This is probably the only change necessary avoid the message reappearing after upgrade.

stefwalter:
@gravitypriest Thinking about this further. I think what I have here now is better complete solution for RHEL 8.1. 

During install or upgrade this places the file in ```/etc/motd.d``` if it doesn't exist. We then never remove the file, only change where it points.

stefwalter:
@martinpitt Could you review?

etc/insights-client.spec
```diff
@@ -136,8 +135,11 @@ if ! [ -d "/var/lib/insights" ]; then
 mkdir -m 644 /var/lib/insights
 fi
 
-# symlink the motd file
-ln -snf /etc/insights-client/insights-client.motd /etc/motd.d/insights-client
+#Link motd into place. Only do so if motd file doesn't exist
+if ! [ -L /etc/motd.d/insights-client ]; then
```
stefwalter: This change depends on the companion change in __init__.py ... a previous version had it as an independent fix, but that's no longer the case here.

stefwalter: And the reason is that we want the motd message to appear on upgrades ... just not reappear erroneously.

etc/insights-client.spec
```diff
@@ -136,8 +135,11 @@ if ! [ -d "/var/lib/insights" ]; then
...
+    mkdir -p /etc/motd.d
+    ln -snf /etc/insights-client/insights-client.motd /etc/motd.d/insights-client
```
martinpitt: The `-f` should be redundant now with the `-L` check from above, but harmless.

etc/insights-client.spec
```diff
@@ -136,8 +135,11 @@ if ! [ -d "/var/lib/insights" ]; then
 mkdir -m 644 /var/lib/insights
 fi
 
-# symlink the motd file
-ln -snf /etc/insights-client/insights-client.motd /etc/motd.d/insights-client
+#Link motd into place. Only do so if motd file doesn't exist
```
martinpitt: What happened to checking for new install vs. upgrade? We still only want to do this on fresh installs, AFAIUI.

stefwalter: No we do want it to appear on upgrades to 8.1 or to 8 z-stream ... just not *reappear* on upgrades if the system already uses insights.

stefwalter: Hence the logic of, if the symlink exists, do nothing. And insights-client always keeps a symlink there ... albeit to ```/dev/null``` once insights-client has been used.

martinpitt: Nevermind, you said above that they *should* appear on upgrades.

etc/insights-client.spec
martinpitt: Actually, I believe this is weird/wrong. The documentation says "by pointing /etc/motd.d/insights-client at an empty file", so you woud be fully entitled to do "touch /etc/motd.d/insights-client" and expect it to be and stay quiet. But that will fail the `-L` check, and this `ln -f` will remove the file and replace it with a symlink. So I'd suggest to drop the `-f` (for safety) and fix the check to

    if [ ! -e /etc/motd.d/insights-client -a ! -L /etc/motd.d/insights-client ]; then

stefwalter: Hmmm. While it may be surprising ```touch /etc/motd.d/insights-client``` would work ... as the symlink would always redirect that somewhere.

In addition the documentation in the manual page states:

```
To change or  remove  this  message, edit the /etc/insights-client/insights-client.motd file.
```

I'm not sure this is a bug and have @gravitypriest doing gymnastics to redo the build, yet again.

martinpitt: For the record, this isn't a dealbreaker, just a tiny glitch. No need to re-spin.

PR #89, opened 2019-08-30
Don't install the motd message if insights-client used

stefwalter:
If insights-client has been used, then don't install the motd
message. As before don't install a motd message if such a symlink
already exists.

Further testing of the latest builds resulted in this fix.


stefwalter:
I have done a scratch build for this and tested all combinations of: 8.0, 8.1, Insights in use before upgrade ... and not in use before upgrade, fresh install, removal, remove then install

PR #100, opened 2019-11-19
Remove references to redhat-access-insights

subpop:
redhat-access-insights is end-of-lifed; the symlinks maintaining backwards compatibility inside `/etc/redhat-access-insights` is no longer necessary.

However, the migration steps remain in the `%post` stage, in case clients do still have the legacy config directory.


subpop:
This is obsoleted by #108

PR #106, opened 2020-01-30
Build system updates

subpop:
What started out as a minimally invasive build system update grew somewhat. I've rebuilt the setuptools-based build script (setup.py) to be relocatable (for example, it can install on Debian). A number of optional arguments have been added to the `install` command; these arguments override default values for installation locations. For example, to install with a prefix of `/usr`, but specifying the config directory as `/etc`, run: `python setup.py install --prefix=/usr --sysconfdir=/etc`. These names and default paths take a lead from autotools variable names (so `/var` is called `--localstatedir`).

The source tarball should also include *all files* that are required to build the project from source on any distribution. Therefore the systemd unit files and cron scripts are included, regardless of which host distribution the tarball is created on. The build system takes care of installing the unit files only if `/usr/lib/systemd/system` exists on the installation target at the time of installation. The source tarball can be constructed from any distribution or platform; it is no longer necessary to generate a tarball for RHEL6 or RHEL6, etc.

* Removed RPM spec file from project (it should exist in dist-git exclusively)
* Removed `Makefile` from project
* Removed `major_version.py` (I'm assuming this was not deliberately part of the `insights_client` public API)
* Download uploader.json and rpm.egg during `sdist`

It is important to note that this project repository can no longer build RPMs. This breaks Jenkins.


patchkez:
test this please

patchkez:
test this please

patchkez:
test this please

patchkez:
test this please

patchkez:
test this please

patchkez:
test this please

patchkez:
test this please

patchkez:
test this please

patchkez:
test this please

patchkez:
test this please

patchkez:
test this please

subpop:
test this please

patchkez:
test this please

patchkez:
test this please

subpop:
test this please

subpop:
Prefer #108

PR #108, opened 2020-03-09
Use autotools

subpop:
Python's setuptools and distutils toolset is riddled with portability and compatibility problems. The *only* advantage setuptools seems to offer is some automatic "egg-info" creation, but since we aren't distributing insights-client via pypi or as an "egg" anyway, the egg-info is useless. Rather, insights-client is distributed more like a traditional software package: some source code, an executable entrypoint script, and some related data files. setuptools is ill-suited for this role, so I ported the distribution and build system to autotools. autotools is designed to solve portability and compatibility problems from the beginning, so it is a far more appropriate toolset to use to distribute and install insights-client.

This PR replaces the old 'Makefile' and 'setup.py' code with a typical autoconf/automake layout. I included an autogen.sh script to bootstrap the configure script. Notably, this PR makes zero changes to the source code and data files[1]. I just drop the old build scripts and replace them with configure.ac and Makefile.am files, where appropriate. I also made use of a drop-in makefile for managing .gitignore files. autotools are notoriously prolific creating scripts and extra files (all due to its goal of being 100% portable), so the git.mk makefile will automatically create .gitignore files where appropriate. If you find your .gitignore files are missing, run 'make gitignore' from the top_srcdir and it will create gitignore files. See the header comments inside git.mk for more details on how to use it.

Tests are now run using autotools. While it is technically still possible to run tests directly with pytest, one would first need to make all the test files (since at least one, test_sed.py, is a generated source file). The easiest way to run the tests with pytest is to use automake's 'check' target. Automake then generates a simple test driver script and invokes pytest. A simple pass/fail report is printed to stdout. It is also possible to pass commandline arguments to pytest by setting the PYTEST_FLAGS autoconf variable (this is done during the ./configure step). For example:

```bash
./configure PYTEST_FLAGS="--capture=sys --junitxml=insights_client_unittests.xml"
make check
```

1: I lied. It makes a single change. 'constants.py' is now a generated source file. constants.py.in is distributed with the source tarball and generates constants.py and compile time. This makes configure.ac the canonical source of truth for the version string. The test file 'test_sed.py' made use of a hard-coded relative path to locate the .exp.sed file. I converted test_sed.py to a generated source file too in order to inject an absolute path to the .exp.sed file, depending on where the file is executed from.


patchkez:
test this please

patchkez:
test this please

patchkez:
test this please

patchkez:
test this please

patchkez:
test this please

subpop:
test this please

patchkez:
test this please

patchkez:
test this please

patchkez:
test this please

patchkez:
test this please

subpop:
test this please

patchkez:
test this please

patchkez:
test this please

patchkez:
test this please

PR #109, opened 2020-03-17
Add "insights-client-results.service"

subpop:
Adds a systemd "path" unit that watches /etc/insights-client/.lastupload for modification. The unit it triggers then sleeps for 120 seconds (:frowning:) before running --check-update.

The Also= clause in insights-client.timer causes insights-client-results.path to be enabled when insights-client.timer is. This keeps the user experience "the same". Enabling just insights-client.timer will also enable insights-client-results.path. insights-client-results.path adds a PartOf= dependency to insights-client.timer, so that it is started and stopped when insights-client.timer is started.

The new ENABLE_CHECK_RESULTS variable in sysconfig enables this same functionality for cron-based systems.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -4,6 +4,9 @@ AM_INIT_AUTOMAKE([1.11 foreign])
 AC_CONFIG_MACRO_DIR([m4])
 AM_SILENT_RULES([yes])
 
+pkgsysconfdir=${sysconfdir}/$PACKAGE
+AC_SUBST([pkgsysconfdir])
+
 AC_PROG_SED
 PKG_PROG_PKG_CONFIG
 if test -z "$PKG_CONFIG"; then
--- a/data/cron/Makefile.am
+++ b/data/cron/Makefile.am
@@ -1,7 +1,7 @@
 pkgsysconfdir = $(sysconfdir)/insights-client
 sysconfigdir  = $(sysconfdir)/sysconfig
 
-dist_pkgsysconf_DATA = insights-client.cron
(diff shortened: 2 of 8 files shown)


subpop:
test this please

patchkez:
test this please

patchkez:
test this please

patchkez:
test this please

mvollmer:
What about using the approach here: https://github.com/mvollmer/insights-post-upload/blob/master/insights-poll.sh ?  That would give us results much earlier in the normal case.

subpop:
> What about using the approach here: https://github.com/mvollmer/insights-post-upload/blob/master/insights-poll.sh ? That would give us results much earlier in the normal case.

I like this approach, definitely. I'd need to run this by our broader platform team to make sure we don't overwhelm the platform services too much. Each invocation of insights-client (even when dropping into the early `--check-results` code) comes with a few of HTTP requests. We would want to test a polling approach like this before going forward with it. I'll file this as an enhancement to the more conservative "sleep 120" approach in this PR and see about improving it in a follow-up PR.

subpop:
Actually, your insights-poll.py might as well be integrated into the `--check-results` code itself. That way the unit triggers once, `insights-client` invokes itself once, and computes the time drift and polls inside the same process invocation. That would surely have less impact on the overall platform. I'll look into doing it that way.

mvollmer:
> Actually, your insights-poll.py might as well be integrated into the `--check-results` code itself.

I like this!

subpop:
I filed a JIRA to follow-up on integrating this into the insights-client directly.

PR #113, opened 2020-04-24
Update manpage 4 13

gravitypriest:
We'll want to mark the added blacklist config files as config in the specfile as well.

Other thought: we may also want to include a placeholder tags.yaml in the RPM


patchkez:
test this please

subpop:
test this please

PR #116, opened 2020-04-28
Remove echo from cron

subpop:
When a cron job generates output on stdout, it creates local mail messages in /var/spool/mail/root. This isn't necessary, especially for informational log messages like "Sleeping for N seconds". This PR removes the echo statement from the cron script.

Fixes: RHCLOUD-6140 RHBZ#1828778


patchkez:
test this please

patchkez:
test this please

PR #121, opened 2020-06-26
Collect runtime statistics about phase execution

subpop:
This PR adds the collection of runtime metrics to the "phase runner" (`run.py`). Before each phase, an "event" object is created. The event object contains details about the phase, include start times, exit codes, exception strings, etc. It uses a specialized HTTP client (not `InsightsConnection`) so that statistics can still be reported back to the platform even in the event of a major failure of `InsightsConnection`. It keys off the RHSM server value. If it's not subscription.rhsm.redhat.com, the base URL, authentication scheme and API path prefix are configured to the "Satellite" values. If the RHSM server value is subscription.rhsm.redhat.com, then it examines the insights-client.conf authmethod value, setting up the base URL, authentication, and API path prefix values for either certificate or basic authentication.

**Note**: ~The `default` test will fail until #120 is merged; the `private-rhel-8.3.0` dist-git branch we use to build the RPM from is already updated to expect files that *are* present in #120, but that are absent in this PR. Merging #120 onto master, and then rebasing this PR on top of master will make that pipeline start passing.~

### DIFF
--- a/src/insights_client/metrics.py
+++ b/src/insights_client/metrics.py
@@ -45,6 +45,7 @@ class MetricsHTTPClient(requests.Session):
         """
         if self.base_url == "cert-api.access.redhat.com":
             api_prefix = "/r/insights/platform"
+            self.verify = "/etc/insights-client/cert-api.access.redhat.com.pem"
         else:
             api_prefix = "/api"
         url = "https://{}{}/module-update-router/v1/event".format(
--- a/src/insights_client/metrics.py
+++ b/src/insights_client/metrics.py
@@ -1,6 +1,11 @@
-import socket
+import re
 
 import requests
+import rhsm
+from six.moves import configparser
+
+AUTH_METHOD_BASIC = "BASIC"
+AUTH_METHOD_CERT = "CERT"
 
 
 class MetricsHTTPClient(requests.Session):
@@ -9,7 +14,9 @@ class MetricsHTTPClient(requests.Session):
     runtime metrics about insights-client back to the Insights Platform.
     """
 
-    def __init__(self, cert_file, key_file, base_url=None):
+    def __init__(
+        self, cert_file=None, key_file=None, base_url=None, config_file=None,
+    ):
         """
         __init__ creates and configures a new MetricsHTTPClient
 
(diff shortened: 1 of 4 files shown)


subpop:
test this please

src/insights_client/metrics.py
```diff
@@ -0,0 +1,53 @@
...
+        if self.base_url is None:
+            raise Exception("Unable to connect to Red Hat Insights Platform")
+
+    def post(self, event):
+        """
+        post sends `event` to the Insights Platform.
+
+        :param event: a dictionary describing an event object
+        """
+        if self.base_url == "cert-api.access.redhat.com":
```
gravitypriest, line 46: We'll need to set `self.verify = /etc/insights-client/<whatever that pem file is>` here if connecting to the classic API.

lphiri:
@subpop  this PR piqued my interest because of some things I was looking at last week :)

Will this work when an http proxy is required? 
Also, we do have scenarios where the certs just wont work and basic auth must be used (e.g.  cloud instances in AWS etc getting updates from RHUI)

subpop:
> @subpop @gravitypriest this PR piqued my interest because of some things I was looking at last week :)
> 
> Will this work when an http proxy is required?

It should. It uses a `requests.Session` underneath, and I *believe* `requests.Session` respects the `HTTPS_PROXY` environment variable.

> Also, we do have scenarios where the certs just wont work and basic auth must be used (e.g. cloud instances in AWS etc getting updates from RHUI)

I'm hoping to avoid any dependency on code inside the egg, in order to make the event tracking isolated from the egg code. I could parse and load the config file to get the authentication mechanism. That way it could be made to work with both certificate and basic authentication.

lphiri:
@subpop  - I remember vaguely that we had some really convulated proxy config logic in auto_config  and connection at some point ( we don't just depend on ENV variables but also parse config). If that's been simplified then its good otherwise your specialized connection class may need to do something similar.

gravitypriest:
@lphiri I think it's mostly the same, but this may be a good opportunity to revisit that

subpop:
I rewrote the URL detection logic, with Satellite in mind. I removed the socket logic that attempts to connect to a list of descending servers. The approach I took varied slightly from the approach in the `auto_config.py` module. It is deliberately simple. It keys off the RHSM server value. If it's *not* `subscription.rhsm.redhat.com`, the base URL, authentication scheme and API path prefix are configured to the "Satellite" values. If the RHSM server value is `subscription.rhsm.redhat.com`, then it examines the `insights-client.conf` `authmethod` value, setting up the base URL, authentication, and API path prefix values for either certificate or basic authentication.

subpop:
test this please

subpop:
test this please

subpop:
@lphiri I rewrote the base URL detection logic. Does this seem more reasonable to you?

subpop:
test this please

patchkez:
test this please

patchkez:
test this please

subpop:
test this please

patchkez:
test this please

patchkez:
test this please

subpop:
test this please

subpop:
test this please

subpop:
test this please

subpop:
test this please

subpop:
test this please

subpop:
test this please

PR #123, opened 2020-08-03
constants: Add location variables to generated constants

subpop:
Add the common location variables, as generated by the `configure` script, to the generated constants.py file, so they are accessible within the program. This is the first in a long process of updating `insights-core.client.constants` to be able to use properly portable filesystem paths. Sadly, nothing in `insights-core` can use these paths until an RPM containing the generated constants file has been released on all RHEL distributions.


patchkez:
test this please

subpop:
test this please

subpop:
test this please

PR #126, opened 2020-08-24
Fix motd.d management

subpop:
Currently, insights-client uses the presence of newest.egg to decide whether to preserve or remove the /etc/motd.d/insights-client snippet. For clients where auto_update is disabled, the newest.egg never gets created, leading to a false claim from the motd stating that the host is not registered, when in reality, it is.

This PR changes the logic of the `update_motd_message` function to use the registration stamp files. If either .registered or .unregistered exist, the symlink in `/etc/motd.d/insights-client` is updated to point to `/dev/null`. Additionally, the function will update the symlink to point to the actual snippet file (`/etc/insights-client/insights-client.motd`) if both registration stamp files do not exist. This enables us to make a single invocation of the function during the RPM `%post` scriptlet that sets up the motd symlink based on local machine state.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -21,10 +21,6 @@ AC_PATH_PROG([WGET], [wget])
 AS_IF([test -z "$WGET"],[
     AC_MSG_ERROR([wget is missing from your system...])
 ])
-AC_PATH_PROG([STRINGS], [strings])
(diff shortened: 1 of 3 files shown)


subpop:
test this please

PR #127, opened 2020-09-04
Make calls to metrics_client.post non-fatal

subpop:
If a machine is not registered with RHSM (i.e. `/etc/pki/consumer/cert.pem` does not exist), the `MetricsHTTPClient` still configures itself to use the certificates, but then fails fatally when it cannot read that file. This PR makes the `metrics_client.post` call non-fatal; instead an error is printed but the phase continues.

### DIFF
--- a/src/insights_client/metrics.py
+++ b/src/insights_client/metrics.py
@@ -19,7 +19,7 @@ class MetricsHTTPClient(requests.Session):
         cert_file=None,
         key_file=None,
         base_url=None,
-        config_file=None,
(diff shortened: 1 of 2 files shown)


subpop:
test this please

gravitypriest (approved):

subpop:
test this please

subpop:
test this please

patchkez:
test this please

subpop:
test this please

PR #128, opened 2020-09-09
REAMDE: EGG var and configuration format

vkrizan:
Update README to note that `EGG` env var should point to built zip. Also document the configuration format and main section.


vkrizan:
@gravitypriest Would you mind having a look. Thank you.

subpop (approved):

PR #129, opened 2020-09-15
remove root check

gravitypriest:
Remove the root check in the client. Note that the GPG checks on the eggs will still fail as non-root until we change the permissions on the `redhattools.pub.gpg` file we ship in the RPM.

### DIFF
--- a/src/insights_client/__init__.py
+++ b/src/insights_client/__init__.py
@@ -163,9 +163,6 @@ def _main():
     attempt to collect and upload with new, then current, then rpm
     if an egg fails a phase never try it again
     """
-    if os.getuid() != 0:
-        sys.exit('Insights client must be run as root.')
-
     # sort rpm and stable eggs after verification
     validated_eggs = sorted_eggs(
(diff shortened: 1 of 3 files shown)


subpop (approved):
LGTM

patchkez:
test this please

ptoscano:
Not sure what's the story/idea of this; that said, I don't think it is a good idea to let `insights-client` run as non-root user:
- the biggest issue here is the different reporting of the status, e.g. when running `insights-client --status` on a system registered to RHSM and connected to Insights:
  - when run as root: it uses the the RHSM identity certificate, and it properly reports "connected"
  - when run as non-root: the POC falls back to basic authentication, and thus the status reported is "not connected"
- I'd imagine bits either in `insights-client` itself or in `insights-core`, even for the code paths other than connection/uploading/check-in, blindly assume that `insights-client` is running as root, and thus give different results

@subpop what do you think?

subpop:
Yea, this is also really old. We can close this.

PR #132, opened 2020-10-21
build: generate constants.py with SCRIPT primary

subpop:
Files listed with the `PYTHON` primary are not subject to any build time rules. The `PYTHON` primary only does stuff during installation like byte-compile the files listed[1]. What was happening was that `constants.py` was not getting generated until necessary, and it was necessary for installation. This PR adds the `nodist_insights_client_PYTHON` sources as a `SCRIPT` primary. `SCRIPT` primaries *are* generated at build time, so this ends up generating `constants.py` earlier in the process. `SCRIPT` primaries are not distributed by default, so it doesn't need to be listed as `nodist`, and it's already not distributed via `nodist_insights_client_PYTHON`.

1: https://www.gnu.org/software/automake/manual/html_node/Python.html
2: https://www.gnu.org/software/automake/manual/html_node/Scripts.html

### DIFF
--- a/src/insights_client/Makefile.am
+++ b/src/insights_client/Makefile.am
@@ -11,6 +11,10 @@ nodist_insights_client_PYTHON = \
 	constants.py \
 	$(NULL)
 
+BUILT_SOURCES = \
+	constants.py \
+	$(NULL)
+
 EXTRA_DIST = constants.py.in
 CLEANFILES = constants.py
 


gravitypriest (approved):

Glutexo:
I just wonder, why don’t we want constants.py to be distributed?

Glutexo:
I have troubles trying to reproduce this. It looksl ike the _constants.py_ file is randomly built or not built regardless of whether the change is in there. Is it possible that I am missing something? Do I have to _touch_ any files so they are picked by _configure_ and _make_?

Glutexo:
@subpop Would you please share a bit of your knowledge. How can I properly test this?

subpop:
After checking out this PR ref, run `make clean` to clean out any built/generated files. Then run `./configure` to rebuild the Makefile (from their source Makefile.am files). And then run `make` again.

TL; DR:

* `make clean`
* `autoreconf` or `./autogen.sh`
* `./configure`
* `make`
* `test -f src/insights_client/constants.py`

Glutexo (approved):
Phew. It looks that because of the clocks of my guest and host machines were constantly not in sync, Make was confused not knowing what to rebuild.

Installing VirtualBox Guest Additions fixed that as it apparently does the synchronization. To install them, kernel development headers must be installed and their version must match with the currently loaded kernel.

Also as I was not completely sure what part of the job is done by _autogen.sh_, _configure_, _make_ and _make clean_, I may have used then in a wrong way. Following your instruction made these tools do what they should and **I can confirm the bug is fixed**. ✅

I just wonder – is adding _constants.py_ to the SCRIPTS primary semantically correct? From what I read in the documentation you linked, this primary is intended for, well, scripts. _constants.py_ is not something we can run. Shouldn’t it be rather added to some other group of files that are built on _make_?

I somehow understand that the [_Python_ primary](https://www.gnu.org/software/automake/manual/html_node/Python.html) is intended for compilation of _.py_ files into _.pycs_. It looks like we are somehow abusing this. But it’s still a mystery to me why this happens on _make install_ and not on _make_. What is _make_ then good for when the binary compilation happens on _make install_?

This is probably an uneducated guess, but couldn’t _AC_CONFIG_FILES_ be the right place to put it? [automake/Scripts](https://www.gnu.org/software/automake/manual/html_node/Scripts.html) mentions it, but I don’t fully understand how it works.

subpop:
Yea, adding it as an value in the `_SCRIPTS` primary is a misuse of what are assumed to be "scripts". The `constants.py` file isn't actually *needed* until installation time either though. Running `insights-client` from the build directory isn't really an expected use case. We could add it to *AC_CONFIG_FILES* as you suggest. That is probably a more appropriate location for this directive than defining a `_SCRIPTS` primary for something that isn't a script.

subpop:
I found a much more appropriate solution to this. Thank you @Glutexo for challenging the first solution.

Glutexo:
I tried to understand this a bit more after our session. I was wondering whether it is possible to make the _constants.py_ file an explicit dependency of either _insights_client/\_\_init\_\_.py_ or _insights-client_. My guess why neither adding it as a `nodist_insights_client_SOURCE` nor as _\_\_init\_\_.py: constants.py_ didn’t work is that _\_\_init\_\_.py_ is actually not being built as it’s already there. Adding it as `insights-client: insights_client/constants.py` resulted in _insights_client/constants.py_ being generated, but without the @ variable replacements and surprisingly also with an executable flag. I have no idea why.

I know we don’t need to waste any more time on this. I am just trying to understand the mechanics. The BUILD_SOURCES solution works fine and it’s definitely more semantically correct. The only thing that bothers me is that _make src/insights_client/constants.py_ doesn’t generate it and neither does _make src/insights_client/\_\_init\_\_.py_ or _make src/insights-client_. That is however mentioned in the linked documentation.

PR #133, opened 2020-11-12
fix: print warning to stderr

subpop:
Printing the warning to stdout breaks some downstream tooling that
assumes the output of insights-client is a certain data format (such as
JSON). Instead, the warning should be printed to stderr.

Fixes: RHCLOUD-10655, RHBZ#1896547
Signed-off-by: Link Dupont <link@sub-pop.net>


gravitypriest (approved):

subpop:
test this please

PR #134, opened 2020-11-23
Check-in SystemD service

Glutexo:
Created a new service for the hourly checkin and a corresponding timer unit. This PR only introduces the units, but not a mechanism to enable them upon registration. To verify their workings, launch them manually with _systemctl_ after running _make install_. Use SystemD drop-ins to change any settings.

### DIFF
--- a/data/systemd/Makefile.am
+++ b/data/systemd/Makefile.am
@@ -1,6 +1,7 @@
 dist_systemdsystemunit_DATA = \
 	insights-client.service \
 	insights-client.timer \
+	insights-client-checkin.timer \
 	$(NULL)
 
 systemdsystemunit_DATA = \
--- a/data/systemd/Makefile.am
+++ b/data/systemd/Makefile.am
@@ -25,7 +25,7 @@ endif
 
 CLEANFILES = $(systemdsystemunit_DATA)
 EXTRA_DIST = \
-	insights-checkin.service.in \
+	insights-client-checkin.service.in \
 	insights-client-results.service.in \
 	insights-client-results.path.in \
 	insights-register.service.in \
--- a/data/systemd/Makefile.am
+++ b/data/systemd/Makefile.am
@@ -4,6 +4,7 @@ dist_systemdsystemunit_DATA = \
 	$(NULL)
 
 systemdsystemunit_DATA = \
+	insights-client-checkin.service \
 	insights-client-results.service \
 	insights-client-results.path \
 	$(NULL)
--- a/data/systemd/Makefile.am
+++ b/data/systemd/Makefile.am
@@ -25,6 +25,7 @@ endif
 
 CLEANFILES = $(systemdsystemunit_DATA)
 EXTRA_DIST = \
+	insights-checkin.service.in \
 	insights-client-results.service.in \
 	insights-client-results.path.in \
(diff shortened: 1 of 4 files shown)


Glutexo:
I need help here, @gravitypriest, @subpop. I don’t know how to make the new service installed. Running _./autogen.sh_, _make_, _make install_ did not do the trick. I am listing the services with `systemctl list-unit-files --type=service --all`. Am I doing something wrong?

subpop:
@Glutexo you need to include the generated file in the `systemdsystemunit_DATA`[1]. It looks like you only have the `.in` file in `EXTRA_DIST`.

1: https://github.com/RedHatInsights/insights-client/blob/e82eb249c02cf7ff5647685fc6d4c9ac85ef632d/data/systemd/Makefile.am#L15

Glutexo:
You were right, @subpop. Adding _insights-client-checkin.service_ to _systemdsystemunit_DATA_ really did the trick. Thank you!

Glutexo:
Confirmed the service unit working. Now, the timer.

Glutexo:
Ready for review.

Glutexo:
What about this one, @subpop and @gravitypriest? Does this need some more care, or can it be passed on to the QA?

subpop (approved):
LGTM

subpop:
Is this PR intended to include the systemd timer unit? Or is that in another PR?

Glutexo:
Oh, I see. I forgot to include the timer. 😞 That means I unintentionally lied in yesterday’s standup. Fixed that and asking for a new review. Thanks!

subpop (approved):
Looks good to me. I think my only outstanding question is around timer activation. When `insights-client --register` is run, it enables the `insights-client.timer` unit. We would either need to modify the insights-client code to also enable the `insights-client-checkin.timer`, or make `insights-client-checkin.timer` dependent on `insights-client.timer` through the systemd dependency directives.

I don't think this is an issue we need to solve now though; we can merge this PR, but not enable it (yet).

Glutexo:
Oh, I see. I didn’t realize the need for a code or dependency for this to be activated. I‘ll create a follow-up ticket for that.

Glutexo:
Created a follow-up [JIRA ticket](https://issues.redhat.com/browse/RHCLOUD-11426) for the dependency. Waiting for the QE to approve this.

Glutexo:
test this please

Glutexo:
test this please

gravitypriest (approved):

Glutexo:
Pavol from the QE says, he has some issues with this. They seem unrelated to me, but we should investigate.

Glutexo:
The QE’s concerns were caused by testing agains Production environment that does not have the check-in feature deployed yet. So this is still being tested.

Glutexo:
Update: The issues were related to my core code. This is now being fixed by https://github.com/RedHatInsights/insights-core/pull/2927.

Glutexo:
Confirmed 🆗  by QE. 🎉

Glutexo:
This is however not going to be released until a feature flag is added.

Glutexo:
The feature flag PR is here: https://github.com/RedHatInsights/insights-client/pull/140

PR #138, opened 2021-02-15
Add systemd unit file for insights-client-boot service

aleccohan:
Based on conversations with @subpop.  This PR will add a systemd unit file that will _upon declaration of a file run_insights_client_next_boot_ will run the insights client on the next boot of the host node.  This will be triggered by playbooks created by remediations that are specified with the `localhost` tag.

See Corresponding Remediations JIRA:  https://github.com/RedHatInsights/insights-remediations/pull/12

### DIFF
--- a/data/systemd/insights-client-boot.service
+++ b/data/systemd/insights-client-boot.service
@@ -24,7 +24,7 @@ TasksMax=300
 BlockIOWeight=100
 ExecStartPost=-/bin/bash -c "echo 2G >/dev/null 2>&1 > /sys/fs/cgroup/memory/system.slice/insights-client.service/memory.memsw.limit_in_bytes"
 ExecStartPost=-/bin/bash -c "echo 1G >/dev/null 2>&1 > /sys/fs/cgroup/memory/system.slice/insights-client.service/memory.soft_limit_in_bytes"
-ExecStartPost=-rm /etc/insights-client/.run_insights_client_next_boot
+ExecStartPost=rm -f /etc/insights-client/.run_insights_client_next_boot
 
 [Install]
 Wants=multi-user.target
--- a/data/systemd/insights-client-boot.service
+++ b/data/systemd/insights-client-boot.service
@@ -21,6 +21,7 @@ WatchdogSec=900
 CPUQuota=30%
 MemoryHigh=1G
 MemoryMax=2G
+MemoryLimit=2G # For compatability with RHEL7
 TasksMax=300
 BlockIOWeight=100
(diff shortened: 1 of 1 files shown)


aleccohan:
@subpop @gravitypriest If you have the time today would you be able to take a look at this?  Thanks!

aleccohan:
@subpop @gravitypriest could you take a look at this PR?  Thanks.

data/systemd/insights-client-boot.service
```diff
@@ -0,0 +1,30 @@
...
+ExecStart=/usr/bin/insights-client --retry 3
+Restart=no
+WatchdogSec=900
+CPUQuota=30%
+MemoryLimit=2G
+TasksMax=300
+BlockIOWeight=100
+ExecStartPost=-/bin/bash -c "echo 2G >/dev/null 2>&1 > /sys/fs/cgroup/memory/system.slice/insights-client.service/memory.memsw.limit_in_bytes"
+ExecStartPost=-/bin/bash -c "echo 1G >/dev/null 2>&1 > /sys/fs/cgroup/memory/system.slice/insights-client.service/memory.soft_limit_in_bytes"
+ExecStartPost=-rm /etc/insights-client/.run_insights_client_next_boot
```
Ichimonji10, line 27: Why ignore return codes on these three commands?

For this final command,  adding the `--force` flag will cause rm to "ignore nonexistent files and arguments, never prompt."

gravitypriest (approved):

data/systemd/insights-client-boot.service
```diff
@@ -0,0 +1,30 @@
...
+After=network.target
+ConditionPathExists=/etc/insights-client/.run_insights_client_next_boot
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/insights-client --retry 3
+Restart=no
+WatchdogSec=900
+CPUQuota=30%
+MemoryLimit=2G
```
Ichimonji10, line 22: `MemoryMax=` replaces `MemoryLimit=`, no need for both.

Ichimonji10:
Thanks for revising!

data/systemd/insights-client-boot.service
gravitypriest: This also needs to work in RHEL 7 -- I'm not sure whether you can just use `MemoryLimit` as an alias or whether you can define both, but we should go for max compatibility here.

Ichimonji10: Gotcha. :+1:

gravitypriest (approved):

PR #141, opened 2021-04-21
Revert "Remove egg is file condition"

gravitypriest:
Reverts RedHatInsights/insights-client#136

This breaks the egg loading flow and incorrectly throws an error on the GPG verification. We'll have to revisit this change set.


Glutexo:
Oh, I see, now.

PR #143, opened 2021-05-11
Various systemd/automake fixes

subpop:
The systemd data Makefile.am needed a bit of cleanup.

* Fixed a bug where the insights-client-checkin.service.in file was missing from the distribution tarball.
* Refactor the CLEANFILES variable to explicitly list all the files that should be cleaned up by the clean target.
* Create two preset files: one is always installed and includes insights-client-boot.service. The other is installed only if the software is configured with --enable-auto-registration, and includes the insights-register.path and insights-unregister.path units.

Fixes: RHBZ#1951750

### DIFF
--- a/data/systemd/80-insights-register.preset
+++ b/data/systemd/80-insights-register.preset
@@ -1,0 +1,2 @@
+enable insights-register.path
+enable insights-unregister.path
--- a/data/systemd/80-insights.preset
+++ b/data/systemd/80-insights.preset
@@ -1,3 +1,1 @@
-enable insights-register.path
-enable insights-unregister.path
 enable insights-client-boot.service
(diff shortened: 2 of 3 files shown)


gravitypriest (approved):

subpop:
test this please

Glutexo:
Why was the _insights-client-checkin.service.in_ file missing in the tarball? It was listed in the _EXTRA_DIST_ list.

subpop:
> Why was the _insights-client-checkin.service.in_ file missing in the tarball? It was listed in the _EXTRA_DIST_ list.

I can't remember now. :-( The diffs certainly show it being there. Maybe I broke something and thought it was always broken. Either way, the cleanup should make the Makefile.am more explicit.

PR #144, opened 2021-06-01
Make metrics usable without RHSM

gravitypriest:
Also act gracefully when

- the rhsm.conf cannot be read
- rhsm.conf _can_ be read but the cert files do not exist (i.e. system was unregistered but conf is still hanging on the disk)

### DIFF
--- a/src/insights_client/metrics.py
+++ b/src/insights_client/metrics.py
@@ -9,10 +9,13 @@ AUTH_METHOD_CERT = "CERT"
 
 
 def _proxy_settings(rhsm_config):
-    hostname = rhsm_config.get("server", "proxy_hostname").strip()
-    port = rhsm_config.get("server", "proxy_port").strip()
-    user = rhsm_config.get("server", "proxy_user").strip()
-    password = rhsm_config.get("server", "proxy_password").strip()
+    try:
+        hostname = rhsm_config.get("server", "proxy_hostname").strip()
+        port = rhsm_config.get("server", "proxy_port").strip()
+        user = rhsm_config.get("server", "proxy_user").strip()
+        password = rhsm_config.get("server", "proxy_password").strip()
+    except (configparser.NoSectionError, configparser.NoOptionError):
+        return None
 
     if not hostname:
         return None
--- a/src/insights_client/metrics.py
+++ b/src/insights_client/metrics.py
@@ -51,25 +51,34 @@ class MetricsHTTPClient(requests.Session):
         rhsm_cfg = configparser.ConfigParser()
         rhsm_cfg.read(rhsm_config_file)
 
-        rhsm_server_hostname = rhsm_cfg.get("server", "hostname")
-        rhsm_server_port = rhsm_cfg.get("server", "port")
-        rhsm_rhsm_repo_ca_cert = rhsm_cfg.get("rhsm", "repo_ca_cert")
-        rhsm_rhsm_consumerCertDir = rhsm_cfg.get("rhsm", "consumerCertDir")
+        try:
+            rhsm_server_hostname = rhsm_cfg.get("server", "hostname")
+            rhsm_server_port = rhsm_cfg.get("server", "port")
+            rhsm_rhsm_repo_ca_cert = rhsm_cfg.get("rhsm", "repo_ca_cert")
+            rhsm_rhsm_consumerCertDir = rhsm_cfg.get("rhsm", "consumerCertDir")
 
-        if cert_file is None:
-            cert_file = os.path.join(rhsm_rhsm_consumerCertDir, "cert.pem")
-        if key_file is None:
-            key_file = os.path.join(rhsm_rhsm_consumerCertDir, "key.pem")
+            if cert_file is None:
+                cert_file = os.path.join(rhsm_rhsm_consumerCertDir, "cert.pem")
+            if key_file is None:
+                key_file = os.path.join(rhsm_rhsm_consumerCertDir, "key.pem")
 
-        match = re.match("subscription.rhsm(.stage)?.redhat.com", rhsm_server_hostname)
-        if match is None:
-            # Assume Satellite-managed and configure for Satellite-proxied access
-            self.base_url = rhsm_server_hostname
-            self.port = rhsm_server_port
-            self.cert = (cert_file, key_file)
-            self.verify = rhsm_rhsm_repo_ca_cert
-            self.api_prefix = "/redhat_access/r/insights/platform"
-        else:
+            match = re.match("subscription.rhsm(.stage)?.redhat.com", rhsm_server_hostname)
+            if match is None:
+                # Assume Satellite-managed and configure for Satellite-proxied access
+                self.base_url = rhsm_server_hostname
+                self.port = rhsm_server_port
+                self.cert = (cert_file, key_file)
+                self.verify = rhsm_rhsm_repo_ca_cert
+                self.api_prefix = "/redhat_access/r/insights/platform"
+                is_satellite = True
+            else:
+                is_satellite = False
+        except (configparser.NoSectionError, configparser.NoOptionError):
+            # cannot read RHSM conf, go straight to insights-client conf parsing
+            # set is_satellite=False to enter the insights-client.conf parsing block
+            is_satellite = False
+
+        if not is_satellite:
             try:
                 auth_method = cfg.get("insights-client", "authmethod")
             except configparser.NoOptionError:
@@ -100,6 +109,7 @@ class MetricsHTTPClient(requests.Session):
             self.proxies = _proxy_settings(rhsm_cfg)
         else:
             self.proxies = None
+
 
     def post(self, event):
         """
--- a/src/insights_client/metrics.py
+++ b/src/insights_client/metrics.py
@@ -14,22 +14,19 @@ def _proxy_settings(rhsm_config):
             "user": None,
             "password": None}
 
-    try:
-        for key in conf:
-            try:
-                conf[key] = rhsm_config.get("server", "proxy_" + key).strip()
-                if key == "hostname" and conf[key] is None:
-                    # hostname is the only required value. return None if empty
-                    return None
-            except configparser.NoOptionError:
-                if key == "hostname":
-                    return None
-    except configparser.NoSectionError:
-        return None
+    for key in conf:
+        try:
+            conf[key] = rhsm_config.get("server", "proxy_" + key).strip() or None
+        except configparser.NoSectionError:
+            return None
+        except configparser.NoOptionError:
+            pass
+        if key == "hostname" and not conf[key]:
+            # hostname is the only required value. return None if empty
+            return None
 
     auth = "%s:%s@" % (conf["user"], conf["password"]) if conf["user"] and conf["password"] else ""
-    path = "%s:%s" % (conf["hostname"], conf["port"] if conf["port"] else "")
-    proxy = "http://%s%s" % (auth, path)
+    proxy = "http://%s%s:%s" % (auth, conf["hostname"], conf["port"] or "")
 
     return {"https": proxy}
 
--- a/src/insights_client/metrics.py
+++ b/src/insights_client/metrics.py
@@ -80,6 +80,8 @@ class MetricsHTTPClient(requests.Session):
                 is_satellite = True
             else:
                 is_satellite = False
+
+        if not is_satellite:
             try:
                 auth_method = cfg.get("insights-client", "authmethod")
             except configparser.NoOptionError:
--- a/src/insights_client/metrics.py
+++ b/src/insights_client/metrics.py
@@ -11,13 +11,12 @@ AUTH_METHOD_CERT = "CERT"
 def _proxy_settings(rhsm_config):
     try:
         hostname = rhsm_config.get("server", "proxy_hostname").strip()
+        if not hostname:
+            return None
         port = rhsm_config.get("server", "proxy_port").strip()
         user = rhsm_config.get("server", "proxy_user").strip()
         password = rhsm_config.get("server", "proxy_password").strip()
     except (configparser.NoSectionError, configparser.NoOptionError):
-        return None
-
-    if not hostname:
         return None
 
     auth = "%s:%s@" % (user, password) if user and password else ""
@@ -55,10 +54,10 @@ class MetricsHTTPClient(requests.Session):
         rhsm_cfg.read(rhsm_config_file)
 
         try:
-            rhsm_server_hostname = rhsm_cfg.get("server", "hostname")
-            rhsm_server_port = rhsm_cfg.get("server", "port")
-            rhsm_rhsm_repo_ca_cert = rhsm_cfg.get("rhsm", "repo_ca_cert")
-            rhsm_rhsm_consumerCertDir = rhsm_cfg.get("rhsm", "consumerCertDir")
+            rhsm_server_hostname = rhsm_cfg.get("server", "hostname").strip()
+            rhsm_server_port = rhsm_cfg.get("server", "port").strip()
+            rhsm_rhsm_repo_ca_cert = rhsm_cfg.get("rhsm", "repo_ca_cert").strip()
+            rhsm_rhsm_consumerCertDir = rhsm_cfg.get("rhsm", "consumerCertDir").strip()
         except (configparser.NoSectionError, configparser.NoOptionError):
             # cannot read RHSM conf, go straight to insights-client conf parsing
             # set is_satellite=False to enter the insights-client.conf parsing block
--- a/src/insights_client/metrics.py
+++ b/src/insights_client/metrics.py
@@ -113,7 +113,6 @@ class MetricsHTTPClient(requests.Session):
         else:
             self.proxies = None
 
-
     def post(self, event):
         """
         post sends `event` to the Insights Platform.
--- a/src/insights_client/metrics.py
+++ b/src/insights_client/metrics.py
@@ -59,7 +59,11 @@ class MetricsHTTPClient(requests.Session):
             rhsm_server_port = rhsm_cfg.get("server", "port")
             rhsm_rhsm_repo_ca_cert = rhsm_cfg.get("rhsm", "repo_ca_cert")
             rhsm_rhsm_consumerCertDir = rhsm_cfg.get("rhsm", "consumerCertDir")
-
+        except (configparser.NoSectionError, configparser.NoOptionError):
+            # cannot read RHSM conf, go straight to insights-client conf parsing
+            # set is_satellite=False to enter the insights-client.conf parsing block
+            is_satellite = False
+        else:
             if cert_file is None:
                 cert_file = os.path.join(rhsm_rhsm_consumerCertDir, "cert.pem")
             if key_file is None:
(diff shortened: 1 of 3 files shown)


subpop (approved):
LGTM

src/insights_client/metrics.py
```diff
@@ -51,25 +54,34 @@ def __init__(
...
+                self.api_prefix = "/redhat_access/r/insights/platform"
+                is_satellite = True
+            else:
+                is_satellite = False
+        except (configparser.NoSectionError, configparser.NoOptionError):
+            # cannot read RHSM conf, go straight to insights-client conf parsing
+            # set is_satellite=False to enter the insights-client.conf parsing block
+            is_satellite = False
+
+        if not is_satellite:
```
Glutexo, line 84: The exceptions can be raised only by _rhsm_cfg.get_ on the first four lines. Let’s _try_ only those and use _else_ for the rest of the code that should not run if an exception occurs.

```suggestion
        try:
            rhsm_server_hostname = rhsm_cfg.get("server", "hostname")
            rhsm_server_port = rhsm_cfg.get("server", "port")
            rhsm_rhsm_repo_ca_cert = rhsm_cfg.get("rhsm", "repo_ca_cert")
            rhsm_rhsm_consumerCertDir = rhsm_cfg.get("rhsm", "consumerCertDir")
        except (configparser.NoSectionError, configparser.NoOptionError):
            # cannot read RHSM conf, go straight to insights-client conf parsing
            # set is_satellite=False to enter the insights-client.conf parsing block
            is_satellite = False
        else:
            if cert_file is None:
                cert_file = os.path.join(rhsm_rhsm_consumerCertDir, "cert.pem")
            if key_file is None:
                key_file = os.path.join(rhsm_rhsm_consumerCertDir, "key.pem")

            match = re.match("subscription.rhsm(.stage)?.redhat.com", rhsm_server_hostname)
            if match is None:
                # Assume Satellite-managed and configure for Satellite-proxied access
                self.base_url = rhsm_server_hostname
                self.port = rhsm_server_port
                self.cert = (cert_file, key_file)
                self.verify = rhsm_rhsm_repo_ca_cert
                self.api_prefix = "/redhat_access/r/insights/platform"
                is_satellite = True
            else:
                is_satellite = False
```

src/insights_client/run.py
```diff
@@ -3,8 +3,12 @@
 
 from insights import package_info
 
+import logging
 import metrics
 import utc
+from requests import ConnectionError
```
Glutexo, line 9: Shouldn’t we keep _requests_ contained in the _metrics_ module? It’s a bit unfortunate that _metrics.MetricsHTTPClient_ is a _requests.Session_ and that it’s _post_ method does violate its parent’s signature, so we can’t really keep all the _requests_ stuff inside. But all _RequestExceptions_ including _ConnectionError_ [are](https://github.com/psf/requests/blob/1466ad713cf84738cd28f1224a7ab4a19e50e361/requests/exceptions.py#L12) _IOErrors_. So this should not be needed at all. See below.

src/insights_client/run.py
```diff
@@ -46,8 +50,8 @@
         event["ended_at"] = utc.make_utc_datetime_rfc3339()
         try:
             metrics_client.post(event)
-        except OSError as e:
-            print("Error: Could not submit event: {0}".format(e))
+        except (OSError, IOError, ConnectionError) as e:
```
Glutexo, line 53: ConnectionError [is](https://github.com/psf/requests/blob/1466ad713cf84738cd28f1224a7ab4a19e50e361/requests/exceptions.py#L12) IOError.

```suggestion
        except (OSError, IOError) as e:
```

should be enough and it would save us from importing requests here.

src/insights_client/run.py
```diff
@@ -46,8 +50,8 @@
...
+            logger.debug("Could not submit event: {0}".format(e))
```
Glutexo, line 54: Unrelated, but whatever.

Glutexo, line 54: Unrelated, but whatever.

src/insights_client/metrics.py
subpop: I believe this suggestion will also fix the case where `proxy_username` is not only set to an empty string, but also when the key is absent entirely.

subpop:
This PR does also make changes to the `_proxy_settings` function. PR #145 does too. I'm fine if you want to incorporate that diff into this PR. Including the patch to the test file, I'm pretty sure that will fix the case where keys are missing from the config file entirely.

Glutexo (commented):
Tests?

src/insights_client/metrics.py
```diff
@@ -9,10 +9,13 @@
...
+    try:
+        hostname = rhsm_config.get("server", "proxy_hostname").strip()
+        port = rhsm_config.get("server", "proxy_port").strip()
+        user = rhsm_config.get("server", "proxy_user").strip()
+        password = rhsm_config.get("server", "proxy_password").strip()
+    except (configparser.NoSectionError, configparser.NoOptionError):
+        return None
 
     if not hostname:
         return None
```
Glutexo, line 21: This can be also moved below the _hostname_ variable assignment to shortcut when we already know that we don’t need to read the rest. But that’s not important.

```suggestion
```

src/insights_client/metrics.py
```diff
@@ -9,10 +9,13 @@
 
 
 def _proxy_settings(rhsm_config):
-    hostname = rhsm_config.get("server", "proxy_hostname").strip()
-    port = rhsm_config.get("server", "proxy_port").strip()
-    user = rhsm_config.get("server", "proxy_user").strip()
-    password = rhsm_config.get("server", "proxy_password").strip()
+    try:
+        hostname = rhsm_config.get("server", "proxy_hostname").strip()
```
Glutexo, line 13:
```suggestion
        hostname = rhsm_config.get("server", "proxy_hostname").strip()
        if not hostname:
            return None        
```

src/insights_client/metrics.py
```diff
@@ -51,25 +54,34 @@ def __init__(
...
-            self.base_url = rhsm_server_hostname
-            self.port = rhsm_server_port
-            self.cert = (cert_file, key_file)
-            self.verify = rhsm_rhsm_repo_ca_cert
-            self.api_prefix = "/redhat_access/r/insights/platform"
+        try:
+            rhsm_server_hostname = rhsm_cfg.get("server", "hostname")
+            rhsm_server_port = rhsm_cfg.get("server", "port")
+            rhsm_rhsm_repo_ca_cert = rhsm_cfg.get("rhsm", "repo_ca_cert")
+            rhsm_rhsm_consumerCertDir = rhsm_cfg.get("rhsm", "consumerCertDir")
```
Glutexo, line 61: Don’t we want to _strip_ here too?

```suggestion
            rhsm_server_hostname = rhsm_cfg.get("server", "hostname").strip()
            rhsm_server_port = rhsm_cfg.get("server", "port").strip()
            rhsm_rhsm_repo_ca_cert = rhsm_cfg.get("rhsm", "repo_ca_cert").strip()
            rhsm_rhsm_consumerCertDir = rhsm_cfg.get("rhsm", "consumerCertDir").strip()
```

But that’s unrelated to this PR.

Glutexo (commented):
Confirmed working.

Glutexo:
This does not only fix the case when a key is missing, but a case when the whole section is missing too.

Glutexo:
I am not really willing to ✅  without tests. I originally added only tests for the code I actually added or changes, that is the proxy settings. I am aware that the module is now mostly untested, but we may increase the coverage by continuing with this strategy. The test changes from @subpop’s #145 can be used directly, some need to be added, but they will be very similar.

gravitypriest:
@Glutexo I brought in @subpop's unit test changes from the branch + your suggestions, and given their content, I assume we want to allow for empty/missing values for all proxy parameters other than hostname. I tweaked the proxy parsing function a bit, let me know your thoughts.

src/insights_client/tests/test_metrics.py
```diff
@@ -120,6 +117,14 @@ def test_proxy_settings_without_auth(rhsm_config_file_factory):
...
 
 
+def test_proxy_settings_with_empty_auth(rhsm_config_file_factory):
+    rhsm_config_file = rhsm_config_file_factory(proxy_hostname="localhost", proxy_port=3128, proxy_user="", proxy_password="")
+    rhsm_config = ConfigParser()
+    rhsm_config.read(rhsm_config_file.name)
+    proxy_settings = _proxy_settings(rhsm_config)
+    assert proxy_settings == {"https": "http://localhost:3128"}
+
+
```
Glutexo, line 127: There are no tests that would verify correct handling of a situation when _proxy_hostname_ key is present, but its value is empty. The first suggested tests verifies a properly empty string, the second one checks the stripping.

```suggestion

def test_proxy_settings_empty_hostname(rhsm_config_file_factory):
    rhsm_config_file = rhsm_config_file_factory(proxy_hostname="")
    rhsm_config = ConfigParser()
    rhsm_config.read(rhsm_config_file.name)
    proxy_settings = _proxy_settings(rhsm_config)
    assert proxy_settings is None


def test_proxy_settings_whitespace_hostname(rhsm_config_file_factory):
    rhsm_config_file = rhsm_config_file_factory(proxy_hostname=" ")
    rhsm_config = ConfigParser()
    rhsm_config.read(rhsm_config_file.name)
    proxy_settings = _proxy_settings(rhsm_config)
    assert proxy_settings is None


```

src/insights_client/tests/test_metrics.py
```diff
@@ -120,6 +117,14 @@ def test_proxy_settings_without_auth(rhsm_config_file_factory):
     assert proxy_settings == {"https": "http://localhost:3128"}
 
 
+def test_proxy_settings_with_empty_auth(rhsm_config_file_factory):
```
Glutexo, line 120: I think we needs tests for a whitespace values too that would verify the _strip_ calls.

src/insights_client/tests/test_metrics.py
```diff
@@ -120,6 +117,14 @@ def test_proxy_settings_without_auth(rhsm_config_file_factory):
...
+    rhsm_config_file = rhsm_config_file_factory(proxy_hostname="localhost", proxy_port=3128, proxy_user="", proxy_password="")
```
Glutexo, line 121: Line too long?

src/insights_client/tests/test_metrics.py
```diff
@@ -37,20 +38,16 @@ def factory(**config):
         config = u"""[server]
```
Glutexo, line 38: There are no tests that would verify a case when this whole _server_ section would be missing.

src/insights_client/metrics.py
```diff
@@ -9,18 +9,27 @@
...
+        for key in conf:
+            try:
+                conf[key] = rhsm_config.get("server", "proxy_" + key).strip()
+                if key == "hostname" and conf[key] is None:
+                    # hostname is the only required value. return None if empty
+                    return None
+            except configparser.NoOptionError:
+                if key == "hostname":
+                    return None
+    except configparser.NoSectionError:
```
Glutexo, line 27: This is not tested at all. All test cases have the _server_ section present.

Glutexo (changes requested):
I found a case when the file would not be parsed correctly. Also some untested cases, part of which would expose the aforementioned. Sorry. 😢

src/insights_client/metrics.py
gravitypriest: Getting there. I just wanted your thoughts on the syntax before I went all-in on it.

src/insights_client/tests/test_metrics.py
```diff
@@ -34,27 +35,92 @@ def factory(config):
...
+    metrics_client = MetricsHTTPClient(
+        config_file=config_file.name, rhsm_config_file=rhsm_config_file.name,
+        cert_file="/path/to/test/cert.pem", key_file="/path/to/test/key.pem")
+    assert not metrics_client.auth
+    assert metrics_client.cert == ("/path/to/test/cert.pem", "/path/to/test/key.pem")
+    assert metrics_client.base_url == "cert.cloud.redhat.com"
+    assert metrics_client.api_prefix == "/api"
+
+@patch("insights_client.metrics._proxy_settings")
+def test_http_client_init_is_not_satellite(proxy_settings, config_file_factory, rhsm_config_file_factory):
```
Glutexo, line 107: ❗️ This is a duplicate function name, causing the previous one to be overwritten.

```suggestion
def test_http_client_init_is_satellite(proxy_settings, config_file_factory, rhsm_config_file_factory):
```

src/insights_client/tests/test_metrics.py
```diff
@@ -34,27 +35,92 @@ def factory(config):
...
+    '''
+    config_file = config_file_factory("")
+    rhsm_config_file = rhsm_config_file_factory(
+        hostname="satellite.example.com",
+        repo_ca_cert="/path/to/cacert/cert.pem", consumerCertDir="/path/to/sat/certs")
+    metrics_client = MetricsHTTPClient(
+        config_file=config_file.name, rhsm_config_file=rhsm_config_file.name)
+    assert not metrics_client.auth
+    assert metrics_client.cert == ("/path/to/sat/certs/cert.pem", "/path/to/sat/certs/key.pem")
+    assert metrics_client.port == "443"
```
Glutexo, line 120: I am not very happy about how this is taken from the default value there in the fixture. All the other tested values are here in this test case.

src/insights_client/tests/test_metrics.py
```diff
@@ -34,27 +35,92 @@ def factory(config):
...
+%s
+""" % tuple(
+            "%s = %s" % (key, config[key]) if key in config else ""
+            for key in ("hostname", "port", "proxy_hostname", "proxy_port", "proxy_user", "proxy_password", "repo_ca_cert", "consumerCertDir"))
         return _tempfile(config)
 
     return factory
 
+@patch("insights_client.metrics._proxy_settings")
+def test_http_client_init_missing_server_section(proxy_settings, config_file_factory):
```
Glutexo, line 61: This case tests two things at the same time:

- Correct handling of a missing _server_ section.
- Recognition of the _BASIC_ authmethod.

It’s a candidate for a split.

src/insights_client/tests/test_metrics.py
```diff
@@ -34,27 +35,92 @@ def factory(config):
...
+    '''
+    Verify that when the "server" section is missing, RHSM configuration is ignored
+    and insights-client configuration is preferred
+    '''
+    config_file = config_file_factory("username=testuser\npassword=testpass\nauthmethod=BASIC")
+    rhsm_config_file = _tempfile("[nonsense]\ntest=test")
+    metrics_client = MetricsHTTPClient(config_file=config_file.name, rhsm_config_file=rhsm_config_file.name)
+    assert metrics_client.auth == ("testuser", "testpass")
+    assert not metrics_client.cert
```
Glutexo, line 70: We may also want to check that _metrics_client.verify_ is not set.

```suggestion
    assert not metrics_client.cert
    assert not metrics_client.verify
```

Glutexo (changes requested):
Picking _Request changes_ only because of the duplicate test name. All other comments just thoughts that don’t need to be taken too seriously.

gravitypriest:
Fixed dupe test name. I'd like to use your additional comments as future improvements as well.

src/insights_client/tests/test_metrics.py
gravitypriest: In this case, `Session().verify` is default to true, but I've now added a test for the value regardless.

src/insights_client/tests/test_metrics.py
```diff
@@ -34,27 +35,92 @@ def factory(config):
...
+    '''
+    Verify that when insights-client configuration is used, CERT auth is selected
+    if no authmethod is defined in the configuration file
+    '''
+    config_file = config_file_factory("username=testuser\npassword=testpass")
+    rhsm_config_file = _tempfile("[nonsense]\nauth=test")
+    metrics_client = MetricsHTTPClient(
+        config_file=config_file.name, rhsm_config_file=rhsm_config_file.name,
+        cert_file="/path/to/test/cert.pem", key_file="/path/to/test/key.pem")
+    assert not metrics_client.auth
```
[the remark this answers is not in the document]
gravitypriest: See above thing about verify

src/insights_client/tests/test_metrics.py
```diff
@@ -34,27 +35,92 @@ def factory(config):
...
+    '''
+    Verify that when the configured RHSM hostname matches one of the Red Hat subscription URLs,
+    it is determined to NOT be a Satellite-subscribed host
+    '''
+    config_file = config_file_factory("")
+    rhsm_config_file = rhsm_config_file_factory(hostname="subscription.rhsm.redhat.com")
+    metrics_client = MetricsHTTPClient(
+        config_file=config_file.name, rhsm_config_file=rhsm_config_file.name,
+        cert_file="/path/to/test/cert.pem", key_file="/path/to/test/key.pem")
+    assert not metrics_client.auth
```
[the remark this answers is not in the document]
gravitypriest: See above thing about verify

gravitypriest:
test this please

src/insights_client/tests/test_metrics.py
Glutexo: Oh.

gravitypriest:
test this please

src/insights_client/metrics.py
```diff
@@ -51,25 +59,34 @@ def __init__(
...
+                self.verify = rhsm_rhsm_repo_ca_cert
+                self.api_prefix = "/redhat_access/r/insights/platform"
+                is_satellite = True
+            else:
+                is_satellite = False
+
+        if not is_satellite:
             try:
                 auth_method = cfg.get("insights-client", "authmethod")
             except configparser.NoOptionError:
```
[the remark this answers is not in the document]
Glutexo: A similar except should probably handle missing username or password in case of a BASIC auth.

```python
                u = cfg.get("insights-client", "username")
                p = cfg.get("insights-client", "password")
```

also raises an uncaught _configparser.NoOptionError_. Do we want to handle that gracefully, just as if authmethod is not defined?

Glutexo (approved):
I think I can ✅. Tested manually and everything seems to work.

There are two ⚠️s I’d like to point out:

- Missing _username_ and _password_ in _insights-client.conf_ for _BASIC_ auth is not handled gracefully. _authmethod_ is.
- _hostname_, _port_ and certs in _rhsm.conf_ are handled inconsistently. An empty value is handled differently than a missing key.

We can address those in a follow-up PR.

src/insights_client/tests/test_metrics.py
```diff
@@ -148,14 +251,16 @@ def test_http_metrics_client_post_proxies(post, config_file_factory, rhsm_config
...
+        hostname="satellite.example.com",
+        repo_ca_cert="/path/to/cacert/cert.pem", consumerCertDir="/path/to/sat/certs")
     metrics_client = MetricsHTTPClient(config_file=config_file.name, rhsm_config_file=rhsm_config_file.name)
 
     event = Mock()
     metrics_client.post(event)
 
     post.assert_called_once_with(
-        "https://cert-api.access.redhat.com:443/redhat_access/r/insights/platform/module-update-router/v1/event",
+        "https://satellite.example.com:443/redhat_access/r/insights/platform/module-update-router/v1/event",
```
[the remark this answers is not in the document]
gravitypriest: Just to avoid confusion. These tests were evidently for testing Sat URLs and cert-api will never be a Sat URL.

gravitypriest:
I'll address the username thing in a follow-up PR. I want to add additional handling for --offline because right now the metrics client does not honor that option. I think in the case of a missing user/pass we just set it to "offline" and abandon sending metrics. At that point the client won't be able to do anything either, but it won't hurt to add additional handling here.

gravitypriest:
test this please

gravitypriest:
this this please

gravitypriest:
test this please

PR #145, opened 2021-06-07
config-fallback

subpop:
Fall back to a default value if the key is missing from the ConfigParser. These default values should still enable a properly constructed proxy object, even in the absence of keys or values in the config file.

Fixes: RHBZ#1932153

### DIFF
--- a/src/insights_client/metrics.py
+++ b/src/insights_client/metrics.py
@@ -9,10 +9,10 @@ AUTH_METHOD_CERT = "CERT"
 
 
 def _proxy_settings(rhsm_config):
-    hostname = rhsm_config.get("server", "proxy_hostname", fallback=None)
-    port = rhsm_config.get("server", "proxy_port", fallback="")
-    user = rhsm_config.get("server", "proxy_user", fallback="")
-    password = rhsm_config.get("server", "proxy_password", fallback="")
+    hostname = rhsm_config.get("server", "proxy_hostname", fallback="").strip()
+    port = rhsm_config.get("server", "proxy_port", fallback="").strip()
+    user = rhsm_config.get("server", "proxy_user", fallback="").strip()
+    password = rhsm_config.get("server", "proxy_password", fallback="").strip()
 
     if not hostname:
         return None
--- a/src/insights_client/metrics.py
+++ b/src/insights_client/metrics.py
@@ -9,10 +9,10 @@ AUTH_METHOD_CERT = "CERT"
 
 
 def _proxy_settings(rhsm_config):
-    hostname = rhsm_config.get("server", "proxy_hostname").strip()
-    port = rhsm_config.get("server", "proxy_port").strip()
-    user = rhsm_config.get("server", "proxy_user").strip()
-    password = rhsm_config.get("server", "proxy_password").strip()
+    hostname = rhsm_config.get("server", "proxy_hostname", fallback=None)
+    port = rhsm_config.get("server", "proxy_port", fallback="")
+    user = rhsm_config.get("server", "proxy_user", fallback="")
+    password = rhsm_config.get("server", "proxy_password", fallback="")
 
     if not hostname:
         return None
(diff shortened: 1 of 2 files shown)


src/insights_client/metrics.py
```diff
@@ -9,10 +9,10 @@
...
 
 def _proxy_settings(rhsm_config):
-    hostname = rhsm_config.get("server", "proxy_hostname").strip()
-    port = rhsm_config.get("server", "proxy_port").strip()
-    user = rhsm_config.get("server", "proxy_user").strip()
-    password = rhsm_config.get("server", "proxy_password").strip()
+    hostname = rhsm_config.get("server", "proxy_hostname", fallback=None)
+    port = rhsm_config.get("server", "proxy_port", fallback="")
+    user = rhsm_config.get("server", "proxy_user", fallback="")
+    password = rhsm_config.get("server", "proxy_password", fallback="")
```
Glutexo, line 15: Do we really want to get rid of the _strip_ calls? They should remove any trailing whitespace that may occur there by accident.

It wouldn’t work for _None_ fallback of hostname, but that can have an empty string fallback too. Empty strings are falsey in conditionals.

src/insights_client/tests/test_metrics.py
```diff
@@ -37,19 +37,19 @@ def factory(**config):
...
 consumerCertDir =
 """ % (
-            config.get("proxy_hostname", ""),
-            config.get("proxy_port", ""),
-            config.get("proxy_user", ""),
-            config.get("proxy_password", ""),
+            "proxy_hostname = %s" % config.get("proxy_hostname") if config.get("proxy_hostname") else "",
+            "proxy_port = %s" % config.get("proxy_port") if config.get("proxy_port") else "",
+            "proxy_user = %s" % config.get("proxy_user") if config.get("proxy_user") else "",
+            "proxy_password = %s" % config.get("proxy_password") if config.get("proxy_password") else "",
```
Glutexo, line 52: We may want to test both variants of the missing values: missing key as well as an empty value.

src/insights_client/tests/test_metrics.py
```diff
@@ -37,19 +37,19 @@ def factory(**config):
...
 consumerCertDir =
 """ % (
-            config.get("proxy_hostname", ""),
-            config.get("proxy_port", ""),
-            config.get("proxy_user", ""),
-            config.get("proxy_password", ""),
+            "proxy_hostname = %s" % config["proxy_hostname"] if "proxy_hostname" in config else "",
+            "proxy_port = %s" % config["proxy_port"] if "proxy_port" in config else "",
+            "proxy_user = %s" % config["proxy_user"] if "proxy_user" in config else "",
+            "proxy_password = %s" % config["proxy_password"] if "proxy_password" in config else "",
```
Glutexo, line 52: This too can be simplified.

```suggestion
""" % tuple(
            "%s = %s" % (key, config[key]) if key in config else ""
            for key in ("proxy_hostname", "proxy_port", "proxy_user", "proxy_password")
      )
```

Glutexo (approved):
Verified, works! Thanks! 🙇🏻‍♂️ Suggested some DRY, but that may impair readability, so not necessary.

LGTM

patchkez:
test this please

Glutexo:
If #144 gets merged before this, we may want to remove the _fallback_ kwargs, because the missing values will have been handled by a exception catching.

src/insights_client/metrics.py
```diff
@@ -9,10 +9,10 @@
 
 
 def _proxy_settings(rhsm_config):
-    hostname = rhsm_config.get("server", "proxy_hostname").strip()
-    port = rhsm_config.get("server", "proxy_port").strip()
-    user = rhsm_config.get("server", "proxy_user").strip()
-    password = rhsm_config.get("server", "proxy_password").strip()
+    hostname = rhsm_config.get("server", "proxy_hostname", fallback="").strip()
```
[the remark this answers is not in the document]
gravitypriest: Unfortunately it doesn't look like `fallback` is available in Python 2 😔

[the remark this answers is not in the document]
subpop: (╯°□°）╯︵ ┻━┻

patchkez:
test this please

Glutexo:
Fixed by #144.

PR #146, opened 2021-06-18
feat(metrics):honor offline mode, add logging

gravitypriest:
Honor the `offline` option when using metrics.

Manually parsing out offline from the various config sources here, because a bug prevents the `InsightsConfig` from loading the CLI twice in one process. Once https://github.com/RedHatInsights/insights-core/pull/3111 is merged, we can replace `_is_offline()` in this PR a call to `InsightsConfig`

In addition, I've moved the `metrics_client` initialization in `run.py` to a point after the core phase, so that logging is available after it was initialized in the egg.


gravitypriest:
test this please

gravitypriest:
test this please

src/insights_client/metrics.py
```diff
@@ -56,6 +83,13 @@ def __init__(
         cfg = configparser.RawConfigParser()
         cfg.read(config_file)
 
+        # TODO: switch to this once the core PR is merged
+        # self.offline = InsightsConfig().load_all().offline
+        self.offline = _is_offline(cfg)
+        if self.offline:
+            logger.debug("Metrics: Offline mode. No metrics will be sent.")
+            return
+
```
Glutexo, line 92: It this the place where we want to do it? I can imagine extracting the decision to a function and then using a dummy class instead. But that would require extract the configuration file loading too, which may not be that bad after all.

```python
def _load_configs(config_file="/etc/insights-client/insights-client.conf", rhsm_config_file="/etc/rhsm/rhsm.conf")
    cfg = configparser.RawConfigParser()
    cfg.read(config_file)

    rhsm_cfg = configparser.ConfigParser()
    rhsm_cfg.read(rhsm_config_file)
    
    return cfg, rhsm_cfg

def get_metrics_client():
    cfg, rhsm_cfg = _load_configs()
    if _is_offline(cfg):
        return OfflineMetricsClient()
    else
        return HTTPMetricsClient(config=cfg, rhsm_config=rhsm_cfg)

class OfflineMetricsClient:
    def __init__():
        logger.debug("Metrics: Offline mode. No metrics will be sent.")

    def post(self, event):
        pass
```

Yes, I see that it would result in a RHSM config file being loaded and parsed even in the offline mode.

But as I see it further, there are more cases when the function falls back to being offline. That makes me think more about extracting the configuration. We probably don’t want to do it here in this pull request.

src/insights_client/metrics.py
```diff
@@ -90,13 +126,19 @@ def __init__(
...
                 self.port = 443
-                u = cfg.get("insights-client", "username")
-                p = cfg.get("insights-client", "password")
+                try:
+                    u = cfg.get("insights-client", "username")
+                    p = cfg.get("insights-client", "password")
+                except configparser.NoOptionError:
+                    logger.debug("Metrics: BASIC auth selected but username and/or password is missing. No metrics will be sent.")
+                    self.offline = True
+                    return
```
Glutexo, line 141: If we are adding this, I think the same should happen for an unknown auth_method.

```python
if auth_method == AUTH_METHOD_BASIC:
    ...
elif auth_method == AUTH_METHOD_CERT:
    ...
else:
    logger.debug("Metrics: Unknown authentication method. No metrics will be sent.")
    self.offline = True

File: src/insights_client/metrics.py [hunk 3]
```diff
@@ -90,13 +126,19 @@ def __init__(
...
                 auth_method = AUTH_METHOD_CERT
 
             if auth_method == AUTH_METHOD_BASIC:
                 self.base_url = "cloud.redhat.com"
                 self.port = 443
-                u = cfg.get("insights-client", "username")
-                p = cfg.get("insights-client", "password")
+                try:
+                    u = cfg.get("insights-client", "username")
+                    p = cfg.get("insights-client", "password")
```
Glutexo, line 137: Shouldn’t we be consistent with other configuration options and strip the values? I know that this may be tricky for credentials. We really don’t want to block out users that really have whitespace at the end of their passwords.

```suggestion
                    u = cfg.get("insights-client", "username").strip()
                    p = cfg.get("insights-client", "password").strip()
```

src/insights_client/metrics.py
```diff
@@ -90,13 +126,19 @@ def __init__(
...
+                except configparser.NoOptionError:
+                    logger.debug("Metrics: BASIC auth selected but username and/or password is missing. No metrics will be sent.")
+                    self.offline = True
+                    return
                 self.auth = (u, p)
```
Glutexo, line 142:
```suggestion
                if not u or not p:
                    logger.debug("Metrics: BASIC auth selected but username and/or password is missing. No metrics will be sent.")
                    self.offline = True
                    return

                self.auth = (u, p)
```

src/insights_client/metrics.py
```diff
@@ -124,7 +166,13 @@ def post(self, event):
...
+        if self.offline:
+            return
         url = "https://{0}:{1}{2}/module-update-router/v1/event".format(
             self.base_url, self.port, self.api_prefix
         )
-        return super(MetricsHTTPClient, self).post(url, json=event, proxies=self.proxies)
+        logger.debug("Metrics: Posting event...")
+        logger.log(NETWORK, "POST %s", url)
+        res = super(MetricsHTTPClient, self).post(url, json=event, proxies=self.proxies)
+        logger.log(NETWORK, "HTTP Status Code: %d", res.status_code)
```
Glutexo, line 177:
```suggestion
        logger.log(NETWORK, "HTTP Status Code: %d %s", res.status_code, res.reason)
```

Glutexo (commented):
This pull request solves rather too many things at the same time:

- Offline configuration option.
- Metrics logging.
- Missing username/password handling.
- Refactoring, although a small one.
- Makefile changes. May be necessary for tests.

I’d rather see those split into more atomic pull requests.

Gave some comments, but in general it looks good. Looking into the tests now and going to manually verify the working.

src/insights_client/metrics.py
```diff
@@ -31,6 +37,27 @@ def _proxy_settings(rhsm_config):
...
+                insights-client config file read into it
+
+    Returns True if offline, False if not offline
+    '''
+    # look in config file first
+    try:
+        offline = cfg.getboolean("insights-client", "offline")
+    except configparser.NoOptionError:
+        offline = False
+    if "--offline" in sys.argv:
```
Glutexo, line 55: Is it this simple? Shouldn’t we use argparser for consistency? I mean, is it possible that _--offline_ somehow appears in _argv_, but it is parsed differently? E.g. as a value to a different argument.

src/insights_client/metrics.py
```diff
@@ -31,6 +37,27 @@ def _proxy_settings(rhsm_config):
...
+        offline = True
+    if os.getenv("INSIGHTS_OFFLINE") == "True":
```
Glutexo, line 57: _config.load_env_ is case insensitive.

Glutexo:
Found a real [inconsistency](https://github.com/RedHatInsights/insights-client/pull/146#discussion_r656983883) with _config.load_all_.

src/insights_client/metrics.py
```diff
@@ -31,6 +37,27 @@ def _proxy_settings(rhsm_config):
...
+        offline = True
```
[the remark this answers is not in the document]
Glutexo: ⚠️ _config.load_all_ [parses](https://github.com/RedHatInsights/insights-core/blob/e2ffbf447a491bee7b85385f0029c2d2a0a6d41d/insights/client/config.py#L648) the environment variables prior to the command line arguments. That means that if _INSIGHTS_OFFLINE=False_ support was implemented, it couldn’t be overridden by _--offline_. This is inconsistent.

```suggestion
    if os.getenv("INSIGHTS_OFFLINE") == "True":
        offline = True
    if "--offline" in sys.argv:
        offline = True
```

Glutexo (changes requested):
Okay, there are more inconsistencies, so requesting changes in the end.

src/insights_client/metrics.py
```diff
@@ -1,9 +1,15 @@
 import os.path
 import re
-
+import sys
+import logging
 import requests
 from six.moves import configparser
+from insights.client.constants import InsightsConstants
+# from insights.client.config import InsightsConfig
```
Glutexo, line 8: Is it safe to import from the core? I remember @subpop requesting to find a way to import from the bundled egg. The justification was that the metrics should work even if there are bugs in the automatically updated core.

src/insights_client/metrics.py
Glutexo:
```suggestion
    parser = ArgumentParser()
    parser.add_argument('--offline', default=False, action="store_true")
    args = parser.parse_args()
    if args.offline:
        offline = True
```

src/insights_client/metrics.py
gravitypriest: Removing this dependency.

src/insights_client/metrics.py
```diff
@@ -31,6 +35,43 @@ def _proxy_settings(rhsm_config):
...
+
+    Returns True if offline, False if not offline
+    '''
+    def _boolify(v):
+        if v.lower() == 'true':
+            return True
+        elif v.lower() == 'false':
+            return False
+        else:
+            return v
```
Glutexo, line 54: This can live outside the __is_offline function_, so it is not redefined with every call.

```suggestion
```

src/insights_client/metrics.py
```diff
@@ -31,6 +35,43 @@ def _proxy_settings(rhsm_config):
     return {"https": proxy}
```
Glutexo, line 37:
```suggestion

def _boolify(v):
    if v.lower() == 'true':
        return True
    elif v.lower() == 'false':
        return False
    else:
        return v


```

src/insights_client/metrics.py
```diff
@@ -31,6 +35,43 @@ def _proxy_settings(rhsm_config):
...
+
+    # get from env
+    env_offline = os.getenv("INSIGHTS_OFFLINE")
+    if env_offline:
+        offline = _boolify(env_offline)
+
+    # parse from CLI
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--offline', action='store_true')
+    args, _ = parser.parse_known_args()
```
Glutexo, line 69: Oh, it’s _parse_known_args_! Great, I couldn’t find it. 👍🏻

src/insights_client/metrics.py
```diff
@@ -90,22 +138,37 @@ def __init__(
...
                 self.port = 443
-                u = cfg.get("insights-client", "username")
-                p = cfg.get("insights-client", "password")
+                try:
+                    u = cfg.get("insights-client", "username").strip()
+                    p = cfg.get("insights-client", "password").strip()
+                except configparser.NoOptionError:
+                    logger.debug("Metrics: BASIC auth selected but username and/or password is missing. No metrics will be sent.")
+                    self.offline = True
+                    return
```
Glutexo, line 153:
```suggestion
                    u, p = None, None
```

Glutexo (commented):
I just couldn’t get the logger to work. Whatever I do, set _loglevel_ in the config file, use _--verbose_, still the _metrics_ logger has a _WARNING_ level.

Glutexo:
Confirmed argument parsing works.

gravitypriest:
Re: logging, are you running with the egg available on your PYTHONPATH? It's a bit clunky but we rely on logging to be set up as part of the phase.

Glutexo (approved):
I tested it manually and it works now consistently with the config module. I could not get the logging working, but that’s not critical. I am still going to take a look at the tests, but if you need it merged, :shipit:.

Glutexo:
Remaining work for @Glutexo:
- [ ] Make the logging work.
- [ ] Review the tests.

I still think we’d better extract the configuration parsing and possibly refactor whatever we can, before the fairly new metrics module start to rot.

gravitypriest:
Agree. We have enough tech debt as it is... I'd happily revisit this once the release is completed.

src/insights_client/tests/test_metrics.py
```diff
@@ -285,3 +285,130 @@ def test_metrics_post_event_proxy(post, config_file_factory, rhsm_config_file_fa
         proxies={"https": "http://user:password@localhost:3128"},
     )
 
+@patch("insights_client.metrics.os.getenv")
+def test_is_offline(os_getenv, config_file_factory):
```
Glutexo, line 289: Pytest has _monkeypatch_ that can be used to patch both the environment and the args.

```suggestion
def test_is_offline(monkeypatch, config_file_factory):
```

src/insights_client/tests/test_metrics.py
```diff
@@ -285,3 +285,130 @@ def test_metrics_post_event_proxy(post, config_file_factory, rhsm_config_file_fa
...
+
+    NOTE: this test will become obsolete once InsightsConfig is used to load config
+    '''
+    # offline not specified
+    config_file = config_file_factory("")
+    cfg = configparser.RawConfigParser()
+    cfg.read(config_file.name)
+    os_getenv.return_value = None
+    with patch('sys.argv', ['insights-client']):
+        assert not _is_offline(cfg)
```
Glutexo, line 301:
```suggestion
    monkeypatch.delenv("INSIGHTS_OFFLINE", raising=False)
    monkeypatch.setattr("sys.argv", ["insights-client"])
    assert not _is_offline(cfg)
```

src/insights_client/tests/test_metrics.py
```diff
@@ -285,3 +285,130 @@ def test_metrics_post_event_proxy(post, config_file_factory, rhsm_config_file_fa
...
+    # offline specified in config
+    config_file = config_file_factory("offline=True\n")
+    cfg = configparser.RawConfigParser()
+    cfg.read(config_file.name)
+    os_getenv.return_value = None
+    with patch('sys.argv', ['insights-client']):
+        assert _is_offline(cfg)
```
Glutexo, line 308:
```suggestion
    monkeypatch.delenv("INSIGHTS_OFFLINE", raising=False)
    monkeypatch.setattr("sys.argv", ["insights-client"])
    assert _is_offline(cfg)
```

src/insights_client/tests/test_metrics.py
```diff
@@ -285,3 +285,130 @@ def test_metrics_post_event_proxy(post, config_file_factory, rhsm_config_file_fa
...
+    with patch('sys.argv', ['insights-client', '--offline']):
+        assert _is_offline(cfg)
+    # offline specified in env
+    config_file = config_file_factory("")
+    cfg = configparser.RawConfigParser()
+    cfg.read(config_file.name)
+    os_getenv.return_value = "True"
+    with patch('sys.argv', ['insights-client']):
+        assert _is_offline(cfg)
+
```
Glutexo, line 323:
```suggestion
    monkeypatch.setenv("INSIGHTS_OFFLINE", "True")
    monkeypatch.setattr("sys.argv", ["insights-client"])
    assert _is_offline(cfg)

```

src/insights_client/tests/test_metrics.py
```diff
@@ -285,3 +285,130 @@ def test_metrics_post_event_proxy(post, config_file_factory, rhsm_config_file_fa
...
+    os_getenv.return_value = None
+    with patch('sys.argv', ['insights-client']):
+        assert _is_offline(cfg)
+    # offline specified in CLI
+    config_file = config_file_factory("")
+    cfg = configparser.RawConfigParser()
+    cfg.read(config_file.name)
+    os_getenv.return_value = None
+    with patch('sys.argv', ['insights-client', '--offline']):
+        assert _is_offline(cfg)
```
Glutexo, line 315:
```suggestion
    monkeypatch.delenv("INSIGHTS_OFFLINE", raising=False)
    monkeypatch.setattr("sys.argv", ["insights-client", "--offline"])
    assert _is_offline(cfg)
```

src/insights_client/tests/test_metrics.py
```diff
@@ -285,3 +285,130 @@ def test_metrics_post_event_proxy(post, config_file_factory, rhsm_config_file_fa
...
+    # offline specified in env
+    config_file = config_file_factory("")
+    cfg = configparser.RawConfigParser()
+    cfg.read(config_file.name)
+    os_getenv.return_value = "True"
+    with patch('sys.argv', ['insights-client']):
+        assert _is_offline(cfg)
+
+@patch("insights_client.metrics._is_offline", Mock(return_value=True))
```
Glutexo, line 324: Parametrization example.

```suggestion
@mark.parametrize(("config", "delenv", "setenv", "sysargv", "is_offline"), (
    ("", ("INSIGHTS_OFFLINE"), (), ["insights-client"], False),
    ("offline=True\n", ("INSIGHTS_OFFLINE"), (), ["insights-client"], True),
    ("", ("INSIGHTS_OFFLINE"), (), ["insights-client", "--offline"], True),
    ("", ("INSIGHTS_OFFLINE"), (("INSIGHTS_OFFLINE", "True"),), ["insights-client"], True),
))
def test_is_offline(monkeypatch, config, delenv, setenv, sysargv, is_offline, config_file_factory):
    '''
    Verify that _is_offline() produces correct output from the given configurations

    NOTE: this test will become obsolete once InsightsConfig is used to load config
    '''
    # offline not specified
    for d in delenv:
        monkeypatch.delenv(d, raising=False)
    for k, v in setenv:
        monkeypatch.setenv(k, v)
    monkeypatch.setattr("sys.argv", sysargv)

    config_file = config_file_factory(config)
    cfg = configparser.RawConfigParser()
    cfg.read(config_file.name)
    assert _is_offline(cfg) is is_offline

@patch("insights_client.metrics._is_offline", Mock(return_value=True))
```

Test split example.

```suggestion
def test_is_offline_default(monkeypatch, config_file_factory):
    config_file = config_file_factory("")
    cfg = configparser.RawConfigParser()
    cfg.read(config_file.name)
    monkeypatch.delenv("INSIGHTS_OFFLINE", raising=False)
    monkeypatch.setattr("sys.argv", ['insights-client'])
    assert not _is_offline(cfg)

def test_is_offline_config(monkeypatch, config_file_factory):
    config_file = config_file_factory("offline=True\n")
    cfg = configparser.RawConfigParser()
    cfg.read(config_file.name)
    monkeypatch.delenv("INSIGHTS_OFFLINE", raising=False)
    monkeypatch.setattr("sys.argv", ["insights-client"])
    assert _is_offline(cfg)

def test_is_offline_arg(monkeypatch, config_file_factory):
    config_file = config_file_factory("")
    cfg = configparser.RawConfigParser()
    cfg.read(config_file.name)
    monkeypatch.delenv("INSIGHTS_OFFLINE", raising=False)
    monkeypatch.setattr("sys.argv", ["insights-client", "--offline"])
    assert _is_offline(cfg)

def test_is_offline_env(monkeypatch, config_file_factory):
    config_file = config_file_factory("")
    cfg = configparser.RawConfigParser()
    cfg.read(config_file.name)
    monkeypatch.setenv("INSIGHTS_OFFLINE", "True")
    monkeypatch.setattr("sys.argv", ["insights-client"])
    assert _is_offline(cfg)

@patch("insights_client.metrics._is_offline", Mock(return_value=True))
```

I am more in favor of the parametrization, although it is a bit more complex, because then it would be simple to add more test cases. I mean those that would include overrides.

In one of the previous iterations I spotted an inconsistency (a bug) with the overrides, so it’s already happened. Also, the _strip_ and _lower_ calls as well as non-True/False strings are currently untested and with parametrization, adding those would become really easy.

src/insights_client/tests/test_metrics.py
```diff
@@ -285,3 +285,130 @@ def test_metrics_post_event_proxy(post, config_file_factory, rhsm_config_file_fa
         proxies={"https": "http://user:password@localhost:3128"},
     )
```
Glutexo, line 287: The RawConfigParser initialization is repeated too many times. I know this doesn’t have to be a fixture, just doing so for consistency with the other factories. It may not be necessary if the _\_is_offline_ test would be made parametrized.

```suggestion

@fixture
def raw_config_parser_factory(config_file_factory):
    def factory(config):
        config_file = config_file_factory(config)
        parser = configparser.RawConfigParser()
        parser.read(config_file.name)
        return parser
    return factory
```

src/insights_client/tests/test_metrics.py
```diff
@@ -285,3 +285,130 @@ def test_metrics_post_event_proxy(post, config_file_factory, rhsm_config_file_fa
...
+def test_is_offline(os_getenv, config_file_factory):
+    '''
+    Verify that _is_offline() produces correct output from the given configurations
+
+    NOTE: this test will become obsolete once InsightsConfig is used to load config
+    '''
+    # offline not specified
+    config_file = config_file_factory("")
+    cfg = configparser.RawConfigParser()
+    cfg.read(config_file.name)
```
Glutexo, line 298:
```suggestion
    cfg = raw_config_file_factory("")
```

src/insights_client/tests/test_metrics.py
Glutexo:
```suggestion
@patch("insights_client.metrics.os.getenv")
def test_is_offline(os_getenv, raw_config_file_factory):
```

src/insights_client/tests/test_metrics.py
```diff
@@ -285,3 +285,130 @@ def test_metrics_post_event_proxy(post, config_file_factory, rhsm_config_file_fa
...
+    os_getenv.return_value = None
+    with patch('sys.argv', ['insights-client']):
+        assert not _is_offline(cfg)
+    # offline specified in config
+    config_file = config_file_factory("offline=True\n")
+    cfg = configparser.RawConfigParser()
+    cfg.read(config_file.name)
```
Glutexo, line 305:
```suggestion
    cfg = raw_config_file_factory("offline=True\n")
```

src/insights_client/tests/test_metrics.py
```diff
@@ -285,3 +285,130 @@ def test_metrics_post_event_proxy(post, config_file_factory, rhsm_config_file_fa
...
+    os_getenv.return_value = None
+    with patch('sys.argv', ['insights-client']):
+        assert _is_offline(cfg)
+    # offline specified in CLI
+    config_file = config_file_factory("")
+    cfg = configparser.RawConfigParser()
+    cfg.read(config_file.name)
```
Glutexo, line 312:
```suggestion
    cfg = config_file_factory("")
```

src/insights_client/tests/test_metrics.py
```diff
@@ -285,3 +285,130 @@ def test_metrics_post_event_proxy(post, config_file_factory, rhsm_config_file_fa
...
+    os_getenv.return_value = None
+    with patch('sys.argv', ['insights-client', '--offline']):
+        assert _is_offline(cfg)
+    # offline specified in env
+    config_file = config_file_factory("")
+    cfg = configparser.RawConfigParser()
+    cfg.read(config_file.name)
```
Glutexo, line 319:
```suggestion
    cfg = config_file_factory("")
```

gravitypriest:
test this please

Glutexo:
Awesome! 🚀

src/insights_client/metrics.py
```diff
@@ -31,6 +35,52 @@ def _proxy_settings(rhsm_config):
...
 
 
+def _boolify(v):
+    if v.lower() == 'true':
+        return True
+    elif v.lower() == 'false':
+        return False
+    else:
+        return v
+
```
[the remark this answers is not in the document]
Glutexo: Oh. There is `not it` too. But it should be equivalent.

```suggestion
    return v.lower() not in ("false", ""):

```

src/insights_client/tests/test_metrics.py
```diff
@@ -285,3 +285,115 @@ def test_metrics_post_event_proxy(post, config_file_factory, rhsm_config_file_fa
...
+    Verify that when the metrics client is set to offline, no further initialization is done
+    '''
+    config_file = config_file_factory("")
+    rhsm_config_file = rhsm_config_file_factory()
+    m = MetricsHTTPClient(config_file=config_file.name, rhsm_config_file=rhsm_config_file.name)
+    assert m.offline
+    assert not hasattr(m, 'base_url')
+    assert not hasattr(m, 'port')
+    assert not m.cert
+    assert m.verify
```
Glutexo, line 322: If we want to test this is the unchanged default value not overwritten by _rhsm_rhsm_repo_ca_cert_.

```suggestion
    assert m.verify is True
```

src/insights_client/tests/test_metrics.py
```diff
@@ -285,3 +285,115 @@ def test_metrics_post_event_proxy(post, config_file_factory, rhsm_config_file_fa
...
+
+@patch("insights_client.metrics._is_offline", Mock(return_value=True))
+@patch("insights_client.metrics._proxy_settings")
+def test_offline_no_init(_proxy_settings, config_file_factory, rhsm_config_file_factory):
+    '''
+    Verify that when the metrics client is set to offline, no further initialization is done
+    '''
+    config_file = config_file_factory("")
+    rhsm_config_file = rhsm_config_file_factory()
+    m = MetricsHTTPClient(config_file=config_file.name, rhsm_config_file=rhsm_config_file.name)
```
Glutexo, line 317:
```suggestion
@patch("insights_client.metrics.configparser")
@patch("insights_client.metrics._is_offline", Mock(return_value=True))
@patch("insights_client.metrics._proxy_settings")
def test_offline_no_init(_proxy_settings, configparser, config_file_factory, rhsm_config_file_factory):
    '''
    Verify that when the metrics client is set to offline, no further initialization is done
    '''
    config_file = config_file_factory("")
    rhsm_config_file = rhsm_config_file_factory()
    m = MetricsHTTPClient(config_file=config_file.name, rhsm_config_file=rhsm_config_file.name)
    configparser.ConfigParser.assert_not_called()
```

src/insights_client/tests/test_metrics.py
```diff
@@ -285,3 +285,115 @@ def test_metrics_post_event_proxy(post, config_file_factory, rhsm_config_file_fa
...
+    '''
+    Verify that if no username is provided for BASIC auth, metrics are not sent (default to offline)
+    '''
+    config_file = config_file_factory("password=testpass\nauthmethod=BASIC")
+    rhsm_config_file = rhsm_config_file_factory()
+    metrics_client = MetricsHTTPClient(
+        config_file=config_file.name, rhsm_config_file=rhsm_config_file.name)
+    assert metrics_client.offline
+
+@patch("insights_client.metrics._is_offline", Mock(return_value=False))
```
Glutexo, line 340: Dtto.

```suggestion
@patch("insights_client.metrics._is_offline", return_value=False)
```

src/insights_client/tests/test_metrics.py
```diff
@@ -285,3 +285,115 @@ def test_metrics_post_event_proxy(post, config_file_factory, rhsm_config_file_fa
...
+@patch("insights_client.metrics._proxy_settings")
+def test_offline_basic_auth_no_pass(_proxy_settings, config_file_factory, rhsm_config_file_factory):
```
Glutexo, line 342:
```suggestion
def test_offline_basic_auth_no_pass(_proxy_settings, is_offline, config_file_factory, rhsm_config_file_factory):
```

src/insights_client/tests/test_metrics.py
```diff
@@ -285,3 +285,115 @@ def test_metrics_post_event_proxy(post, config_file_factory, rhsm_config_file_fa
...
+    '''
+    Verify that if no password is provided for BASIC auth, metrics are not sent (default to offline)
+    '''
+    config_file = config_file_factory("username=testuser\nauthmethod=BASIC")
+    rhsm_config_file = rhsm_config_file_factory()
+    metrics_client = MetricsHTTPClient(
+        config_file=config_file.name, rhsm_config_file=rhsm_config_file.name)
+    assert metrics_client.offline
+
+@patch("insights_client.metrics._is_offline", Mock(return_value=False))
```
Glutexo, line 352:
```suggestion
@patch("insights_client.metrics._is_offline", return_value=False)
```

src/insights_client/tests/test_metrics.py
```diff
@@ -285,3 +285,115 @@ def test_metrics_post_event_proxy(post, config_file_factory, rhsm_config_file_fa
...
+@patch("insights_client.metrics._proxy_settings")
+def test_offline_basic_auth_empty_user(_proxy_settings, config_file_factory, rhsm_config_file_factory):
```
Glutexo, line 354:
```suggestion
def test_offline_basic_auth_empty_user(_proxy_settings, is_offline, config_file_factory, rhsm_config_file_factory):
```

src/insights_client/tests/test_metrics.py
```diff
@@ -285,3 +285,115 @@ def test_metrics_post_event_proxy(post, config_file_factory, rhsm_config_file_fa
...
+    '''
+    Verify that if an empty username is provided for BASIC auth, metrics are not sent (default to offline)
+    '''
+    config_file = config_file_factory("username=\npassword=testpass\nauthmethod=BASIC")
+    rhsm_config_file = rhsm_config_file_factory()
+    metrics_client = MetricsHTTPClient(
+        config_file=config_file.name, rhsm_config_file=rhsm_config_file.name)
+    assert metrics_client.offline
+
+@patch("insights_client.metrics._is_offline", Mock(return_value=False))
```
Glutexo, line 364:
```suggestion
@patch("insights_client.metrics._is_offline", return_value=False)
```

src/insights_client/tests/test_metrics.py
```diff
@@ -285,3 +285,115 @@ def test_metrics_post_event_proxy(post, config_file_factory, rhsm_config_file_fa
...
+@patch("insights_client.metrics._proxy_settings")
+def test_offline_basic_auth_empty_pass(_proxy_settings, config_file_factory, rhsm_config_file_factory):
```
Glutexo, line 366:
```suggestion
def test_offline_basic_auth_empty_pass(_proxy_settings, is_offline, config_file_factory, rhsm_config_file_factory):
```

src/insights_client/tests/test_metrics.py
```diff
@@ -285,3 +285,115 @@ def test_metrics_post_event_proxy(post, config_file_factory, rhsm_config_file_fa
...
+    '''
+    Verify that if an empty password is provided for BASIC auth, metrics are not sent (default to offline)
+    '''
+    config_file = config_file_factory("username=testuser\npassword=\nauthmethod=BASIC")
+    rhsm_config_file = rhsm_config_file_factory()
+    metrics_client = MetricsHTTPClient(
+        config_file=config_file.name, rhsm_config_file=rhsm_config_file.name)
+    assert metrics_client.offline
+
+@patch("insights_client.metrics._is_offline", Mock(return_value=False))
```
Glutexo, line 376:
```suggestion
@patch("insights_client.metrics._is_offline", return_value=False)
```

src/insights_client/tests/test_metrics.py
```diff
@@ -285,3 +285,115 @@ def test_metrics_post_event_proxy(post, config_file_factory, rhsm_config_file_fa
...
+@patch("insights_client.metrics._proxy_settings")
+def test_offline_basic_auth_stripping(_proxy_settings, config_file_factory, rhsm_config_file_factory):
```
Glutexo, line 378:
```suggestion
def test_offline_basic_auth_stripping(_proxy_settings, is_offline, config_file_factory, rhsm_config_file_factory):
```

src/insights_client/tests/test_metrics.py
```diff
@@ -285,3 +285,115 @@ def test_metrics_post_event_proxy(post, config_file_factory, rhsm_config_file_fa
         proxies={"https": "http://user:password@localhost:3128"},
     )
 
+@mark.parametrize(("config", "delenv", "setenv", "sysargv", "is_offline"), (
+    ("", ("INSIGHTS_OFFLINE"), (), ["insights-client"], False),
+    ("offline=True\n", ("INSIGHTS_OFFLINE"), (), ["insights-client"], True),
+    ("", ("INSIGHTS_OFFLINE"), (), ["insights-client", "--offline"], True),
+    ("", ("INSIGHTS_OFFLINE"), (("INSIGHTS_OFFLINE", "True"),), ["insights-client"], True),
+))
```
Glutexo, line 293:
```suggestion
    ("offline=True\n", (), (("INSIGHTS_OFFLINE", "False"),), ["insights-client"], False),
    ("offline=False\n", (), (("INSIGHTS_OFFLINE", "False"),), ["insights-client", "--offline"], True),
))
```

- Verify that it is possible to override the config file by the environment value.
- Verify that the command line argument trumps all.

Glutexo:
Okay, logging mysteriously started working. I don’t know if it’s related to the _PYTHONPATH_.

src/insights_client/tests/Makefile.am
```diff
@@ -1,15 +1,17 @@
 
 nodist_check_SCRIPTS = test_sed.py
+check_SCRIPTS = test_metrics.py \
```
Glutexo, line 3: SCRIPTS are [not distributed](https://www.gnu.org/software/automake/manual/html_node/Scripts.html) by default. The existing _test_sed.py.in_ being listed in EXTRA_DIST, I assume we want to distribute them, so they can be run by _make check_.

```suggestion
dist_check_SCRIPTS = test_metrics.py \
```

Glutexo (commented):
`make dist` does not include _test_metrics.py_.

PR #148, opened 2021-06-28
🩹 Make metrics tests distributed

Glutexo:
_SCRIPTS_ are not distributed by default. We want to distribute tests though. Added _dist__ to the test list, so they are included.

Follow-up to #146.


Glutexo:
@subpop Is this the right way to do it?

subpop (approved):
Yes! Looks great.

PR #150, opened 2021-07-09
Require egg to be a file only if GPG is enabled

Glutexo:
It is only necessary for an egg to be a file if GPG verification is enabled. Without it, there is no verification and adding a folder instead works just as fine. This makes development and debugging easier.

Better attempt on #136. Should fix problems mentioned in the revert #141.

Can @gravitypriest confirm?


Glutexo:
Moved to GitLab. https://gitlab.cee.redhat.com/insights-platform/insights-client/-/merge_requests/8

PR #152, opened 2023-11-09
feat: add Packit integration

ptoscano:
Add a basic configuration for building PRs using Packit, using EL 8/9 and supported Fedora's.

The actions are needed to invoke "rpkg" manually, which is what Copr does.


nforro (approved):
The config is clean and straighforward, I think we could/should use it as an example in Packit documentation :slightly_smiling_face:

subpop (approved):
This looks good to me too.

PR #154, opened 2023-11-24
fix: Do not create /root/.gnupg/ directory by accident

m-horky:
* Card ID: CCT-131
* Card ID: RHEL-2480
* Card ID: RHEL-2482

GPG creates a directory $HOME/.gnupg/ every time it performs some operation. When run under root, but not manually (e.g. via subscription-manager Cockpit plugin), it tries to create and write to this directory, which causes SELinux denials.

This patch utilizes the `--homedir` argument GPG supports in order to move the home directory to a temporary directory for the time of the transaction. After the GPG action is performed, the directory is cleaned up.

This is a companion PR to https://github.com/RedHatInsights/insights-core/pull/3930 and they should be reviewed together.

### DIFF
--- a/src/insights_client/__init__.py
+++ b/src/insights_client/__init__.py
@@ -4,17 +4,18 @@
 """
 from __future__ import print_function
 import os
+import shutil
 import sys
+import tempfile
+
 import six
 import subprocess
 from subprocess import Popen, PIPE
 from distutils.version import LooseVersion
-import shlex
 import logging
 import logging.handlers
 
 GPG_KEY = "/etc/insights-client/redhattools.pub.gpg"
-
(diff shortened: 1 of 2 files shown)


src/insights_client/__init__.py
```diff
@@ -4,18 +4,21 @@
...
 import subprocess
 from subprocess import Popen, PIPE
 from distutils.version import LooseVersion
-import shlex
 import logging
 import logging.handlers
 
 GPG_KEY = "/etc/insights-client/redhattools.pub.gpg"
-
+EXTRA_GPG_KEY = os.environ.get("EXTRA_GPG_KEY")
```
ahitacat, line 19: The addition of this extra gpg key should be part of the effort of CCT-270?

ahitacat (approved):
LGTM

PR #158, opened 2023-12-08
feat: basic integration testing with packit + testing-farm + tmt

ptoscano:
Introduce a minimal setup for integration tests of insights-client.
The tests are written using pytest and an helper plugin for 
insights-client and other Red Hat client tools [1].

The tests are wrapped as single "integration" test for tmt [2], which 
makes it possible to run them automatically e.g. in testing-farm.

Extend the packit configuration to trigger tests on testing-farm after
the completion of the build jobs.

The tests are split for the different distributions supported; this is
due to the current limitation of testing-farm in the way additional
repositories can be supplied: as the name of a Fedora Copr repository
cannot be specified, each distribution needs to get the repository URL
for it.

Manually specify the RHEL targets/composes due to the limitations in the
packit<->testing-farm mapping of RHEL composes.

It looks like gnupg as available in Fedora 39 and newer does not
validate anymore signatures in case their key is not part of the current
keyring (which is what insights-client does).

Since the way both insights-core and insights-client validate GPG
signatures is being reworked/improved [3][4], add a crude workaround on
Fedora 39+.

[1] https://github.com/ptoscano/pytest-client-tools/
[2] https://github.com/teemtee/tmt/
[3] https://github.com/RedHatInsights/insights-core/pull/3930
[4] https://github.com/RedHatInsights/insights-client/pull/154


ahitacat (approved):
LGTM

PR #159, opened 2023-12-28
feat: add subpackage for ROS and config file

upadhyeammit:
This add the subpackage ros and separate configuration file ros.conf with paramter ros_collect to enable the ROS data collection.

Considering there could be two ways to do this and unsure about what can be accepted there is another pr https://github.com/RedHatInsights/insights-client/pull/160


upadhyeammit:
The logs link do not redirect anywhere, maybe I am not authorized? I tried to install package using enabling repository on Fedora however I can't find the package from the repository

ptoscano:
> The logs link do not redirect anywhere, maybe I am not authorized?

The logs are there, and lead to https://copr.fedorainfracloud.org/coprs/packit/RedHatInsights-insights-client-159/build/6823609/

The build fails because the spec file is now not correct:
```
error: source 0 defined multiple times
```
which is correct, since there is already `Source:` defined (unversioned, so it means 0).

So rename `SOURCE0` to `Source1` (changing its usage accordingly), and move it together with `Source:`, so they are grouped together.

Also, please rebase on top of the current `master` branch, so it gets also the changes that were done since you created this PR.

subpop:
This spec file is only used to build CI/CD RPMs. This won't accomplish shipping the changes in RHEL directly. Additional changes to the spec files hosted in dist-git will be needed to include the content in RHEL. Additionally, RHEL does not automatically include subpackages; they need to go through a [review process](https://one.redhat.com/rhel-development-guide/#proc_shipping-subpackages_assembly_managing-RHEL-content) very similar to new packages added to RHEL.

ptoscano:
This still does not work:
```
error: Bad file: out/ros.conf: No such file or directory
```
My guess here is that `SourceN` must be file names, which RPM will fetch from a certain location for the sources. So move `ros.conf` directly in the same directory of the spec file.

Also, style note: please use the same style for naming that field, so `Source1` instead of `SOURCE1`.

upadhyeammit:
Hello @ptoscano and @subpop!

I had discussion with @xiangce who is handling the further changes related to this https://github.com/RedHatInsights/insights-core/pull/3979 .  Accordingly he prefers to have ros_collect in insights-client.conf only.

If its okay for you as well to go with https://github.com/RedHatInsights/insights-client/pull/160 then I would close this pull request.

ptoscano:
ACK, closing as requested.

PR #160, opened 2023-12-28
feat: add subpackage for ROS

upadhyeammit:
This add the subpackage ros and config paramter ros_collect to insights-client.conf file to enable the ROS data collection.

Considering there could be two ways to do this and unsure about what can be accepted there is another pr https://github.com/RedHatInsights/insights-client/pull/159

### DIFF
--- a/data/insights-client.conf
+++ b/data/insights-client.conf
@@ -54,3 +54,5 @@
 
 # Location of the tags file for this system
 #tags_file=/etc/insights-client/tags.yaml
+
+#ros_collect=True
--- a/data/insights-client.conf
+++ b/data/insights-client.conf
@@ -54,5 +54,3 @@
 
 # Location of the tags file for this system
 #tags_file=/etc/insights-client/tags.yaml
-
-#ros_collect=True
--- a/insights-client.spec
+++ b/insights-client.spec
@@ -43,6 +43,16 @@ BuildRequires: systemd-rpm-macros
 %description
 Sends insightful information to Red Hat for automated analysis
 
+%package ros
+Requires: pcp-zeroconf
+Summary: The subpackage for Resource Optimization Service
+
+%description ros
+
+The ros subpackage add ros_collect configuration parameter to insights-client.conf file,
+the parameter is set to True by default. The system starts sending PCP archives to
+Resource Optimization service upon modifying ros_collect parameter to True.
+
 %prep
 {{{ git_dir_setup_macro }}}
 
@@ -67,6 +77,16 @@ if [ -d %{_sysconfdir}/motd.d ]; then
     fi
 fi
 
+%post ros
+echo
+echo "Removing custom PCP configuration required for Resource Optimization service!"
+echo
+rm -f /var/lib/pcp/config/pmlogger/config.ros
+sed -i "/PCP_LOG_DIR\/pmlogger\/ros/d" /etc/pcp/pmlogger/control.d/local
+echo
+echo "Enabling ros_collect to send PCP archives for Resource Optimization service!"
+sed -i "s/#ros_collect=True/ros_collect=True/" %{_sysconfdir}/insights-client/insights-client.conf
+
 %preun
 %systemd_preun %{name}.timer
 %systemd_preun %{name}.service
@@ -76,6 +96,12 @@ fi
 %systemd_postun %{name}.timer
 %systemd_postun %{name}.service
 %systemd_postun insights-client-boot.service
+
+%postun ros
+echo
+echo "Disabling the ros_collect configuration of Resource Optimization service!"
+sed -i "s/ros_collect=True/#ros_collect=True/" %{_sysconfdir}/insights-client/insights-client.conf
+
 
 # Clean up files created by insights-client that are unowned by the RPM
 if [ $1 -eq 0 ]; then
--- a/insights-client.spec
+++ b/insights-client.spec
@@ -45,7 +45,7 @@ Sends insightful information to Red Hat for automated analysis
 
 %package ros
 Requires: pcp-zeroconf
-Summary: The subpackage for Resource Optimization Service
+Summary: The subpackage for Insights resource optimization service
 
 %description ros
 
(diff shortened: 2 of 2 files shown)


ptoscano:
What's the difference between #159 and this?

upadhyeammit:
> What's the difference between #159 and this?

The #159 has separate file to store the ros_collect parameter. In this pull request the ros_collect resides in insights-client.conf file. 

I wish to go with #159, however if that way is not accepted then I have created this pr.

ptoscano:
Well, it depends on what exactly you need to do, and both PRs are very much scarce of details/explanations/references...

data/insights-client.conf
```diff
@@ -54,3 +54,5 @@
 
 # Location of the tags file for this system
 #tags_file=/etc/insights-client/tags.yaml
+
+#ros_collect=True
```
subpop, line 58: If we're going to include a default suggested value in the config file, it should mirror the actual internal value of this config option. Is this additional data collection really enabled by default?

Also, if we're going to include this in the default config file, it needs documentation explaining what it does. And lastly, it will need an entry in `insights-client.conf.5`, again defining what the option does.

insights-client.spec
```diff
@@ -43,6 +43,16 @@ BuildRequires: systemd-rpm-macros
 %description
 Sends insightful information to Red Hat for automated analysis
 
+%package ros
+Requires: pcp-zeroconf
+Summary: The subpackage for Resource Optimization Service
```
subpop, line 48: This should probably declare are relation to Red Hat Insights somehow. Maybe:
```suggestion
Summary: Insights Resource Optimization Service
```

insights-client.spec
```diff
@@ -43,6 +43,16 @@ BuildRequires: systemd-rpm-macros
...
+
+%description ros
+
+The ros subpackage add ros_collect configuration parameter to insights-client.conf file,
+the parameter is set to True by default. The system starts sending PCP archives to
+Resource Optimization service upon modifying ros_collect parameter to True.
```
subpop, line 54: This isn't accurate. The subpackage does not add a configuration value to insights-client.conf; it is present in the config file even if this subpackage is not installed.

insights-client.spec
```diff
@@ -67,6 +77,16 @@ if [ -d %{_sysconfdir}/motd.d ]; then
     fi
 fi
 
+%post ros
+echo
+echo "Removing custom PCP configuration required for Resource Optimization service!"
+echo
```
subpop, line 83: I think it's discouraged to echo output during scriptlet execution. Do you have any example precedent that shows that echoing  output to stdout is accepted behavior?

insights-client.spec
```diff
@@ -67,6 +77,16 @@ if [ -d %{_sysconfdir}/motd.d ]; then
...
+rm -f /var/lib/pcp/config/pmlogger/config.ros
+sed -i "/PCP_LOG_DIR\/pmlogger\/ros/d" /etc/pcp/pmlogger/control.d/local
+echo
+echo "Enabling ros_collect to send PCP archives for Resource Optimization service!"
+sed -i "s/#ros_collect=True/ros_collect=True/" %{_sysconfdir}/insights-client/insights-client.conf
```
subpop, line 88: Since the config file is a use-editable file, care needs to be taken when automatically editing it. If a diligent user read about this new config file option and before updating their RPM, they add "ros_collect=False" to their configuration file. When this scriptlet runs, this sed line will fail; the file `/etc/insights-client/insights-client.conf` already exists on the user's system from the previously installed RPM. Because this file is a `%config` file, it will not be overwritten when the RPM is upgraded. This scriptlet will not successfully enable the `ros_collect` value because the string "`#ros_collect=True`" is not present in the file.

subpop (changes requested):

insights-client.spec
pgarciaq: Branding rules mandate only "Insights" should be capitalized:

```suggestion
Summary: The subpackage for Insights resource optimization service
```

insights-client.spec
pgarciaq: If the default in insights-core will be ros_collect=False, we won't need to run `` may not need to run `sed` at all but just to add this to the original/template (I don't know what the right term is) insights-client.conf file and users will later add it to set it to true if they so want.
@subpop what's your advice?

subpop: I think the approach that works best for most tooling like this is to append the desired value to the file, flanked by marker lines:

```suggestion
echo >> /etc/insights-client/insights-client.conf <EOF
### Begin insights-client-ros ###
ros_collect=True
### End insights-client-ros ###
EOF
```

It's more wordy, but it explicitly informs the user where this automatic editing of their configuration files came from. It also allows for more targeted removal, should you wish to delete the line upon removal.

```
sed -i 's/### Begin insights-client-ros ###\nros_collect=True\n### End insights-client-ros ###/d' /etc/insights-client/insights-client.conf
```

I think this might handle cases where the user had manually modified the value on their own too. You could also consider adding some conditionality to the scriptlet, where it will only make this modification if the file does not already contain a line beginning with `ros_collect`.

upadhyeammit: Makes sense!

insights-client.spec
upadhyeammit: I don't have any example, however I was neither able spot any guide/doc stating `echo` is discouraged/not allowed. I can remove that, though I wanted to have something for users to know what is happening.

insights-client.spec
upadhyeammit: Agree, updating it

insights-client.spec
upadhyeammit: Okay I can see below scenarios,

1. User already enabled the ros with ros_collect=True . In this case we don't need anything to happen as he already opted in for that.
2. User already disabled the ros with ros_collect=False. In this case as well we should not enable it forcefully.
3. It can be True or False however commented. We are only enabling it if its #ros_collect=True.

The current sed takes care of above scenarios. Considering your suggestion to make it more wordy if I understand it means,

1. Remove the default commented config from insights-client.conf
2. Add the param if its(not commented)  not there.
3. What should be correct action if its commented and either True or False? I think we should not update it and enable in this situation? because in current sed we enable it if its #ros_collect=True

data/insights-client.conf
upadhyeammit: As per new approach the param is not included in file by default. Marking this resolved accordingly.

insights-client.spec
upadhyeammit: I kept this as is as now it adds the param.

insights-client.spec
subpop: > Remove the default commented config from insights-client.conf

I think that this configuration flag is obscure enough that it does not warrant a commented out default value. We don't have anything related to malware detection or compliance in the default values/comments in insights-client.conf, so I don't think that ROS should be there either.

> Add the param if its(not commented) not there.

Typically, when automatically modifying configuration files that are also expected to be edited manually by users, great care should be taken not to overwrite or otherwise overrule the user modifications. If an uncommented line beginning with `ros_collect` is already present in the file, no automatic action should take place; the user has clearly already indicated their preference through their explicit action editing the file. So your scriptlet should only take action when no explicit value is already set in the file.

> What should be correct action if its commented and either True or False? I think we should not update it and enable in this situation? because in current sed we enable it if its #ros_collect=True and only not handling #ros_collect=False situation.

If the default state of the configuration file is to *not* have this value present, the presence of a commented "ros_collect" line again means user interaction; ignore it.

This is why using a sentinel-marked comment block makes for a clear indication of what a script is doing to a configuration file. Your scriptlets need only interact with the lines between (inclusively) your sentinel comments.

upadhyeammit: Thanks for explanations. So  request to review latest commit which has the modifications accordingly. Thank you!

subpop (approved):

PR #161, opened 2024-02-02
build: require subscription-manager

ptoscano:
Since the BASIC authentication is deprecated, explicitly require `subscription-manager` so the CERT authentication can be used (since `auto_config` is enabled by default).


ahitacat (approved):
LGTM

PR #162, opened 2024-02-05
Easier maintenance and debugging

m-horky:
This set of commits makes it easier to read the code and to debug any problems that may exist during runtime. See individual commit descriptions for details.

This PR should not be introducing any behavioral changes.

### DIFF
--- a/src/insights_client/__init__.py
+++ b/src/insights_client/__init__.py
@@ -3,28 +3,30 @@
  Red Hat Insights
 """
 from __future__ import print_function
+
+import logging
 import os
 import shutil
+import subprocess
 import sys
 import tempfile
 
 import six
-import subprocess
-from subprocess import Popen, PIPE
 from distutils.version import LooseVersion
-import logging
-import logging.handlers
 
-GPG_KEY = "/etc/insights-client/redhattools.pub.gpg"
+
 BYPASS_GPG = os.environ.get("BYPASS_GPG", "").lower() == "true"
+GPG_KEY = "/etc/insights-client/redhattools.pub.gpg"
+
 ENV_EGG = os.environ.get("EGG")
 NEW_EGG = "/var/lib/insights/newest.egg"
-REGISTERED_FILE = "/etc/insights-client/.registered"
-UNREGISTERED_FILE = "/etc/insights-client/.unregistered"
 STABLE_EGG = "/var/lib/insights/last_stable.egg"
 RPM_EGG = "/etc/insights-client/rpm.egg"
-MOTD_FILE = "/etc/motd.d/insights-client"
+
 MOTD_SRC = "/etc/insights-client/insights-client.motd"
+MOTD_FILE = "/etc/motd.d/insights-client"
+REGISTERED_FILE = "/etc/insights-client/.registered"
+UNREGISTERED_FILE = "/etc/insights-client/.unregistered"
 
 TEMPORARY_GPG_HOME_PARENT_DIRECTORY = "/var/lib/insights/"
 
--- a/src/insights_client/__init__.py
+++ b/src/insights_client/__init__.py
@@ -15,6 +15,9 @@ import six
 from distutils.version import LooseVersion
 
 
+INSIGHTS_DEBUG = os.environ.get("INSIGHTS_DEBUG", "").lower() == "true"
+NO_COLOR = os.environ.get("NO_COLOR") is not None
+
 BYPASS_GPG = os.environ.get("BYPASS_GPG", "").lower() == "true"
 GPG_KEY = "/etc/insights-client/redhattools.pub.gpg"
 
@@ -32,6 +35,28 @@ TEMPORARY_GPG_HOME_PARENT_DIRECTORY = "/var/lib/insights/"
 
 
 logger = logging.getLogger(__name__)
+
+
+def client_debug(message):
+    """Debug a log message when logging isn't available yet.
+
+    Set 'INSIGHTS_DEBUG' variable to enable this method.
+
+    :param message: Text to display
+    :type message: str
+    """
+    if not INSIGHTS_DEBUG:
+        return
+
+    prefix = "insights-client debug"
+    suffix = ""
+    if NO_COLOR or not sys.stdout.isatty():
+        prefix += ":"
+    else:
+        prefix = "\033[40m" + prefix + "\033[0m\033[33m"
+        suffix = "\033[0m"
+
+    print(prefix + " " + message + suffix, file=sys.stderr)
 
 
 def log(msg):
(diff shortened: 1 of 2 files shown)


src/insights_client/__init__.py
```diff
@@ -159,32 +159,49 @@ def run_phase(phase, client, validated_eggs):
...
+        ))
+        return
+
+    if os.path.exists(REGISTERED_FILE) or os.path.exists(UNREGISTERED_FILE):
+        # .registered/.unregistered exist, MOTD should not be displayed
+        if os.path.exists(MOTD_FILE):
+            try:
+                os.symlink(os.devnull, MOTD_FILE)
+                logger.debug("pointed the MOTD file '{path}' to '/dev/null'".format(path=MOTD_FILE))
+            except Exception:
```
ptoscano, line 186: I'd restrict the exception caught to `OSError`: the old code caught that too, and I don't think that `os.symlink()` should raise anything else than that

src/insights_client/__init__.py
```diff
@@ -159,32 +159,49 @@ def run_phase(phase, client, validated_eggs):
...
+
+    if not os.path.exists(REGISTERED_FILE) and not os.path.exists(UNREGISTERED_FILE):
+        # .registered & .unregistered do not exist, MOTD should be displayed
+        if not os.path.exists(MOTD_FILE):
+            try:
+                os.symlink(MOTD_SRC, MOTD_FILE)
+                logger.debug("pointed the MOTD file '{source}' to '{motd}'".format(
+                    source=MOTD_SRC, motd=MOTD_FILE,
+                ))
+            except Exception:
```
ptoscano, line 198: ditto, let's catch `OSError` here, rather than a very broad one

src/insights_client/__init__.py
```diff
@@ -237,9 +237,14 @@ def _main():
...
-            print("Client: %s" % constants.version)
-            print("Core: %s" % InsightsClient().version())
+            try:
+                from insights_client.constants import InsightsConstants
+            except ImportError:
+                # The source file is build from 'constants.py.in' and is not available during development
+                class InsightsConstants(object):
+                    version = "upstream"
+            print("Client: %s" % InsightsConstants.version)
+            print("Core:   %s" % InsightsClient().version())
```
ptoscano, line 247: I'd avoid changing the formatting of this output; for weird as it is, there might be something somewhere parsing this...

src/insights_client/__init__.py
```diff
@@ -237,9 +237,14 @@ def _main():
...
         if config["version"]:
-            from insights_client.constants import InsightsConstants as constants
-            print("Client: %s" % constants.version)
-            print("Core: %s" % InsightsClient().version())
+            try:
+                from insights_client.constants import InsightsConstants
+            except ImportError:
+                # The source file is build from 'constants.py.in' and is not available during development
+                class InsightsConstants(object):
+                    version = "upstream"
```
ptoscano, line 245: nitpick: more than "upstream" I'd say "development": you will get the same also when trying to run an old version that is not necessarily "upstream"

ptoscano (commented):
I had a quick look, and it seems OK from that (although it needs a more in-depth review); nevertheless, I added few notes.

m-horky:
Discussion questions:
- `feat: Stop sorting eggs`: When package gets updated, the `rpm.egg` may be newer than the `last_stable.egg`. How do we handle that? Do we let the last stable update to the latest one without using the RPM one, despite being newer?
  - If we want to continue sorting eggs, we should vendorize the `distutils.version` from the latest Python version that included it.
- `feat: Add client_debug() function`: Is there a better way to do this? Having two loggers (one for client, one for core; or even one separate for collectors) would have been nice, but the Core replaces the root logger, messing things up.

subpop:
> * `feat: Stop sorting eggs`: When package gets updated, the `rpm.egg` may be newer than the `last_stable.egg`. How do we handle that? Do we let the last stable update to the latest one without using the RPM one, despite being newer?
>   
>   * If we want to continue sorting eggs, we should vendorize the `distutils.version` from the latest Python version that included it.

I don't think it's worth trying to sort the eggs by version and use the newest versioned egg. We should always try `newest.egg`, `last_stable.egg` and then `rpm.egg`. The rollback logic when a phase fails should follow that order too. Even if the version of `rpm.egg` is newer than `last_stable.egg`, it makes things easier to follow the priority order of `newest.egg` if it exists, `last_stable.egg` if it exists, and then `rpm.egg`.

src/insights_client/__init__.py
```diff
@@ -32,6 +35,28 @@
...
+
+    :param message: Text to display
+    :type message: str
+    """
+    if not INSIGHTS_DEBUG:
+        return
+
+    prefix = "insights-client debug"
+    suffix = ""
+    if NO_COLOR:
```
ptoscano, line 48: colors ought to be disabled also when the output is not a tty (eg a file)

.gitignore
```diff
@@ -0,0 +1,3 @@
+**/__pycache__/
+**/.idea/
+**/.vscode/
```
ptoscano, line 3: since those two are only found in the top-level of the sources, their entries can be simplified as
```suggestion
/.idea/
/.vscode/
```

.gitignore
```diff
@@ -0,0 +1,3 @@
+**/__pycache__/
```
ptoscano, line 1: we likely want to ignore this subdirectory anywhere in the sources, so
```suggestion
__pycache__/
```

ptoscano (commented):
Few minor comments added; most likely, we better need #172 (and in general running unit tests in CI) to be done before this

.gitignore
ptoscano: it works fine also without; see the documentation [1]:

> A leading "\*\*" followed by a slash means match in all directories. For example, "\*\*/foo" matches file or directory "foo" anywhere, **the same as pattern "foo"**.

(emphasis mine) :)

[1] https://git-scm.com/docs/gitignore

.gitignore
```diff
@@ -0,0 +1,3 @@
+__pycache__/
+.idea/
+.vscode/
```
subpop, line 3: While it's harmless to populate a project-local .gitignore file with content like this, you'll find yourself duplicating the patterns in every project you use an IDE with. Rather than repeating the same patterns in each project you touch, I suggest creating a global gitignore file that excludes patterns like these. [This blog post](https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/) does a good job describing the steps. This way you can ignore `.idea` or `.vscode` in *every* project you touch.

m-horky:
I've moved the egg sorting commit into #179.

subpop (approved):

ptoscano (approved):
LGTM, thanks!

PR #163, opened 2024-02-05
Fix location of ros %postun scriptlet

subpop:
It was inserted halfway between the main package %postun scriptlet by mistake.


subpop:
I didn't notice this when accepting #160.

ptoscano (approved):

PR #164, opened 2024-02-07
Conditionally build ros subpackage

subpop:
Only build the ros subpackage when rpmbuild is invoked with the "--with ros" flag.

### DIFF
--- a/insights-client.spec
+++ b/insights-client.spec
@@ -1,5 +1,9 @@
 %define _binaries_in_noarch_packages_terminate_build 0
 
+# This conditional build macro adds a "--with ros" commandline option to
+# rpmbuild. The default behavior is to build without it.
+%bcond_with ros
+
 Name:                   insights-client
 Summary:                Uploads Insights information to Red Hat on a periodic basis
 Version:                {{{ git_dir_version lead=3.2 }}}
@@ -45,6 +49,7 @@ BuildRequires: systemd-rpm-macros
 %description
 Sends insightful information to Red Hat for automated analysis
 
+%if %{with ros}
 %package ros
 Requires: pcp-zeroconf
 Summary: The subpackage for Insights resource optimization service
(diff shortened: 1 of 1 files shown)


insights-client.spec
```diff
@@ -1,5 +1,9 @@
 %define _binaries_in_noarch_packages_terminate_build 0
 
+# This conditional build macro adds a "--with ros" commandline option to
+rpmbuild. The default behavior is to build without it.
```
ptoscano, line 4: missing comment character here, which breaks the build

ptoscano (changes requested):
Other than the spec syntax issue, there's the "condtionally" typo in both the PR title and the commit message.

insights-client.spec
```diff
@@ -54,6 +59,7 @@ Summary: The subpackage for Insights resource optimization service
 The ros subpackage add ros_collect configuration parameter to insights-client.conf file,
 the parameter is set to True by default. The system starts sending PCP archives to
 Resource Optimization service upon modifying ros_collect parameter to True.
+%{endif}
```
ptoscano, line 62:
```suggestion
%endif
```

insights-client.spec
```diff
@@ -90,6 +97,7 @@ ros_collect=True
 ### End insights-client-ros ###
 EOF
 fi
+%{endif}
```
ptoscano, line 100:
```suggestion
%endif
```

insights-client.spec
```diff
@@ -112,9 +120,10 @@ if [ $1 -eq 0 ]; then
     rm -f %{_sysconfdir}/insights-client/.*.etag
 fi
 
+%if %{with ros}
 %postun ros
 sed -i '/### Begin insights-client-ros ###/,/### End insights-client-ros ###/d;/ros_collect=True/d' %{_sysconfdir}/insights-client/insights-client.conf
-
+%{endif}
```
ptoscano, line 126:
```suggestion
%endif
```

subpop:
I fixed the typos and added an empty `%files ros` section, which I discovered is required in order to actually create the RPM.

PR #167, opened 2024-02-16
fix(gpg): Ensure GPG can write into its temporary home directory

m-horky:
Existing SELinux rules prohibit GPG from accessing arbitrary directories, including those in /tmp/. Both tools can read/write in /var/lib/insights/ because this is where we store the eggs.


ptoscano:
Ah, the problem is that the upstream packaging does not ship all the empty directories in `/var`, and thus the change here refers to a directory that does not exist. Even with only a simple `insights-client --version`, the current functional testing infrastructure shows good results \o/

#168 hopefully fixes the packaging, so this MR needs to wait for it.

ptoscano (approved):
LGTM, thanks!

PR #168, opened 2024-02-16
build: update %files from the RHEL packaging

ptoscano:
Ensure that the upstream packaging has all the files & directories as shipped in the RHEL packaging; in particular, ship all the empty directories in `/var`.

Also, reorder a couple of other entries to match the RHEL spec.

### DIFF
--- a/insights-client.spec
+++ b/insights-client.spec
@@ -73,6 +73,12 @@ Resource Optimization service upon modifying ros_collect parameter to True.
 %install
 %{meson_install}
 
+# Create different insights directories in /var
+mkdir -p %{buildroot}%{_localstatedir}/log/insights-client/
+mkdir -p %{buildroot}%{_localstatedir}/lib/insights/
+mkdir -p %{buildroot}%{_localstatedir}/cache/insights/
+mkdir -p %{buildroot}%{_localstatedir}/cache/insights-client/
+
 %post
 %systemd_post %{name}.timer
 if [ -d %{_sysconfdir}/motd.d ]; then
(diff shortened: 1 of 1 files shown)


insights-client.spec
```diff
@@ -133,11 +139,15 @@ sed -i '/### Begin insights-client-ros ###/,/### End insights-client-ros ###/d;/
 %{_sysconfdir}/insights-client/rpm.egg*
 %{_bindir}/*
 %{_unitdir}/*
-%{_presetdir}/*
 %attr(444,root,root) %{_sysconfdir}/insights-client/*.pem
 %attr(444,root,root) %{_sysconfdir}/insights-client/redhattools.pub.gpg
-%{_defaultdocdir}/%{name}
 %{python3_sitelib}/insights_client/
+%{_defaultdocdir}/%{name}
```
jirihnidek, line 145: Shouldn't we also try to create this directory: `%{_defaultdocdir}/%{name}` in spec file?

jirihnidek (changes requested):
All changes make sense, but I have two questions/suggestions.

insights-client.spec
ptoscano: AFAIK, rpm creates it automatically

insights-client.spec
```diff
@@ -133,11 +139,15 @@ sed -i '/### Begin insights-client-ros ###/,/### End insights-client-ros ###/d;/
...
+%{_presetdir}/*.preset
```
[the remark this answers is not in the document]
ptoscano: I think it should be fine; it isn't like anything else is in the sources, so even a generic wildcard has almost no risk of mispackaging things.

PR #169, opened 2024-02-20
Fix the grep command of ros postscript

upadhyeammit:
I ended up forming the grep cmd out of new file which was just having/not having ros_collect. That resulted in completely different outcome. Below are the details about proposed fix,

The new script example,
~~~
# cat test.sh 
if ! grep -q "^ros_collect" /etc/insights-client/insights-client.conf; then
cat <<EOF >> /etc/insights-client/insights-client.conf
### Begin insights-client-ros ###
ros_collect=True
EOF
fi
~~~

1. When ros_collect already present in file, it does not append config one more time,
~~~
# grep ros_collect /etc/insights-client/insights-client.conf -A 1 -B 1
### Begin insights-client-ros ###
ros_collect=True

# sh test.sh 
# grep ros_collect /etc/insights-client/insights-client.conf -A 1 -B 1
### Begin insights-client-ros ###
ros_collect=True
~~~

2. When ros_collect is not there it adds the config,
~~~
# grep ros_collect /etc/insights-client/insights-client.conf -A 1 -B 1
[no output]
# sh test.sh 
# grep ros_collect /etc/insights-client/insights-client.conf -A 1 -B 1
### Begin insights-client-ros ###
ros_collect=True
~~~

3. Likewise the false param stays False,
~~~
# grep ros_collect /etc/insights-client/insights-client.conf -A 1 -B 1
ros_collect=False

# sh test.sh 

# grep ros_collect /etc/insights-client/insights-client.conf -A 1 -B 1
ros_collect=False
~~~

### DIFF
--- a/insights-client.spec
+++ b/insights-client.spec
@@ -90,7 +90,7 @@ fi
 rm -f /var/lib/pcp/config/pmlogger/config.ros
 sed -i "/PCP_LOG_DIR\/pmlogger\/ros/d" /etc/pcp/pmlogger/control.d/local
 
-if grep -qv "^ros_collect" %{_sysconfdir}/insights-client/insights-client.conf; then
+if ! grep -q "^ros_collect" %{_sysconfdir}/insights-client/insights-client.conf; then
 cat <<EOF >> %{_sysconfdir}/insights-client/insights-client.conf
 ### Begin insights-client-ros ###
 ros_collect=True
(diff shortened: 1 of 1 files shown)


upadhyeammit:
@ptoscano I have added one more commit to this pr which adds insights-client package as dependency

PR #170, opened 2024-02-21
Tests: Added tests for insights-client --status option

Archana-PandeyM:


ptoscano (changes requested):
Thanks for the PR, which is a good start. I left some inline comments.

-----

I find a mismatch in the actual output messages:
- the ones that in the current PR are for "not legacy_upoad", I get them for "legacy_upoad"
- the ones for "legacy_upoad" are slightly different

-----

Regarding logging: pytest-client-tools already logs all the commands executed by the various classes (including `InsightsClient`), and they are logged with a DEBUG priority; you can get them by tweaking the log level for pytest:
```bash
# pytest --log-level=DEBUG
```
and you can find them in the `test.log` file for each test in the `artifacts/` subdirectory.

Maybe they should be changed (in pytest-client-tools) to INFO (i.e. the default of pytest)? 

-----

The `InsightsClient` class provides an helper `is_registered` property; in these two tests, maybe it would be better to also check that that propery returns the right results; i.e. in `test_status_registered_uploading`:
```python
assert insights_client.is_registered
```
This way, in other tests we can use the property directly.

-----

Thanks!

integration-tests/conftest.py
```diff
@@ -0,0 +1,10 @@
+import pytest
+
+
+@pytest.fixture(scope="session", autouse=True)
+def register_subman(external_candlepin, subman_session, test_config):
```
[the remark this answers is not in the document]
Archana-PandeyM: I still would like consider the option to use auto use at once for all tests. In my understanding multiple files would need to run the tests on a registered system that will cause unnecessary registration/unregistration. and most importantly a test like test_version or man page tests would not suffer from a system being registered already. And a big positive would be eliminating multiple subman register/unregister process. Your thoughts??

integration-tests/test_status.py
```diff
@@ -0,0 +1,48 @@
...
+    reg_status = insights_client.register()
+    logger.info(f"reg status is: {reg_status}")
+    time.sleep(5)
+    registration_status = insights_client.run("--status")
+    logger.info(f"status is : {registration_status}")
+    if insights_client.config.legacy_upload:
+        assert registration_status.returncode == 0
+        assert (
+            b"Insights API confirms registration." in registration_status.stdout
+        ), fail_message
```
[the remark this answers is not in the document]
Archana-PandeyM: Ok. Thanks!

integration-tests/conftest.py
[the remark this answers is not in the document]
ptoscano: Ah, that should not be an issue: since your new fixture has a "session" scope, even if it is used in multiple places it is instantiated only once. So having the marker I suggested in multiple files will not have a negative side effect.

integration-tests/test_status.py
```diff
@@ -0,0 +1,42 @@
...
+from time import sleep
+
+pytestmark = pytest.mark.usefixtures("register_subman")
+
+
+def test_status_registered(external_candlepin, insights_client):
+    """
+    Test `insights-client --status` when the client is registered;
+    in case of legacy_upload status command returns a different output.
+
```
ptoscano, line 12: nitpick: extra empty line

integration-tests/test_status.py
```diff
@@ -0,0 +1,42 @@
...
+        assert b"Insights API confirms registration." in registration_status.stdout
+    else:
+        assert b"This host is registered.\n" == registration_status.stdout
+
+
+def test_status_unregistered(external_candlepin, insights_client):
+    """
+    Test `insights-client --status` when the client is unregistered;
+    in case of legacy_upload status command returns a different output.
+
```
ptoscano, line 30: nitpick: extra empty line

integration-tests/test_status.py
```diff
@@ -0,0 +1,42 @@
...
+    """
+    insights_client.register()
+    assert insights_client.is_registered
+    # Adding a small wait to ensure inventory is up-to-date
+    sleep(5)
+    registration_status = insights_client.run("--status")
+    if insights_client.config.legacy_upload:
+        assert registration_status.returncode == 0
```
ptoscano, line 20: this check is not needed: `InsightsClient.run()` by default has `check=True`, which means it will fail when the return code of `insights-client` is different than 0

integration-tests/test_status.py
```diff
@@ -0,0 +1,42 @@
...
+def test_status_unregistered(external_candlepin, insights_client):
+    """
+    Test `insights-client --status` when the client is unregistered;
+    in case of legacy_upload status command returns a different output.
+
+    """
+    # running registration to ensure system is unregistered
+    with contextlib.suppress(Exception):
+        insights_client.unregister()
+    registration_status = insights_client.run("--status")
```
ptoscano, line 35: I notice that the return code of `insights-client --status` in legacy upload mode is 1, and that means the `run()` will fail; since the exit code is expected, then I'd say to pass `check=False` as parameter of `run()` here ...

integration-tests/test_status.py
```diff
@@ -0,0 +1,42 @@
...
+    if insights_client.config.legacy_upload:
+        assert (
+            b"Insights API says this machine is NOT registered."
+            in registration_status.stdout
+        )
```
ptoscano, line 40: ... and then here check that the return code is 1:
```python
        assert registration_status.returncode == 1
```

integration-tests/test_status.py
```diff
@@ -0,0 +1,42 @@
...
+    else:
+        assert b"This host is unregistered.\n" == registration_status.stdout
```
ptoscano, line 42: ... and then here check that the return code is 0:
```python
        assert registration_status.returncode == 0
```

integration-tests/test_status.py
```diff
@@ -0,0 +1,42 @@
...
+        assert b"This host is registered.\n" == registration_status.stdout
+
+
+def test_status_unregistered(external_candlepin, insights_client):
+    """
+    Test `insights-client --status` when the client is unregistered;
+    in case of legacy_upload status command returns a different output.
+
+    """
+    # running registration to ensure system is unregistered
```
ptoscano, line 32: possible typo here?
```suggestion
    # running unregistration to ensure system is unregistered
```

ptoscano (changes requested):
Thanks for the updates! It seems to work much nicer now; there is only one thing to fix, and few really minor things to tweak.

-----

Thanks for checking
```python
assert insights_client.is_registered
```
in `test_status_registered`! If you can also add the same (as `not is_registered`, of course) in `test_status_unregistered`, we can validate that case too.

-----

Other than the above notes, I think it would be good to start editing this commit with a nice message; something along the lines:
```
tests: add new status integration tests

Add new integration tests to check that the status after the
registration and when not registered is what we expect.

Signed-off-by: Your Name <your@email>
```

ptoscano (approved):
Thank you for the updates! The newly added tests LGTM.

m-horky (approved):
LGTM

PR #171, opened 2024-02-21
build: update the spec file more

ptoscano:
Update a bit the spec file used for local builds, so it is closer to what used in RHEL:
- drop very old `redhat-access-insights`/`redhat-access-proactive` references
- use https for an URL
- fix & improve handling of `insights-client-boot.service`

Also, convert the `License` header to SPDX.


m-horky (approved):
LGTM

PR #172, opened 2024-02-26
Add pytest into CI for the master branch

m-horky:


.github/workflows/pytest.yml
```diff
@@ -0,0 +1,45 @@
...
+    runs-on: "ubuntu-latest"
+    container:
+      image: ${{ matrix.image }}
+
+    steps:
+      - name: "Checkout the repository"
+        uses: actions/checkout@v4
+
+      - name: "Install dependencies"
+        # Setuptools is required because we still rely on 'distutils' which have been dropped in Python 3.12
```
ptoscano, line 32: hm considering the client still needs distutils, wouldn't it be better to keep it in `requirements.txt` for now? this avoids having the dependencies in two different places (`requirements.txt` and here in this workflow)

in the end, once the distutils usage is dropped, also `requirements.txt` can be cleaned

.github/workflows/pytest.yml
```diff
@@ -0,0 +1,45 @@
...
+        run: |
+          dnf --setopt install_weak_deps=False install -y git-core python3-pip
+          python3 -m pip install --upgrade pip wheel
+          python3 -m pip install -r src/insights_client/tests/requirements.txt
+          python3 -m pip install git+https://github.com/RedHatInsights/insights-core.git
```
ptoscano, line 37: I think it'd be better to use the egg that is already in the sources (see `data/rpm.egg`), pointing the tests to it via the `EGG` environment variable

you can also write a `conftest.py` for the tests that would set the `EGG` environment variable if not already set; this way, you can run the test suite with a custom egg, if needed/wanted

.github/workflows/pytest.yml
```diff
@@ -0,0 +1,45 @@
...
+          python3 -m pip install setuptools
```
ptoscano, line 38: ditto (wrt manually installing distutils/setuptools)

.github/workflows/pytest.yml
```diff
@@ -0,0 +1,45 @@
+name: pytest (RHEL 8, RHEL 9)
```
ptoscano, line 1: I'd keep only "pytest", as you get the actual tested distros by looking at the matrix jobs

ptoscano (changes requested):
It looks like a good start; I see few things to improve, added some notes.

src/insights_client/tests/requirements.txt
```diff
@@ -1,5 +1 @@
-pytest~=4.6.11
-mock~=3.0.5
-requests~=2.25.1
-six~=1.15.0
```
ptoscano, line 4: `six` is still needed, used both in insights-client itself and in one test (`test_sed`)

.github/workflows/pytest.yml
ptoscano: a potential `src/insights_client/tests/conftest.py` might be:
```python
def pytest_configure(config):
    top = pathlib.Path(__file__).parent.parent.parent.parent
    if "EGG" not in os.environ:
        os.environ["EGG"] = str((top / "data" / "rpm.egg").resolve())
```

.github/workflows/pytest.yml
```diff
@@ -0,0 +1,45 @@
...
+          python3 -m pip install -r src/insights_client/tests/requirements.txt
+          python3 -m pip install git+https://github.com/RedHatInsights/insights-core.git
+          python3 -m pip install setuptools
+
+      - name: "Run pytest"
+        env:
+          PYTEST_ADDOPTS: "--color=yes --code-highlight=yes --showlocals"
+          PYTHONPATH: "src/"
+        run: |
+          python3 -m pytest src/insights_client/tests
```
ptoscano, line 45: I notice that we have one test (`test_sed`) that is created at build time with meson; because of that, running `pytest` directly from the sources will ignore it.

So, as much as extra steps this looks like, I think we need to run the tests using meson, i.e.:
```bash
$ meson setup --prefix /usr build
$ meson test -C build -v
```
this means also installing `meson` when setting up things in this workflow

ptoscano:
One more note on the changes to the `requirements.txt` file: a couple of the requirements listed there are actually needed for using the egg, meaning that, in case we use the egg that is in the sources, those requirements are actually needed. Hence, what I'd do in this case is to keep them in an own file, e.g. `core-requirements.txt`, that is still installed during the workflow run; while the result would not change much (e.g. `requests` will still get installed), at least there is a clear separation between the own requirements of the client, and those of the egg.

ptoscano:
Also: I tried to run the tests manually on Fedora 39 as non-root user using `data/rpm.egg`, and got the following failures:
```
______________________ test_exit_when_run_phases_no_sudo _______________________

    @patch('insights_client.sys.argv', ['--no-gpg'])
    def test_exit_when_run_phases_no_sudo():
        with raises(SystemExit) as pytest_wrapped_e:
            with patch('os.getuid', return_value=1):
>               insights_client._main()

../src/insights_client/tests/test_commands.py:42: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../src/insights_client/__init__.py:208: in _main
    if gpg_validate(ENV_EGG):
../src/insights_client/__init__.py:94: in gpg_validate
    home = tempfile.mkdtemp(dir=TEMPORARY_GPG_HOME_PARENT_DIRECTORY)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

suffix = '', prefix = 'tmp', dir = '/var/lib/insights/'

    def mkdtemp(suffix=None, prefix=None, dir=None):
        """User-callable function to create and return a unique temporary
        directory.  The return value is the pathname of the directory.
    
        Arguments are as for mkstemp, except that the 'text' argument is
        not accepted.
    
        The directory is readable, writable, and searchable only by the
        creating user.
    
        Caller is responsible for deleting the directory when done with it.
        """
    
        prefix, suffix, dir, output_type = _sanitize_params(prefix, suffix, dir)
    
        names = _get_candidate_names()
        if output_type is bytes:
            names = map(_os.fsencode, names)
    
        for seq in range(TMP_MAX):
            name = next(names)
            file = _os.path.join(dir, prefix + name + suffix)
            _sys.audit("tempfile.mkdtemp", file)
            try:
>               _os.mkdir(file, 0o700)
E               FileNotFoundError: [Errno 2] No such file or directory: '/var/lib/insights/tmphla7_z4d'

/usr/lib64/python3.12/tempfile.py:384: FileNotFoundError
___________________________ test_keyboard_interrupt ____________________________

os_uid = <MagicMock name='load_all' id='140501039431408'>
client = <MagicMock name='getuid' id='140501037566336'>

    @patch('os.getuid', return_value = 0)
    @patch('insights.client.InsightsConfig.load_all', side_effect = KeyboardInterrupt)
    def test_keyboard_interrupt(os_uid, client):
        with pytest.raises(SystemExit) as sys_exit:
>           insights_client._main()

../src/insights_client/tests/test_client.py:22: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../src/insights_client/__init__.py:208: in _main
    if gpg_validate(ENV_EGG):
../src/insights_client/__init__.py:94: in gpg_validate
    home = tempfile.mkdtemp(dir=TEMPORARY_GPG_HOME_PARENT_DIRECTORY)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

suffix = '', prefix = 'tmp', dir = '/var/lib/insights/'

    def mkdtemp(suffix=None, prefix=None, dir=None):
        """User-callable function to create and return a unique temporary
        directory.  The return value is the pathname of the directory.
    
        Arguments are as for mkstemp, except that the 'text' argument is
        not accepted.
    
        The directory is readable, writable, and searchable only by the
        creating user.
    
        Caller is responsible for deleting the directory when done with it.
        """
    
        prefix, suffix, dir, output_type = _sanitize_params(prefix, suffix, dir)
    
        names = _get_candidate_names()
        if output_type is bytes:
            names = map(_os.fsencode, names)
    
        for seq in range(TMP_MAX):
            name = next(names)
            file = _os.path.join(dir, prefix + name + suffix)
            _sys.audit("tempfile.mkdtemp", file)
            try:
>               _os.mkdir(file, 0o700)
E               FileNotFoundError: [Errno 2] No such file or directory: '/var/lib/insights/tmpbq8r_4fy'

/usr/lib64/python3.12/tempfile.py:384: FileNotFoundError
____________________________ test_load_config_error ____________________________

os_uid = <MagicMock name='load_all' id='140501020673616'>
insightsConfig = <MagicMock name='getuid' id='140501020677408'>

    @patch('os.getuid', return_value=0)
    @patch('insights.client.InsightsConfig.load_all', side_effect = ValueError('mocked error'))
    def test_load_config_error(os_uid, insightsConfig):
        with pytest.raises(SystemExit) as sys_exit:
>           insights_client._main()

../src/insights_client/tests/test_client.py:13: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../src/insights_client/__init__.py:208: in _main
    if gpg_validate(ENV_EGG):
../src/insights_client/__init__.py:94: in gpg_validate
    home = tempfile.mkdtemp(dir=TEMPORARY_GPG_HOME_PARENT_DIRECTORY)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

suffix = '', prefix = 'tmp', dir = '/var/lib/insights/'

    def mkdtemp(suffix=None, prefix=None, dir=None):
        """User-callable function to create and return a unique temporary
        directory.  The return value is the pathname of the directory.
    
        Arguments are as for mkstemp, except that the 'text' argument is
        not accepted.
    
        The directory is readable, writable, and searchable only by the
        creating user.
    
        Caller is responsible for deleting the directory when done with it.
        """
    
        prefix, suffix, dir, output_type = _sanitize_params(prefix, suffix, dir)
    
        names = _get_candidate_names()
        if output_type is bytes:
            names = map(_os.fsencode, names)
    
        for seq in range(TMP_MAX):
            name = next(names)
            file = _os.path.join(dir, prefix + name + suffix)
            _sys.audit("tempfile.mkdtemp", file)
            try:
>               _os.mkdir(file, 0o700)
E               FileNotFoundError: [Errno 2] No such file or directory: '/var/lib/insights/tmpife662cv'

/usr/lib64/python3.12/tempfile.py:384: FileNotFoundError
```
it looks like the `gpg_validate` short-circuiting
```python
@patch("insights_client.gpg_validate", lambda egg: True)
```
also needs to be applied to `test_load_config_error`, `test_keyboard_interrupt`, and `test_help_command`. Alternatively, we could set `BYPASS_GPG=true` as environment variable when running using `data/rpm.egg` (see my previous comment above on this, with a proposed `conftest.py`), removing the above `@patch` from `test_exit_when_run_phases_no_sudo`.

src/insights_client/tests/conftest.py
```diff
@@ -0,0 +1,29 @@
+import os
+import pathlib
+import sys
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    import _pytest.config
+
+
+def pytest_configure(config: "_pytest.config.Config"):
```
ptoscano, line 10: while I'm all for type hints, I'm wondering here if referring to the internal pytest type won't possibly break any time in the future; considering `config` is not even used at the moment, IMHO it could be fine to leave it untyped

.github/workflows/pytest.yml
```diff
@@ -0,0 +1,42 @@
...
+    runs-on: "ubuntu-latest"
+    container:
+      image: ${{ matrix.image }}
+
+    steps:
+      - name: "Checkout the repository"
+        uses: actions/checkout@v4
+
+      - name: "Install dependencies"
+        # Setuptools is required because we still rely on 'distutils' which have been dropped in Python 3.12
```
ptoscano, line 32: I guess this comment goes better in `requirements.txt` to document `setuptools` there; this way, once `distutils` is dropped, also its comment can go (and there is no change needed here)

ptoscano (approved):
LGTM, thanks!

Sadly the github actions do not run if there are no workflows already in place... so we'll see soon if there are problems, and we'll fix them as needed.

PR #173, opened 2024-02-26
Add pytest into CI for RHEL 7 branch

m-horky:


src/insights_client/tests/Makefile.am
```diff
@@ -1,25 +1,16 @@
...
 TESTS = pytest.sh
 endif
 
-pytest.sh: $(nodist_check_SCRIPTS) $(dist_check_SCRIPTS)
-	echo 'PYTHONPATH=$(top_srcdir)/src:$(top_srcdir)/data/rpm.egg $(PYTEST) $(PYTEST_FLAGS) ./*.py' > $@
+pytest.sh: $(dist_check_SCRIPTS)
+	echo 'PYTHONPATH=$(top_srcdir)/src:$(top_srcdir)/data/rpm.egg EGG=$(top_srcdir)/data/rpm.egg $(PYTEST) $(PYTEST_FLAGS) ./*.py' > $@
 	chmod +x $@
 
-CLEANFILES = pytest.sh $(nodist_check_SCRIPTS)
```
ptoscano, line 13:
```
CLEANFILES = pytest.sh
```
is still needed

.github/workflows/pytest.yml
```diff
@@ -20,7 +20,7 @@ jobs:
 
     steps:
       - name: "Checkout the repository"
-        uses: actions/checkout@v4
+        uses: actions/checkout@v3
```
ptoscano, line 23: this change needs to be moved together with the commit that adds the github workflow, and mention that v4 uses a too much newer nodejs that does not run on EL7

.github/workflows/pytest.yml
```diff
@@ -33,7 +33,7 @@ jobs:
 
       - name: "Run pytest"
         env:
-          PYTEST_ADDOPTS: "--color=yes --code-highlight=yes --showlocals"
+          PYTEST_ADDOPTS: "--color=yes --showlocals"
```
ptoscano, line 36: ditto, to be done directly in the commit that adds the github workflow

.github/workflows/pytest.yml
```diff
@@ -0,0 +1,41 @@
...
+        uses: actions/checkout@v3
+
+      - name: "Install dependencies"
+        # We can't run 'pip install' over the core repo, pip 20 doesn't understand the 'extras_require' in its setup.py
+        run: |
+          yum install -y epel-release git-core
+          yum install -y python-pip
+          python -m pip install --upgrade 'pip<21'
+          python -m pip install -r src/insights_client/tests/requirements.txt
+          git clone https://github.com/RedHatInsights/insights-core.git /tmp/egg
```
ptoscano, line 32: most likely we need to checkout the `3.0` branch of insights-core, which currently represents the code shipped in production for EL7

ptoscano (changes requested):
Thanks! Generally LGTM, with few things to fix.

> - Six is not required, since this branch only covers Python 2.7

Unfortunately `six` is still imported unconditionally:
```
src/insights_client/__init__.py:8:import six
src/insights_client/__init__.py:47:    if six.PY3:
```
and most likely the egg needs it too; similar situation for `distutils`, needed by both the client and the egg.

ptoscano (approved):
LGTM, thanks!

PR #174, opened 2024-02-27
TASK/CCT-296: Remove meson configuration for older versions of insights

grunwmar:

Done Criteria:

 - There is not unused code in meson build files (like cron configuration)
 
 ---

My jobs
 - I iterated `meson` files and found unused parts
 - I removed unnecessary conditions from these
 - I removed unused `data/cron` directory from the project
 - due to previous point, I also iterated `makefiles` and did similar changes as in `meson` files to prevent errors
 - i removed `data/cron` from `configure.ac`

*As I was mentally more or less limited (words from Diablo II "Not enough manna." quite fits) for some months I am now very unconfident and unsure, so every notes, points on this pull request, even smallest details, are more then welcome.*


data/meson.build
```diff
@@ -16,9 +16,6 @@ install_data(
...
 subdir('logrotate.d')
-if systemd.found()
-  subdir('systemd')
-  subdir('tmpfiles.d')
-else
-  subdir('cron')
-endif
+subdir('systemd')
+subdir('tmpfiles.d')
+
```
ptoscano, line 21: minor nitpic: extra empty line

ptoscano (changes requested):
Hi @grunwmar,

thanks for the changes, they mostly LGTM.

One additional change here is to make systemd properly required in meson:
https://github.com/RedHatInsights/insights-client/blob/c68e3073915bb19e12d1c7878f2ec38d58edc0f9/meson.build#L12
Most likely, the `required` parameter for `dependency()` can be dropped altogether, as the default is `true` anyway:
https://mesonbuild.com/Reference-manual_functions.html#dependency

Other than that, please rebase the branch of this PR to the `master` branch of the upstream repository, so it gets all the new build updates (such as packit, the functional tests, etc).

Thanks!

ptoscano:
Also forgot one more note; I'd say to squash the commits into one: removing the old cron data and adapting the buildsystem to that removal is most likely one single change, so let's keep it as such.

PR #176, opened 2024-02-28
TASK/CCT-296: Remove meson configuration for older versions of insights

grunwmar:
***[reopened]***

Done Criteria:
* There is not unused code in meson build files (like cron configuration)
---
My jobs:
* I iterated `meson` files and found unused parts
* I removed unnecessary conditions from these
* I removed unused `data/cron` directory from the project due to previous point, I also iterated makefiles and did similar changes as in meson files to prevent errors
* i removed `data/cron` from `configure.ac`
* i also edited `/meson.build` file and took out `required: false` attribute from `dependency()` for `systemd`

*As I was mentally more or less limited (words from Diablo II "Not enough manna." quite fits) for some months I am now very unconfident and unsure, so every notes, points on this pull request, even smallest details, are more then welcome.*

### DIFF
--- a/src/insights_client/tests/test_client.py
+++ b/src/insights_client/tests/test_client.py
@@ -24,13 +24,10 @@ def test_keyboard_interrupt(os_uid, client):
 @patch('os.getuid', return_value = 0)
 @patch('insights.client.phase.v1.get_phases')
 @patch('insights.client.InsightsClient')
-@patch('insights_client.sorted_eggs', return_value = "/var/lib/insights/newest.egg")
 @patch('insights_client.subprocess.Popen')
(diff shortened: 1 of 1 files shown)


ptoscano (changes requested):
Thanks for the changes! They generally LGTM.

Would you please amend the commit to mention that you are making systemd mandatory, and dropping the support for non-systemd configurations (such as the cron file); also, the task ID can be in its own like to avoid filling the first line of the commit message.

I suggest something on the lines of:
```
feat: drop support for non-systemd configuration

Make systemd mandatory, and drop support for non-systemd configurations
such as the cron file.

Card ID: CCT-296
```

PR #178, opened 2024-02-29
Add installing katello rpm for satellite env

zhangqianqian:


integration-tests/conftest.py
```diff
@@ -1,10 +1,27 @@
...
+import os
+import subprocess
 
 
+@pytest.fixture(scope="session", autouse=True)
+def install_katello_rpm(test_config):
+    if "satellite" in os.getenv('ENV_FOR_DYNACONF'):
+        # install katello rpm before register system against Satellite
+        satellite_hostname = test_config.get("candlepin", "host")
+        cmd = 'rpm -qa | grep katello | xargs rpm -e;rpm -Uvh http://%s/pub/katello-ca-consumer-latest.noarch.rpm' %satellite_hostname
```
ptoscano, line 11: > rpm -qa | grep katello | xargs rpm -e

This can be improved a bit:
- `rpm -q` accepts wildcards, so you can do `rpm -qa 'katello*'` and drop the `grep`
- `xargs` complains in case there are no inputs, which is the case of no katello packages already installed; please pass `--no-run-if-empty` as argument to `xargs` so it does not fail in that case

ptoscano, line 11: Please split this command in two invocations:
- one that removes the old packages
- another one that installs the new one

This way it is easier to check that each command succeeded.

integration-tests/conftest.py
```diff
@@ -1,10 +1,27 @@
...
+        ret, _ = subprocess.getstatusoutput(cmd)
```
ptoscano, line 12: `getstatusoutput()` is a very old API, I'd get away from it :)

A better replacement is `subprocess.call()`, or even better `subprocess.check_call()` since we need both the commands to succeed. Please note that, in case you run shell commands, you need to pass `shell=True`.

ptoscano (changes requested):
Thanks for the changes! I see the changes done here, and they generally go to a good direction. That said, there are some things and there to improve, and I added notes/questions about what I found so far.

integration-tests/conftest.py
```diff
@@ -1,10 +1,27 @@
...
 
+@pytest.fixture(scope="session")
+def install_katello_rpm(test_config):
+    if "satellite" in test_config.environment:
+        # install katello rpm before register system against Satellite
+        cmd = "rpm -qa 'katello*' | xargs --no-run-if-empty rpm -e"
+        ret = subprocess.check_call(cmd, shell=True)
+        
+        satellite_hostname = test_config.get("candlepin", "host")
+        cmd = "rpm -Uvh http://%s/pub/katello-ca-consumer-latest.noarch.rpm" %satellite_hostname
```
ptoscano, line 13: please split this command invocation as list of arguments, ie
```python
  cmd = ["rpm", "-Uvh", ...]
```

integration-tests/conftest.py
```diff
@@ -1,10 +1,27 @@
...
+        ret = subprocess.check_call(cmd, shell=True)
```
ptoscano, line 14: `shell=True` is not needed for this command, as it is a simple command invocation

integration-tests/conftest.py
```diff
@@ -1,10 +1,27 @@
...
 @pytest.fixture(scope="session")
 def register_subman(external_candlepin, subman_session, test_config):
-    subman_session.register(
-        username=test_config.get("candlepin", "username"),
-        password=test_config.get("candlepin", "password"),
-    )
+    activationkey=test_config.get("candlepin", "activation_keys")
+    org=test_config.get("candlepin", "org")
+    if activationkey and org:
+        subman_session.register(activationkey=activationkey, org=org)
```
ptoscano, line 21: there are also environments supported in pytest-client-tools now, so you can add `environments=test_config.get("candlepin", "environments")` as parameter for this

integration-tests/test_status.py
```diff
@@ -2,7 +2,7 @@
 import pytest
 from time import sleep
 
-pytestmark = pytest.mark.usefixtures("register_subman")
+pytestmark = pytest.mark.usefixtures("install_katello_rpm", "register_subman")
```
ptoscano, line 5: rather than this way, I'd add the `install_katello_rpm` fixture to the `register_subman` one: this way it is used automatically every time `register_subman` is used

integration-tests/conftest.py
```diff
@@ -1,10 +1,27 @@
 import pytest
+import subprocess
 
 
+@pytest.fixture(scope="session")
+def install_katello_rpm(test_config):
+    if "satellite" in test_config.environment:
+        # install katello rpm before register system against Satellite
+        cmd = "rpm -qa 'katello*' | xargs --no-run-if-empty rpm -e"
```
ptoscano, line 9: while `katello*` is good as pattern, is there a slightly more precise pattern? what about `katello-ca-consumer*`?

ptoscano (commented):
Thanks for the improvements! I added some inline notes.

------

Should `install_katello_rpm` also cleanup the katello-ca package as cleanup operation?

------

One more style note: we will soon reformat the whole codebase using the `black` tool [1], adding CI checks for it; since the files in `integration-tools` are de-facto already formatted that way, please run `black` locally on the files changed by this PR.

[1] https://github.com/psf/black

integration-tests/conftest.py
zhangqianqian: I do not think we should clean up katello-ca package if we test like the following 2 ways:
1.  Use a machine to test insights-client with satellite, and then re-install this machine with specified RHEL version/compose for another round of testing.
2. Use a vm to test insights-client with satellite, and then delete the vm. When start another round of testing, then create a fresh vm with specified RHEL version/compose.

When I start a testing, I prefer to use a totally new/clear system with specified RHEL version, so clean up katello-ca package is not necessary in my opinion.

ptoscano: While it seems not useful on throwaway systems, it makes a difference on systems that you want to reuse again for running the tests. This is why, when using the fixtures of tools of pytest-client-tools, there are cleanup steps done after each test and at the end of the test session.

I think it is important also when re-running the test on an existing system, for example to debug failures: you want to ensure the system is "clean", so in case something changes on the server then the local system is not affected by that.

zhangqianqian: This command is used in our code: cmd = "rpm -qa 'katello-ca-consumer*' | xargs --no-run-if-empty rpm -e"
when re-running insights-client+satellite test on an existing system, this command can cleanup the existing katello-ca(if exists) and re-install the latest one from specified satellite.

zhangqianqian: By the way, I have run 'black' command on the file affected by this PR.

integration-tests/conftest.py
```diff
@@ -1,10 +1,36 @@
...
+        ]
+        ret = subprocess.check_call(cmd)
+
+
+@pytest.fixture(scope="session")
+def register_subman(
+    external_candlepin, subman_session, test_config, install_katello_rpm
+):
+    environments = test_config.get("candlepin", "environments")
+    if "satellite" in environments[0]:
```
ptoscano, line 26: `candlepin.environments` specifies the environments to use for registration, not what is the current dynaconf environment (yes, unfortunately the same word has different meanings...)

you need to pass `candlepin.environments` as `environments` parameter to the `register()`, either when using AKs or in both cases

regarding the dynaconf environment: please use `test_config.environment`, which returns the dynaconf environment currently in use -- exactly as done in `install_katello_rpm` :)

integration-tests/conftest.py
```diff
@@ -1,10 +1,36 @@
...
-def register_subman(external_candlepin, subman_session, test_config):
-    subman_session.register(
-        username=test_config.get("candlepin", "username"),
-        password=test_config.get("candlepin", "password"),
-    )
+def install_katello_rpm(test_config):
+    if "satellite" in test_config.environment:
+        # install katello rpm before register system against Satellite
+        cmd = "rpm -qa 'katello-ca-consumer*' | xargs --no-run-if-empty rpm -e"
+        ret = subprocess.check_call(cmd, shell=True)
```
ptoscano, line 10: `ret` is not used, so there is no need to keep it

integration-tests/conftest.py
```diff
@@ -1,10 +1,36 @@
...
+
+        satellite_hostname = test_config.get("candlepin", "host")
+        cmd = [
+            "rpm",
+            "-Uvh",
+            "http://%s/pub/katello-ca-consumer-latest.noarch.rpm" % satellite_hostname,
+        ]
+        ret = subprocess.check_call(cmd)
```
ptoscano, line 18: ditto (unused variable)

ptoscano (changes requested):

integration-tests/conftest.py
ptoscano: > this command can cleanup the existing katello-ca(if exists) and re-install the latest one from specified satellite.

while this is correct, in general it's better for something (a test, in this case) to clean after itself when done

integration-tests/conftest.py
zhangqianqian: > you need to pass `candlepin.environments` as `environments` parameter to the `register()`, either when using AKs or in both cases
> 

Pino, you mean here: https://github.com/ptoscano/pytest-client-tools/blob/main/pytest_client_tools/subscription_manager.py#L56 ?

For now, it not necessary to pass environment parameter to subman register(). From Archana's PR for conftest, we know that environment is not needed for her testing. And if I use activationkey and org for subman registration in insights-client+satellite testing, environment parameter is also not needed for me.

We are clear that Archana use username/password for prod_cert env, and use activationkey/org for prod_cert_satellite env, which are the easiest way for both testings. I know both the envs can use both username/password mode and activation key mode for sub-man registration. But I prefer not to change this code now. If someday, Archana has the requirement to use activationkey in the prod_cert env, we can change the code then.

integration-tests/conftest.py
zhangqianqian: Can we keep this idea in mind, and not to write the cleanup code for now? If we can launch regular testings when we have a jenkins job, from which we can figure out the real requirement. If we have this test requirement from real testings, we can add the cleanup code then.

integration-tests/conftest.py
ptoscano: OK wrt not using environments for registration for now

OTOH, the two lines I mentioned use the wrong way to check for the dynaconf environment, and thus need to be fixed.

integration-tests/conftest.py
ptoscano: > write the cleanup code for now?

in the end, it'd be only about switching where the cleanup is done -- something like this (extremely simplified):
```python
@pytest.fixture()
def install_sat_packages()
    if is_sat:
        subprocess.check_call(["rpm", "-Uvh", ...])
    yield
    if is_sat:
        subprocess.check_call(["rpm", "-e", ...])
```
in the end, what would be the point of doing the cleanup before starting the test, if the use case you have in mind is running the tests on a newly provisioned jenkins worker? there is definitely nothing to cleanup there :)

zhangqianqian: Ah, I see. you suggest to split my steps into 2 parts. Done as your suggested.

ptoscano (changes requested):
Almost there! There are a couple of minor things to fix, and it looks nice to me. I'll need to test it with a Satellite server.

In the meanwhile, please
- squash all the changes to a single commit, since there is no need for multiple "fixup" steps (they make the git history harder to read/analyze)
- make sure that the resulting commit has the proper author, i.e. name/email (right now I see "root")
- make sure that the resulting commit has a proper commit message; something like:
```
tests: adapt registration for Satellite

When registering against a Satellite server, install the
katello-ca-consumer RPM package, so the right CA certificates for that
Satellite are available for subscription-manager.

Also modify the registration to use activation keys if configured,
so they can be used instead of username & password.
```
- rebase your resulting local branch on top of the current `master` branch

Let me know if you need help with any of the operations above, I'll gladly help.

zhangqianqian:
> Almost there! There are a couple of minor things to fix, and it looks nice to me. I'll need to test it with a Satellite server.
I develop code with iqe satellite613, I can share my .secret.toml to you.

ptoscano (approved):
LGTM now, thanks!

PR #179, opened 2024-02-29
feat: Stop sorting eggs

m-horky:
* Card ID: CCT-397

We know the preferred order for eggs: ENV > LAST > STABLE > RPM. It is unlikely that the RPM egg will be newer than STABLE.

### DIFF
--- a/src/insights_client/__init__.py
+++ b/src/insights_client/__init__.py
@@ -9,11 +9,8 @@ import logging
 import os
 import shutil
 import subprocess
-from subprocess import Popen, PIPE
 import sys
 import tempfile
-
-from distutils.version import LooseVersion
 
 
 INSIGHTS_DEBUG = os.environ.get("INSIGHTS_DEBUG", "").lower() == "true"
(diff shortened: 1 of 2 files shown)


src/insights_client/tests/test_client.py
```diff
@@ -24,13 +24,12 @@ def test_keyboard_interrupt(os_uid, client):
...
 @patch('insights_client.subprocess.Popen')
 @patch('insights_client.enumerate', return_value = [("1", "egg")])
 @patch('insights_client.os.path.isfile', return_value = True)
-def test_phase_error_100(isfile, enumerate, mock_subprocess, sorted_eggs, client, p, os_uid):
+def test_phase_error_100(isfile, enumerate, mock_subprocess, client, p, os_uid):
     with pytest.raises(SystemExit) as sys_exit:
         mock_subprocess.return_value.returncode= 100
         mock_subprocess.return_value.communicate.return_value = ('output', 'error')
-        insights_client.run_phase(p, client, sorted_eggs)
+        insights_client.run_phase(p, client, validated_eggs=[""])
```
ptoscano, line 34: most likely `validated_eggs` can be an empty list; `run_phase()` picks `ENV_EGG` anyway

src/insights_client/__init__.py
```diff
@@ -252,8 +214,7 @@ def _main():
     if an egg fails a phase never try it again
     """
     # sort rpm and stable eggs after verification
```
ptoscano, line 216: this comment needs to be tweaked/dropped, as it will not be true anymore

src/insights_client/tests/test_client.py
ptoscano: #180 obsoletes this change

PR #181, opened 2024-03-01
fix(gpg): Try to shutdown GPG agent before removing its home directory

m-horky:
* Card ID: CCT-405

### DIFF
--- a/src/insights_client/__init__.py
+++ b/src/insights_client/__init__.py
@@ -95,6 +95,31 @@ def sorted_eggs(eggs):
         return [eggs[1], eggs[0]]
 
 
+def _remove_gpg_home(home):
+    """Clean GPG's temporary home directory at path 'home'.
+
+    :param home: Path to the GPG's temporary home.
+    :type home: str
+    :rtype: None
+    """
+    # Shut down GPG's home agent
+    shutdown_process = subprocess.Popen(
+        ["/usr/bin/gpgconf", "--homedir", home, "--kill", "all"],
+        stdout=subprocess.PIPE, stderr=subprocess.PIPE,
+        universal_newlines=True,
+    )
+    stdout, stderr = shutdown_process.communicate()
+    if shutdown_process.returncode != 0:
+        log("Could not kill the GPG agent, got return code {rc}".format(rc=shutdown_process.returncode))
+        if stdout:
+            log(stdout)
+        if stderr:
+            log(stderr)
+
+    # Delete the temporary directory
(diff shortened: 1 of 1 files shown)


src/insights_client/__init__.py
```diff
@@ -95,6 +95,34 @@ def sorted_eggs(eggs):
...
+    """
+    # Shut down GPG's home agent
+    shutdown_process = subprocess.Popen(
+        ["/usr/bin/gpgconf", "--homedir", home, "--kill", "all"],
+        stdout=subprocess.PIPE, stderr=subprocess.PIPE,
+        universal_newlines=True,
+    )
+    stdout, stderr = shutdown_process.communicate()
+    if shutdown_process.returncode != 0:
+        log("Could not clean GPG home, got return code {rc}".format(rc=shutdown_process.returncode))
```
ptoscano, line 113: hm IMHO this message can be easily mislead for the error that happens when executing `shutil.rmtree()`; maybe I'd be more explicit:
```suggestion
        log("Could not kill the GPG agent using gpgconf, got return code {rc}".format(rc=shutdown_process.returncode))
```

src/insights_client/__init__.py
```diff
@@ -95,6 +95,34 @@ def sorted_eggs(eggs):
...
+        if stdout:
+            log(stdout)
+        if stderr:
+            log(stderr)
+
+    # Delete the temporary directory
+    try:
+        shutil.rmtree(home)
+    except Exception as exc:
+        log("Could not remove GPG home, got {exc}".format(exc=exc))
```
ptoscano, line 123: should we still do not ignore the failures of `shutil.rmtree()` for now, to check whether anything wrong still happens on cleanup?

ptoscano (commented):
The approach seems OK.

src/insights_client/__init__.py
m-horky: I am a bit scared to do that, but it might be a good idea.

ptoscano (approved):
Looking at the CI failures since they were introduced, the
```
FileNotFoundError: [Errno 2] No such file or directory: 'S.gpg-agent.extra'
```
error happens quite often, so hopefully this should help. In any case, we still have lots of time to stress this more with other PRs, and fix it again if needed.

LGTM, thanks!

ptoscano:
Ah, the failure of the `testing-farm:fedora-40-x86_64:unit/fedora` job seems an infra issue; hence merging it regardless.

PR #182, opened 2024-03-04
chore: drop the autotools buildsystem

ptoscano:
The official buildsystem is meson now, so drop the old autotools buildsystem.


m-horky (approved):
ACK

PR #183, opened 2024-03-04
feat: drop support for non-systemd configuration

grunwmar:

Done Criteria:
* There is not unused code in meson build files (like cron configuration)
---
Make systemd mandatory, and drop support for non-`systemd` configurations such as the `cron` file.

`Card ID: CCT-296`


ptoscano (approved):
LGTM now, thanks!

PR #184, opened 2024-03-05
build: update pytest executables

ptoscano:
- add "py.test", which is the old executable in pytest 3
- drop "pytest-3" and "pytest3-6", which refer to a Python 3 version (and this branch is for Python 2)


m-horky (approved):
LGTM

PR #185, opened 2024-03-05
Update handling of the MOTD file

m-horky:
1/ Improve logging; be more explicit about what is being done
2/ Remove the file instead of symlinking to /dev/null
3/ Handle situations in which the motd symlink is broken

(Updates 0ee697131b611c427986098b3c260eb5f14dc051)

### DIFF
--- a/src/insights_client/__init__.py
+++ b/src/insights_client/__init__.py
@@ -253,29 +253,43 @@ def update_motd_message():
         ))
         return
 
+    if os.path.exists(MOTD_FILE) and os.path.samefile(os.devnull, MOTD_FILE):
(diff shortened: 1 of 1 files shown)


DuckBoss:
I noticed the motd symlink code was broken resulting in errors when building insights-client from scratch. I'm assuming this includes fixes for that issue. I'll have a look next week if no one else has already reviewed this by then :+1:

m-horky:
> I'm assuming this includes fixes for that issue.
I hope I've hit the same problem as you did during build. The issue is only upstream, I didn't report any bug for it since it hasn't been built yet.

ptoscano (approved):

PR #187, opened 2024-03-05
fix: Respect PYTHONPATH environment

m-horky:
If PYTHONPATH is set to non-empty value, include it when running the egg. The Egg path is always the first to ensure the correct egg will be run, but the rest of the paths is now included.

This is useful during development. For example, `rhsm` may only be available on non-standard path (= not as installed package), and without this patch it is not possible to use it.

### DIFF
--- a/src/insights_client/__init__.py
+++ b/src/insights_client/__init__.py
@@ -193,10 +193,14 @@ def run_phase(phase, client, validated_eggs):
             continue
         client_debug("phase '%s'; egg '%s'" % (phase['name'], egg))
 
-        # set up the env
+        # prepare the environment
+        pythonpath = str(egg)
+        env_pythonpath = os.environ.get("PYTHONPATH", "")  # type: str
+        if env_pythonpath:
+            pythonpath += ":" + env_pythonpath
         insights_env = {
             "INSIGHTS_PHASE": str(phase['name']),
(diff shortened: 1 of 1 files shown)


src/insights_client/__init__.py
```diff
@@ -193,11 +193,10 @@ def run_phase(phase, client, validated_eggs):
...
         client_debug("Attempting phase '%s' with egg '%s'" % (phase['name'], egg))
 
-        # setup the env
-        insights_env = {
-            "INSIGHTS_PHASE": str(phase['name']),
-            "PYTHONPATH": str(egg)
-        }
+        # prepare the environment
+        pythonpath = os.environ.get("PYTHONPATH", "")  # type: str
+        pythonpath = str(egg) if not pythonpath else "{}:{}".format(egg, pythonpath)
```
ptoscano, line 198: i'd suggest something that i find slightly more readable:
```python
env_pythonpath = str(egg)
pythonpath = os.environ.get("PYTHONPATH", None)
if pythonpath:
    env_pythonpath += ":" + pythonpath
```
ideally it's a perfect use case for the walrus operator, sadly it's too new for this code (added in 3.8, baseline for the code is 3.6)

ptoscano (approved):

PR #188, opened 2024-03-06
Test : Added tests for insights-client --unregister option

Archana-PandeyM:


integration-tests/test_unregister.py
```diff
@@ -0,0 +1,43 @@
...
+
+def test_unregister(insights_client):
+    """Test insights-client --unregister after registering it
+
+    If the unregistration is successful, `unregister` method returns True
+    """
+    insights_client.register()
+    assert insights_client.is_registered
+
+    registration_status = insights_client.run("--unregister")
```
ptoscano, line 14:
```suggestion
    registration_status = insights_client.unregister()
```

integration-tests/test_unregister.py
```diff
@@ -0,0 +1,43 @@
...
+    assert (
+        "Successfully unregistered from the Red Hat Insights Service"
+        in registration_status.stdout
+    )
```
ptoscano, line 18: after this, you can also assert that is `not is_registered`

integration-tests/test_unregister.py
```diff
@@ -0,0 +1,43 @@
...
+    """Test insights-client --unregister on unregistered system
+
+    If the unregistration is successful, `--unregister` returns True on first attempt and
+    false on subsequent attempts.
+    """
+    insights_client.register()
+    assert insights_client.is_registered
+
+    # unregister once
+    registration_status = insights_client.run("--unregister")
```
ptoscano, line 31: ditto (`unregister()`)

integration-tests/test_unregister.py
```diff
@@ -0,0 +1,43 @@
...
+    assert not insights_client.is_registered
+    assert (
+        "Successfully unregistered from the Red Hat Insights Service"
+        in registration_status.stdout
+    )
+    # unregister twice
+    registration_status = insights_client.run("--unregister", check=False)
```
ptoscano, line 38: ditto (`unregister()`)

ptoscano (changes requested):
Generally LGTM; added a couple of notes.

----

Do we really need two different tests? After all, `test_unregister_twice` is exactly `test_unregister` plus an extra check; IMHO it would seem easier to do only one test (`test_unregister`) that check for unregistration, also if done twice.

integration-tests/test_unregister.py
ptoscano: ah scratch this, you cannot pass any additional parameters to `unregister()`

integration-tests/test_unregister.py
Archana-PandeyM: In my opinion I do not see a problem with this, also it looks symmetrical throughout test(aesthetically )

integration-tests/test_unregister.py
Archana-PandeyM: ok I wil add this.

Archana-PandeyM:
> Generally LGTM; added a couple of notes.
> 
> Do we really need two different tests? After all, `test_unregister_twice` is exactly `test_unregister` plus an extra check; IMHO it would seem easier to do only one test (`test_unregister`) that check for unregistration, also if done twice.

I think these two are different tests and hence two tests.

jstavel:
looks good to me. /lgtm

ptoscano:
> I think these two are different tests and hence two tests.

While that is true for a "scenario" POV, in practice running `test_unregister_twice` alone is enough to check that everything related to the unregistration works, so `test_unregister` is redundant.

Archana-PandeyM:
> > I think these two are different tests and hence two tests.
> 
> While that is true for a "scenario" POV, in practice running `test_unregister_twice` alone is enough to check that everything related to the unregistration works, so `test_unregister` is redundant.

In my opinion two different tests are needed because the intent of both tests are different.

zhangqianqian:
I think the 2 cases are different.
test_unregister is for the basic function
test_unregister_twice is negative testing: check what happens when run "insights-client --unregister "on a unregistered system. So I agreee with Archana.

integration-tests/test_unregister.py
```diff
@@ -0,0 +1,45 @@
...
+
+def test_unregister(insights_client):
+    """Test insights-client --unregister after registering it
+
+    If the unregistration is successful, `unregister` method returns True
+    """
+    insights_client.register()
+    assert insights_client.is_registered
+
+    registration_status = insights_client.run("--unregister")
```
ptoscano, line 14: this variable is not used -- simply do not assign the result of `run()`

integration-tests/test_unregister.py
```diff
@@ -0,0 +1,45 @@
...
+    """Test insights-client --unregister on unregistered system
+
+    If the unregistration is successful, `--unregister` returns True on first attempt and
+    false on subsequent attempts.
+    """
+    insights_client.register()
+    assert insights_client.is_registered
+
+    # unregister once
+    registration_status = insights_client.run("--unregister")
```
ptoscano, line 32: ditto (unused variable)

integration-tests/test_unregister.py
```diff
@@ -0,0 +1,45 @@
...
+    assert not insights_client.is_registered
+    assert (
+        "Successfully unregistered from the Red Hat Insights Service"
+        in registration_status.stdout
+    )
+    # unregister twice
+    registration_status = insights_client.run("--unregister", check=False)
```
ptoscano, line 39: maybe `s/registration_status/unregistration_status/` for clarity?

ptoscano (changes requested):
OK, I'm still not convinced about the duplication of tests, however I'm fine in keeping it for now. Please then amend the commit message to explain why both are needed, even though one test does the job also of another test: this way, it will be less confusing when browsing through the git history in the future.

I added a couple of minor notes that I missed in the previous review.

After these changes, please rebase against the `master` branch.

Thanks!

integration-tests/test_unregister.py
Archana-PandeyM: I did not understand ? It is used in assert statement.

ptoscano: <facepalm>

ignore this, sorry

integration-tests/test_unregister.py
ptoscano: *facepalm*

ignore this, sorry

integration-tests/test_unregister.py
Archana-PandeyM: registration status is status of system, I mean status of system. this can be true/false.

ptoscano: well, in this case it is the status of the unregistration; what you describe would be the output of `--status`

ptoscano (approved):
LGTM, thanks!

waiting for the successful runs of the tests before merging

PR #190, opened 2024-03-08
Test: add test cases for insights-client with satellite

zhangqianqian:
https://gitlab.cee.redhat.com/insights-qe/iqe-insights-client-plugin/-/blob/master/iqe_insights_client/tests/test_satellite.py

### DIFF
--- a/integration-tests/test_client_options.py
+++ b/integration-tests/test_client_options.py
@@ -1,0 +1,20 @@
+import pytest
+from time import sleep
+
+pytestmark = pytest.mark.usefixtures("register_subman")
+
+
+def test_set_ansible_host_info(insights_client):
+    """
+    Test if the ansible-host can be set with satellite
+    Related to https://issues.redhat.com/browse/RHEL-3826
+    """
+    # Register system against Satellite, and register insights through satellite
+    insights_client.register()
+    sleep(10)
+    assert insights_client.is_registered
+
+    # Update ansible-host
+    ret = insights_client.run("--ansible-host=foo.example.com", check=False)
+    assert "Could not update Ansible hostname" not in ret.stdout
+    assert ret.returncode == 0
--- a/integration-tests/test_obfuscation.py
+++ b/integration-tests/test_obfuscation.py
@@ -1,0 +1,72 @@
+import pytest
+import subprocess
+import os
+import yaml
+from time import sleep
+import re
+
+pytestmark = pytest.mark.usefixtures("register_subman")
+CONTENT_REDACTION_FILE = "/etc/insights-client/file-content-redaction.yaml"
+
+
+def test_obfuscation(insights_client, tmp_path):
+    """
+    Test if the satellite hostname is obfuscated
+    Related to https://issues.redhat.com/browse/RHEL-2471
+
+    """
+    # Register system to insights
+    insights_client.register()
+    sleep(5)
+    assert insights_client.is_registered
+
+    # Record the current hostname
+    with open("/etc/hostname", "r") as f:
+        previous_hostname = f.read().strip()
+
+    try:
+        # Set a new hostname
+        hostname_to_test = "insights-client-satellite"
+        set_hostname(hostname_to_test)
+        cmd = ["hostnamectl", "status"]
+        output = subprocess.check_output(cmd, universal_newlines=True)
+        assert hostname_to_test in output
+
+        # redaction on hostname:
+        data_redaction(hostname_to_test)
+
+        # check the collection, if redaction pattern is filtered out
+        # collect data
+        archive_result = insights_client.run("--no-upload")
+
+        # extract the collected tar file
+        m = re.search("^Archive saved at (.+)$", archive_result.stdout, re.MULTILINE)
+        assert m
+        archive_location = m.group(1)
+        cmd = ["tar", "-vxzf", archive_location, "-C", tmp_path]
+        subprocess.check_call(cmd)
+
+        # the redaction data should not locate in the collected files
+        archive_extract = os.path.basename(archive_location).split(".")[0]
+        path_to_check = tmp_path / archive_extract
+        cmd = ["grep", "-rnw", ".", "-e", hostname_to_test]
+        # should not find the obfuscate pattern, the return code should not zero.
+        ret = subprocess.run(cmd, cwd=path_to_check)
+        assert ret.returncode == 1
+    finally:
+        # Recover the previous hostname for the system
+        set_hostname(previous_hostname)
+        # Remove all data redaction
+        os.remove(CONTENT_REDACTION_FILE)
+
+
+def set_hostname(hostname_to_set):
+    cmd = ["hostnamectl", "set-hostname", hostname_to_set]
+    subprocess.check_call(cmd)
+
+
+def data_redaction(pattern_string):
+    file_content = {"patterns": {"regex": [pattern_string]}}
+    with open(CONTENT_REDACTION_FILE, "w") as f:
+        yaml.dump(file_content, f, default_flow_style=False)
+    os.chmod(CONTENT_REDACTION_FILE, 0o600)
--- a/integration-tests/test_registration.py
+++ b/integration-tests/test_registration.py
@@ -1,5 +1,5 @@
 import os
-
+import contextlib
 import pytest
 
 
@@ -65,3 +65,31 @@ def test_double_registration(insights_client):
         machine_id_new = f.read()
 
     assert machine_id_new == machine_id_old
+
+
+@pytest.mark.parametrize(
+    "legacy_upload_value",
+    [
+        pytest.param(True, marks=pytest.mark.xfail),
+        pytest.param(False),
+    ],
+)
+def test_register_group_option(insights_client, legacy_upload_value):
+    """
+    Bug https://issues.redhat.com/browse/RHINENG-7567 exists on both
+    production env and satellite env with legacy_upload=True.
+    With legacy_upload=False, "insights-client --register --group=tag"
+    works well on both envs.
+    """
+    # make sure the system is not registered to insights
+    with contextlib.suppress(Exception):
+        insights_client.unregister()
+    assert not insights_client.is_registered
+    insights_client.config.legacy_upload = legacy_upload_value
+    insights_client.config.save()
+    register_group_option = insights_client.run(
+        "--register",
+        "--group=tag",
+        check=False,
+    )
+    assert register_group_option.returncode == 0
--- a/integration-tests/requirements.txt
+++ b/integration-tests/requirements.txt
@@ -1,1 +1,2 @@
 git+https://github.com/ptoscano/pytest-client-tools@main
+pyyaml


integration-tests/test_satellite.py
```diff
@@ -0,0 +1,113 @@
...
+    insights_client.register()
+    sleep(10)
+    assert insights_client.is_registered
+
+    # Update ansible-host
+    set_ansible_host = insights_client.run("--ansible-host='foo.example.com'")
+    assert set_ansible_host.returncode == 0
+
+
+def test_register_group_option(insights_client):
```
ptoscano, line 81: this test seems related to the registration, and thus IMHO ought to better be in `test_register.py`; IMHO the fact that it fails on Satellite does not make it "specific" to Satellite, and running on Satellite seems simply a different scenario/setup for the test

integration-tests/test_satellite.py
```diff
@@ -0,0 +1,113 @@
...
+
+def data_redaction(pattern_string):
+    insights_folder = "/etc/insights-client/"
+    if not os.path.exists(insights_folder):
+        os.makedirs(insights_folder)
+    file_content = {"patterns": {"regex": [pattern_string]}}
+    data_redaction_file = "/etc/insights-client/file-content-redaction.yaml"
+    with open(data_redaction_file, "w") as f:
+        yaml.dump(file_content, f, default_flow_style=False)
+    os.chmod(data_redaction_file, int("600", 8))
```
ptoscano, line 113: while `int("600", 8)` works, IMHO it is not that readable :-) simply use the Python prefixes for non-decimal numbers:
```suggestion
    os.chmod(data_redaction_file, 0o600)
```

ptoscano (changes requested):
thanks!

integration-tests/test_satellite.py
```diff
@@ -0,0 +1,113 @@
...
+    with contextlib.suppress(Exception):
+        insights_client.unregister()
+    assert not insights_client.is_registered
+    """ Register the system to insights with group option
+    Use insights_client.run can not find the existing bug, so try subprocess
+    register_with_group = insights_client.run("--register", "--group='tag'")
+    assert register_with_group.returncode == 0
+    """
+    cmd = ["insights-client", "--register", "--group='tag'"]
+    subprocess.check_call(cmd, shell=True)
```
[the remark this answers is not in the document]
zhangqianqian: I forgot the details with the invocation after a weekend... Let me try when I'm free.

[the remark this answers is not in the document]
zhangqianqian: ------------------------------------------------------------- live log call -------------------------------------------------------------
DEBUG    pytest_client_tools:util.py:87 running ['insights-client', '--unregister'] with options {'check': True, 'stdout': -1, 'stderr': -1, 'text': True}
DEBUG    pytest_client_tools:util.py:95 result: CompletedProcess(args=['insights-client', '--unregister'], returncode=1, stdout='This host is not registered, unregistration is not applicable.\n', stderr='')
DEBUG    pytest_client_tools:util.py:87 running ['insights-client', '--status'] with options {'check': False, 'stdout': -1, 'stderr': -1, 'text': True}
DEBUG    pytest_client_tools:util.py:95 result: CompletedProcess(args=['insights-client', '--status'], returncode=0, stdout='This host is unregistered.\n', stderr='')
DEBUG    pytest_client_tools:util.py:87 running ['insights-client', '--register', "--group='tag'"] with options {'check': True, 'stdout': -1, 'stderr': -1, 'text': True}
DEBUG    pytest_client_tools:util.py:95 result: CompletedProcess(args=['insights-client', '--register', "--group='tag'"], returncode=0, stdout='Automatic scheduling for Insights has been enabled.\nStarting to collect Insights data for dell-per740-68-vm-03.lab.eng.pek2.redhat.com\nWriting RHSM facts to /etc/rhsm/facts/insights-client.facts ...\nUploading Insights data.\nSuccessfully uploaded report for dell-per740-68-vm-03.lab.eng.pek2.redhat.com.\nView the Red Hat Insights console at https://console.redhat.com/insights/\n', stderr='')
PASSED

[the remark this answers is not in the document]
zhangqianqian: When I run "insights_client.run("--register", "--group='tag'")", I got PASS result.

integration-tests/test_satellite.py
zhangqianqian: Jan is focusing the test_register.py migration(https://issues.redhat.com/browse/CCT-340). We can move this case when his PR is merged. What's your thoughts? @Archana-PandeyM @jstavel

integration-tests/test_satellite.py
```diff
@@ -0,0 +1,113 @@
...
+import yaml
+import traceback
+import contextlib
+from time import sleep
+
+
+pytestmark = pytest.mark.usefixtures("register_subman")
+
+
+def test_obfuscation(insights_client):
```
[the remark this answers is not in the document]
zhangqianqian: test_redaction.py is not migrated yet, no one takes it now. We can debut this case, and then move it to test_redaction.py with a proper chance. @Archana-PandeyM @jstavel , FYI.

integration-tests/test_satellite.py
```diff
@@ -0,0 +1,108 @@
...
+        # check the collection, if redaction pattern is filtered out
+        # collect data
+        archive_result = insights_client.run("--no-upload")
+
+        # extract the collected tar file
+        m = re.search("^Archive saved at (.+)$", archive_result.stdout, re.MULTILINE)
+        assert m
+        archive_location = m.group(1)
+        cmd = "tar -xzvf %s -C %s" % (archive_location, tmp_path)
+        subprocess.check_call(cmd, shell=True)
```
ptoscano, line 48: i'd split this command invocation into a list, passing that to `subprocess.check_call()` with no need for `shell=True`; using the shell as less as possible
1. makes command executions faster
2. avoids the need to deal with shell quoting, which is always a potential recipe for troubles

integration-tests/test_satellite.py
```diff
@@ -0,0 +1,108 @@
...
+
+        # the redaction data should not locate in the collected files
+        archive_extract = archive_location.split(".")[0].split("/")[-1]
+        path_to_check = tmp_path / archive_extract
+        cmd = "grep -rnw . -e %s" % hostname_to_test
+        # should not find the obfuscate pattern, the return code should not zero.
+        ret = subprocess.run(cmd, shell=True, cwd=path_to_check)
```
ptoscano, line 55: as above: split the path into a list, and run it without a shell

integration-tests/test_satellite.py
```diff
@@ -0,0 +1,108 @@
...
+        assert ret.returncode == 1
+    finally:
+        # Recover the previous hostname for the system
+        set_hostname(previous_hostname)
+        # Remove all data redaction
+        os.remove("/etc/insights-client/file-content-redaction.yaml")
```
ptoscano, line 61: since this path is used in a couple of places, it makes sense to factorize it into a module level variable: for example, add
```python
CONTENT_REDACTION_FILE = "/etc/insights-client/file-content-redaction.yaml"
```
before the tests (right after the `import`s), and use it in the two places

integration-tests/test_satellite.py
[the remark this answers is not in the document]
ptoscano: > When I run "insights_client.run("--register", "--group='tag'")", I got PASS result.

that sounds like it should be good to use... or am I missing anything?

integration-tests/test_satellite.py
ptoscano: other test cases can be added to an existing file, not an issue IMHO :)

integration-tests/test_satellite.py
```diff
@@ -0,0 +1,108 @@
+import pytest
+import subprocess
+import os
+import yaml
+import traceback
```
ptoscano, line 5: this import is no more needed

integration-tests/test_satellite.py
```diff
@@ -0,0 +1,108 @@
...
+
+def test_obfuscation(insights_client, tmp_path):
+    """
+    Test if the satellite hostname is obfuscated
+    Related to https://bugzilla.redhat.com/show_bug.cgi?id=1963917
+
+    """
+    # Register system to insights
+    insights_client.register()
+    # assert insights_client.is_registered
```
ptoscano, line 22: hm isn't this needed?

ptoscano (changes requested):
Thanks for the improvements! I'll give it a try soon.

------

One more missing thing: since the `yaml` module is used by the new tests, you need to ensure it is installed when setting up the virtualenv for the tests; to do so, add `pyyaml` to `integration-tests/requirements.txt`

integration-tests/test_satellite.py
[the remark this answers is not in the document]
zhangqianqian: Bug https://issues.redhat.com/browse/RHINENG-7567 is not fixed yet, and I can reproduce this bug manually, and I can reproduce this bug by subprocess way to run insights-client. I don't know why "insights_client.run("--register", "--group='tag'")" does not reproduce this bug.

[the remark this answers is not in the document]
ptoscano: I have a potential idea: what if you try:
```python
insights_client.run("--register", "--group=tag")
```
i.e. without single quotes around the value of `--group`.

I'm thinking that, looking at the code:
```python
    cmd = ["insights-client", "--register", "--group='tag'"]
    subprocess.check_call(cmd, shell=True)
```
since this uses a shell, in the end it is like invoking `insights-client --register --group='tag'`, and the quoting of `tag` is eaten by the shell, so `--group` only gets `tag` without quotes. When using `run()` the shell is not used, so `--group` gets a literal `'tag'` (with quotes) as value.

LMK (Let Me Know) :)

integration-tests/test_satellite.py
zhangqianqian: Running "assert insights_client.is_registered" always failed without any reason, which blocks me debugging the following code.

integration-tests/test_satellite.py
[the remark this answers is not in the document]
zhangqianqian: The following is what if I use insights_client.run("--register", "--group=tag"):


------------------------------------------------------------- live log call -------------------------------------------------------------
DEBUG    pytest_client_tools:util.py:87 running ['insights-client', '--unregister'] with options {'check': True, 'stdout': -1, 'stderr': -1, 'text': True}
DEBUG    pytest_client_tools:util.py:95 result: CompletedProcess(args=['insights-client', '--unregister'], returncode=1, stdout='This host is not registered, unregistration is not applicable.\n', stderr='')
DEBUG    pytest_client_tools:util.py:87 running ['insights-client', '--status'] with options {'check': False, 'stdout': -1, 'stderr': -1, 'text': True}
DEBUG    pytest_client_tools:util.py:95 result: CompletedProcess(args=['insights-client', '--status'], returncode=0, stdout='This host is unregistered.\n', stderr='')
DEBUG    pytest_client_tools:util.py:87 running ['insights-client', '--register', '--group=tag'] with options {'check': True, 'stdout': -1, 'stderr': -1, 'text': True}
DEBUG    pytest_client_tools:util.py:95 result: CompletedProcess(args=['insights-client', '--register', '--group=tag'], returncode=0, stdout='Automatic scheduling for Insights has been enabled.\nStarting to collect Insights data for dell-per740-68-vm-03.lab.eng.pek2.redhat.com\nWriting RHSM facts to /etc/rhsm/facts/insights-client.facts ...\nUploading Insights data.\nSuccessfully uploaded report for dell-per740-68-vm-03.lab.eng.pek2.redhat.com.\nView the Red Hat Insights console at https://console.redhat.com/insights/\n', stderr='')
PASSED

integration-tests/test_satellite.py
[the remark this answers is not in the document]
Archana-PandeyM: > test_redaction.py is not migrated yet, no one takes it now. We can debut this case, and then move it to test_redaction.py with a proper chance. @Archana-PandeyM @jstavel , FYI.

makes sense to move obfuscation and redaction tests to their respective modules.

integration-tests/test_satellite.py
```diff
@@ -0,0 +1,108 @@
...
+    Related to https://issues.redhat.com/browse/RHELPLAN-160348,
+    and https://bugzilla.redhat.com/show_bug.cgi?id=222729
+    """
+    # Register system against Satellite, and register insights through satellite
+    insights_client.register()
+    sleep(10)
+    assert insights_client.is_registered
+
+    # Update ansible-host
+    insights_client.run("--ansible-host='foo.example.com'")
```
Archana-PandeyM, line 77: Please add check=False in run method call ,e.g. insights_client.run("--ansible-host='foo.example.com'", check=False)

integration-tests/test_satellite.py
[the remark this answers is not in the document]
zhangqianqian: I tried to use insights_client.run("--register", "--group=tag") and set insights.legacy_upload = true in .secret.tomal. Yes, this case fails.

But bug https://issues.redhat.com/browse/RHINENG-7567 is not fixed when insights-client(legacy_upload=true) integrates with satellite. I prefer to add a check for the satellite env.

integration-tests/test_satellite.py
zhangqianqian: When we have a solution for insights_client.run("--register", "--group=tag"), I will move this case.

integration-tests/test_satellite.py
zhangqianqian: it works with sleep(5)

integration-tests/test_satellite.py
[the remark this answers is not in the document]
zhangqianqian: I tried manually, and found:
1. Bug https://issues.redhat.com/browse/RHINENG-7567 exists on both  production env and satellite env with legacy_upload=True.
2. When configuring legacy_upload=False, "insights-client --register --group=tag" works well on both envs.

integration-tests/conftest.py
```diff
@@ -16,6 +17,7 @@ def install_katello_rpm(test_config):
     yield
     if "satellite" in test_config.environment:
         cmd = "rpm -qa 'katello-ca-consumer*' | xargs rpm -e"
+        logging.debug("ERASING katello ca rpm...........")
```
ptoscano, line 20: after #198 was merged, this should not be needed anymore

integration-tests/conftest.py
```diff
@@ -1,5 +1,6 @@
 import pytest
 import subprocess
+import logging
```
ptoscano, line 3: ... including this

integration-tests/test_satellite.py
```diff
@@ -0,0 +1,117 @@
...
+
+def test_set_ansible_host_info(insights_client):
+    """
+    Test if the ansible-host can be set with satellite
+    Related to https://issues.redhat.com/browse/RHELPLAN-160348,
+    and https://bugzilla.redhat.com/show_bug.cgi?id=222729
+    """
+    # Register system against Satellite, and register insights through satellite
+    insights_client.register()
+    sleep(10)
```
ptoscano, line 73: if 5 seconds are enough in `test_obfuscation`, then they should be enough here too

integration-tests/test_satellite.py
```diff
@@ -0,0 +1,117 @@
...
+    assert ret.returncode == 0
+
+
+@pytest.mark.parametrize("legacy_upload_value", [True, False])
+def test_register_group_option(insights_client, legacy_upload_value):
+    """
+    Bug https://issues.redhat.com/browse/RHINENG-7567 exists on both
+    production env and satellite env with legacy_upload=True.
+
+    With egacy_upload=False, "insights-client --register --group=tag"
```
ptoscano, line 88:
```suggestion
    With legacy_upload=False, "insights-client --register --group=tag"
```

integration-tests/test_satellite.py
```diff
@@ -0,0 +1,117 @@
...
+    so I check  legacy_upload=True to cover bug.
+    """
+    # make sure the system is not registered to insights
+    with contextlib.suppress(Exception):
+        insights_client.unregister()
+    assert not insights_client.is_registered
+    if legacy_upload_value is True:
+        cmd = ["insights-client", "--register", "--group=tag"]
+        subprocess.check_call(cmd, shell=True)
+        insights_client.unregister()
```
ptoscano, line 102: not need to explicitly unregister, the `insights_client` fixture does it already at the end the the test

integration-tests/test_satellite.py
```diff
@@ -0,0 +1,117 @@
...
+    else:
+        insights_client.run("--register", "--group=tag")
+        insights_client.unregister()
```
ptoscano, line 105: ditto

ptoscano, line 105: the parametrized value of `legacy_upload` is not actually configured for `insights-client`, which means it will run using its default (False); you need to manually set it:
```python
    insights_client.config.legacy_upload = legacy_upload_value
```

also, since the failing combination is `legacy_upload` + Satellite, then I'd do the following:
- unconditionally use the `run()` way of registering, also adding `check=False` and storing the command return
- in case the current environment is a Satellite one + the `legacy_upload` value is `False` + the test fails, use `pytest.xfail()` to properly mark the known failure situation as such

ptoscano (changes requested):
in addition to the mentioned notes, I'd start moving the tests to (IMHO) better locations than a catch-all `test_satellite.py`:
- `test_obfuscation` -> a new `test_obfuscation.py`
- `test_set_ansible_host_info` -> a new `test_ansible_host.py`, or some other file more specific to these properties
- `test_register_group_option` -> the existing `test_registration.py`

thanks!

integration-tests/test_satellite.py
ptoscano: actually, if it fails in any case when using `legacy_upload`, then I'd mark the `False` value of the `legacy_upload_value` parameter as xfail

integration-tests/test_satellite.py
zhangqianqian: 5 seconds does not work in this test case:

integration-tests/test_satellite.py
zhangqianqian: > the parametrized value of `legacy_upload` is not actually configured for `insights-client`, which means it will run using its default (False); you need to manually set it:

Yes, it is. At the same time, bug RHINENG-7567 happens when legacy_upload=True. 
So we can use pytest.xfail when legacy_upload=True. And run this test case with legacy_upload=False(default set in .secrets.toml) without specifying its value manually.

integration-tests/test_registration.py
```diff
@@ -75,13 +76,21 @@ def test_register_group_option(insights_client, legacy_upload_value):
     With legacy_upload=False, "insights-client --register --group=tag"
     works well on both envs.
     """
-    if legacy_upload_value:
-        pytest.xfail(reason="RHINENG-7567 exists with legacy_upload=True.")
     # make sure the system is not registered to insights
     with contextlib.suppress(Exception):
         insights_client.unregister()
     assert not insights_client.is_registered
+    set_legacy_upload_conf(legacy_upload_value)
```
ptoscano, line 83: no need for this - pytest-client-tools can do this already:
```python
insights_client.conf.legacy_upload = legacy_upload_value
insights_client.conf.save()
```

integration-tests/test_obfuscation.py
```diff
@@ -0,0 +1,72 @@
...
+import os
+import yaml
+from time import sleep
+import re
+
+pytestmark = pytest.mark.usefixtures("register_subman")
+CONTENT_REDACTION_FILE = "/etc/insights-client/file-content-redaction.yaml"
+
+
+def test_obfuscation(insights_client, tmp_path):
```
ptoscano, line 12: this test always fails, and the open bugs for it shows this too; hence, please use the `pytest.mark.xfail()` marker to mark this test as "expected to fail"

integration-tests/test_registration.py
```diff
@@ -65,3 +65,25 @@ def test_double_registration(insights_client):
         machine_id_new = f.read()
 
     assert machine_id_new == machine_id_old
+
+
+@pytest.mark.parametrize("legacy_upload_value", [True, False])
+def test_register_group_option(insights_client, legacy_upload_value):
```
ptoscano, line 71: this test fails on Satellite with legacy upload, and it's a known issue (you linked the issue too); hence, I'd mark this combination as known failure -- something like:
```python
@pytest.mark.parametrize("legacy_upload_value", [
    pytest.param(True, marks=pytest.mark.xfail),  # fails on Satellite
    pytest.param(False),
])
```
this will mark it as xfail also on non-Sat, even if it works there; ideally there should be a better logic to add the xfail mark to the `True` parameter only when needed; I'd say to defer this to a later time, the general xfail is IMHO good enough for now

integration-tests/test_client_options.py
```diff
@@ -0,0 +1,21 @@
...
+from time import sleep
+
+pytestmark = pytest.mark.usefixtures("register_subman")
+
+
+def test_set_ansible_host_info(insights_client):
+    """
+    Test if the ansible-host can be set with satellite
+    Related to https://issues.redhat.com/browse/RHELPLAN-160348,
+    and https://bugzilla.redhat.com/show_bug.cgi?id=222729
```
ptoscano, line 11: hmm is this BZ reference correct? that's a RHEL 5 bug for libtiff...

integration-tests/test_client_options.py
```diff
@@ -0,0 +1,21 @@
+import pytest
+from time import sleep
+
+pytestmark = pytest.mark.usefixtures("register_subman")
+
+
+def test_set_ansible_host_info(insights_client):
+    """
+    Test if the ansible-host can be set with satellite
+    Related to https://issues.redhat.com/browse/RHELPLAN-160348,
```
ptoscano, line 10: I'd use the "canonical" reference here, i.e. https://issues.redhat.com/browse/RHEL-3826

integration-tests/test_obfuscation.py
```diff
@@ -0,0 +1,72 @@
...
+    """
+    Test if the satellite hostname is obfuscated
+    Related to https://bugzilla.redhat.com/show_bug.cgi?id=1963917
```
ptoscano, line 15: migrated to jira, so https://issues.redhat.com/browse/RHEL-2471 instead

ptoscano (changes requested):
generally good improvements here, thanks! there are few notes for expected failures that need to be added, and small nits I found while reviewing/testing

since it is getting closer to being merged, then please:
- rebase this on top of the current `master` branch
- squash it in a single commit, or a commit per test (either option is fine)
- improve the messages of the commit(s) by describing what is being added/changed, why, etc

thanks!

integration-tests/test_obfuscation.py
zhangqianqian: Bug https://issues.redhat.com/browse/RHINENG-8846 is going to merged, so l think we can leave this code unchanged. Because bug will be already fixed when we use this test case in regular testing.

integration-tests/test_registration.py
zhangqianqian: Not only on satellite, but also on production the bug exists. In another word, it is a common bug, not specific to satellite env.

ptoscano: oh ok, then simply drop the comment about Satellite from my above code snippet :)

zhangqianqian:
> generally good improvements here, thanks! there are few notes for expected failures that need to be added, and small nits I found while reviewing/testing
> 
> since it is getting closer to being merged, then please:
> 
> * rebase this on top of the current `master` branch
> * squash it in a single commit, or a commit per test (either option is fine)
> * improve the messages of the commit(s) by describing what is being added/changed, why, etc
> 
> thanks!

Done. Thank you.

integration-tests/test_obfuscation.py
ptoscano: That fix will not be available for 6/7 days yet, and I would like to merge this PR (which does not have left concerns from me) rather than waiting for some days... :) In the end, removing the xfail marker can be done quickly/easily later on.

WDYT?

zhangqianqian: Sure, pls merge this PR.

ptoscano (approved):

ptoscano:
There is something wrong when provisioning Fedora 40 systems for testing; OTOH this update was manually tested with a real Insights server, and all the other tests pass, so merging anyway.

PR #191, opened 2024-03-08
feat(test): Add tests for registration and unregistration

m-horky:
* Card ID: CCT-161

### DIFF
--- a/integration-tests/test_registration.py
+++ b/integration-tests/test_registration.py
@@ -1,0 +1,68 @@
+import os
+import subprocess
+
+import pytest
+
+
+pytestmark = pytest.mark.usefixtures("register_subman")
+
+MACHINE_ID_FILE: str = "/etc/insights-client/machine-id"
+
+
+def test_machineid_exists_only_when_registered(insights_client):
+    """`machine-id` is only present when insights-client is registered."""
+    assert not insights_client.is_registered
+    assert not os.path.exists(MACHINE_ID_FILE)
+
+    res = insights_client.run(check=False)
+    assert (
+        "This host has not been registered. Use --register to register this host."
+    ) in res.stdout
+    assert res.returncode != 0
+    assert not os.path.exists(MACHINE_ID_FILE)
+
+    insights_client.register()
+    assert os.path.exists(MACHINE_ID_FILE)
+
+    insights_client.unregister()
+    assert not os.path.exists(MACHINE_ID_FILE)
+
+
+def test_machineid_changes_on_new_registration(insights_client):
+    """machine-id content changes when insights-client is un- & registered."""
+    insights_client.register()
+    with open(MACHINE_ID_FILE, "r") as f:
+        machine_id_old = f.read()
+
+    insights_client.unregister()
+    assert not os.path.exists(MACHINE_ID_FILE)
+
+    insights_client.register()
+    with open(MACHINE_ID_FILE, "r") as f:
+        machine_id_new = f.read()
+
+    assert machine_id_new != machine_id_old
+
+
+def test_double_registration(insights_client):
+    """`--register` can be passed multiple times.
+
+    Even system that is already registered should allow the `--register` flag to be
+    passed in, without resulting in non-zero exit code.
+
+    This behavior has changed multiple times during the package lifetime.
+    """
+    assert not insights_client.is_registered
+
+    insights_client.register()
+    assert os.path.exists(MACHINE_ID_FILE)
+    with open(MACHINE_ID_FILE, "r") as f:
+        machine_id_old = f.read()
+
+    res = insights_client.register()
+    assert "This host has already been registered" in res.stdout
+    assert os.path.exists(MACHINE_ID_FILE)
+    with open(MACHINE_ID_FILE, "r") as f:
+        machine_id_new = f.read()
+
+    assert machine_id_new == machine_id_old


zpetrace:
The tests look fine from my side, it definitely should test the client side from the CCT-161 perspective enough so good by me.

integration-tests/test_registration.py
```diff
@@ -0,0 +1,70 @@
...
+    """`machine-id` is only present when insights-client is registered."""
+    assert not insights_client.is_registered
+    assert not os.path.exists(MACHINE_ID_FILE)
+
+    res = insights_client.run(check=False)
+    assert "This host has not been registered. Use --register to register this host." in res.stdout
+    assert not insights_client.is_registered
+    assert not os.path.exists(MACHINE_ID_FILE)
+
+    subman.register(
```
Archana-PandeyM, line 19: there is a session scoped fixture in conftest that registers subman at the beginning of test, you dont need to reregister. is its needed for some specific reason?

integration-tests/test_registration.py
```diff
@@ -0,0 +1,70 @@
...
+
+    insights_client.unregister()
+    assert not insights_client.is_registered
+    assert not os.path.exists(MACHINE_ID_FILE)
+
+
+@pytest.mark.skip("This has always been failing")
+def test_double_register(external_candlepin, subman, insights_client, test_config):
+    """`--register` fails on registered system."""
+    subman.register(
```
Archana-PandeyM, line 35: same applies here, No need to register subman, take a look here- https://github.com/RedHatInsights/insights-client/blob/master/integration-tests/conftest.py#L22.
you can use this fixture in your test like this https://github.com/RedHatInsights/insights-client/blob/master/integration-tests/test_status.py#L5

integration-tests/test_registration.py
m-horky: The test plan for CCT-161 included sub-man registrations, that's why I copied them over. Now that I'm reading it again, the RHSM registration doesn't change anything, we can start with it already being registered.

integration-tests/test_registration.py
```diff
@@ -0,0 +1,56 @@
...
+    assert not os.path.exists(MACHINE_ID_FILE)
+
+    insights_client.register()
+    assert os.path.exists(MACHINE_ID_FILE)
+
+    insights_client.unregister()
+    assert not os.path.exists(MACHINE_ID_FILE)
+
+
+@pytest.mark.skip("This has always been failing")
```
Archana-PandeyM, line 31: what is exactly failing always? Is it the 'register' operation second time? I see the expected message when registering twice"This host has already been registered" and exit code 0. This is the so far expected behavior and no bugs logged for exit code.

zpetrace: Hi, this test failed in the manual verification, you can take a look into the documentation of cct-161 manual verification on the version of insights:) that's why (I think) Matyas said that that's still failing as we both tried it and failed it for both of us while manually verifying it.

m-horky: `RHELPLAN-117674` should have more context. Maybe we should mention that somehow.

```py
@pytest.mark.skipped("Broken use-case: RHELPLAN-117674")
```

Archana-PandeyM: In my opinion I think we are confusing two different test cases here, the issue/bug you highlighted is applicable if insights is unregistered and machine-id exist

ptoscano: or rather using `pytest.mark.xfail()`, possibly with an explicit `strict=False`: this way, in case this is fixed in insights-core, old versions/tags of insights-client not will fail (and only get a non-fatal XPASS)

integration-tests/test_registration.py
```diff
@@ -0,0 +1,58 @@
...
+def test_double_register(insights_client):
+    """`--register` can be passed multiple times.
+
+    This behavior has changed multiple times during the package lifetime.
+    """
+    assert not insights_client.is_registered
+
+    insights_client.register()
+    assert os.path.exists(MACHINE_ID_FILE)
+
```
Archana-PandeyM, line 56: good to have verificatios - assert insights_client.is_registered

integration-tests/test_registration.py
```diff
@@ -0,0 +1,58 @@
...
+    with pytest.raises(subprocess.CalledProcessError):
+        insights_client.register()
```
Archana-PandeyM, line 58: also something like this -
register_status = insights_client.register()
assert "This host has already been registered" in register_status.stdout

integration-tests/test_registration.py
m-horky: That method will [raise an exception](https://github.com/ptoscano/pytest-client-tools/blob/main/pytest_client_tools/util.py#L96-L97) if it fails (unless `check=False` is passed), I don't think it is necessary.

integration-tests/test_registration.py
```diff
@@ -0,0 +1,58 @@
...
+
+    insights_client.register()
+    with open(MACHINE_ID_FILE, "r") as f:
+        machine_id_new = f.read()
+
+    assert machine_id_new != machine_id_old
+
+
+def test_double_register(insights_client):
+    """`--register` can be passed multiple times.
```
ptoscano, line 48: I'd expand the description here a bit, to be more explicit on what happens; something like:
```
    `--register` can be invoked without failures also on an already
    registered system
```

integration-tests/test_registration.py
```diff
@@ -0,0 +1,58 @@
...
+    assert not os.path.exists(MACHINE_ID_FILE)
+
+    insights_client.register()
+    with open(MACHINE_ID_FILE, "r") as f:
+        machine_id_new = f.read()
+
+    assert machine_id_new != machine_id_old
+
+
+def test_double_register(insights_client):
```
ptoscano, line 47: nitpick: `s/register/registration/` to be consistent with the name of the file in which this test is, and also with other tests in the same file

integration-tests/test_registration.py
```diff
@@ -0,0 +1,58 @@
...
+
+    This behavior has changed multiple times during the package lifetime.
+    """
+    assert not insights_client.is_registered
+
+    insights_client.register()
+    assert os.path.exists(MACHINE_ID_FILE)
+
+    res = insights_client.register()
+    assert "This host has already been registered" in res.stdout
```
ptoscano, line 58: in addition to checking that a double registration does not fail, should the test check also that
- the machine-id file still exists
- compare the machine-id file after the first registration and after the second for equality? ("similar" to what `test_machineid_changes_on_new_registration` does)

?

ptoscano (commented):
generally good, left some notes

zpetrace (approved):
looking good

Archana-PandeyM:
LGTM

ptoscano (approved):
LGTM, thanks!

Since Archana and Zdenek acked this too, merging.

PR #193, opened 2024-03-11
test: Adding tests for register operation

jstavel:
The PR provides tests for register command.

### DIFF
--- a/integration-tests/test_registration.py
+++ b/integration-tests/test_registration.py
@@ -3,11 +3,62 @@ import pytest
 import contextlib
 from pytest_client_tools.util import Version
 import conftest
-
+from constants import MACHINE_ID_FILE
 
 pytestmark = pytest.mark.usefixtures("register_subman")
 
-MACHINE_ID_FILE: str = "/etc/insights-client/machine-id"
+
+def test_register(insights_client):
+    """Test `insights-client --register` on an unregistered client"""
+    register_result = insights_client.run("--register")
+    assert insights_client.is_registered
+
+    assert "Starting to collect Insights data" in register_result.stdout
+    assert "Successfully uploaded report" in register_result.stdout
+    assert "View the Red Hat Insights console" in register_result.stdout
+
+
+def test_register_auth_proxy(insights_client, test_config):
+    """This test verifies that `insights-client --register` successfully
+    registers host when auth proxy is configured.
+    """
+    try:
+        proxy_host = test_config.get("auth_proxy", "host")
+        proxy_user = test_config.get("auth_proxy", "username")
+        proxy_pass = test_config.get("auth_proxy", "password")
+        proxy_port = str(test_config.get("auth_proxy", "port"))
+    except KeyError:
+        pytest.skip("Skipping because this test needs proxy settings to be configured")
+
+    auth_proxy = f"http://{proxy_user}:{proxy_pass}@{proxy_host}:{proxy_port}"
+
+    # save proxy information in insights-client.conf
+    insights_client.config.proxy = auth_proxy
+    insights_client.config.save()
+
+    register_result = insights_client.run("--register", "--verbose")
+    assert insights_client.is_registered
+    assert "Proxy Scheme: http://" in register_result.stdout
+    assert f"Proxy Location: {proxy_host}" in register_result.stdout
+    assert f"Proxy User: {proxy_user}" in register_result.stdout
+
+
+def test_register_noauth_proxy(insights_client, test_config):
+    """This test verifies that `insights-client --register` successfully
+    registers host when no-auth proxy is configured.
+    """
+    try:
+        proxy_host = test_config.get("noauth_proxy", "host")
+        proxy_port = str(test_config.get("noauth_proxy", "port"))
+    except KeyError:
+        pytest.skip("Skipping because this test needs proxy settings to be configured")
+    no_auth_proxy = f"http://{proxy_host}:{proxy_port}"
+    insights_client.config.proxy = no_auth_proxy
+    insights_client.config.save()
+
+    register_result = insights_client.run("--register", "--verbose")
+    assert insights_client.is_registered
+    assert f"CONF Proxy: {no_auth_proxy}" in register_result.stdout
 
 
 def test_machineid_exists_only_when_registered(insights_client):
--- a/integration-tests/constants.py
+++ b/integration-tests/constants.py
@@ -1,1 +1,2 @@
 HOST_DETAILS: str = "/var/lib/insights/host-details.json"
+MACHINE_ID_FILE: str = "/etc/insights-client/machine-id"


integration-tests/test_register.py
```diff
@@ -0,0 +1,47 @@
+import contextlib
+import pytest
+import logging
+
+logger = logging.getLogger(__name__)
```
ptoscano, line 5: this import && variables are not used

integration-tests/test_register.py
```diff
@@ -0,0 +1,47 @@
...
+pytestmark = pytest.mark.usefixtures("register_subman")
+
+
+def test_register(insights_client):
+    """
+    Test `insights-client --register` when the client is registered
+    """
+    response=insights_client.run('--register', check=False)
+    assert response.returncode == 0,\
+        "An application returns 0 after successful registration"
```
ptoscano, line 16: the default behaviour of `run()` is with `check=True`, so it can be dropped from the arguments; also, the `returncode` assertion for it can be dropped too

integration-tests/test_register.py
```diff
@@ -0,0 +1,47 @@
...
+    
+def test_register_twice(insights_client):
+    """
+    Test `insights-client --register` when the client is registered
+    """
+    insights_client.run('--register')
+    assert insights_client.is_registered
+    
+    response = insights_client.run('--register',check=False)
+    assert response.returncode == 0, "An application should return 0"
```
ptoscano, line 37: ditto

integration-tests/test_register.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def test_register(insights_client):
+    """
+    Test `insights-client --register` when the client is registered
+    """
+    response=insights_client.run('--register', check=False)
+    assert response.returncode == 0,\
+        "An application returns 0 after successful registration"
+    stdout = response.stdout.decode("UTF-8")
```
ptoscano, line 17: the default behaviour of `run()` for classes of tools in pytest-client-tools now is to use `universal_newlines=True`, so the `decode()` can be dropped

integration-tests/test_register.py
```diff
@@ -0,0 +1,47 @@
...
+def test_register_twice(insights_client):
+    """
+    Test `insights-client --register` when the client is registered
+    """
+    insights_client.run('--register')
+    assert insights_client.is_registered
+    
+    response = insights_client.run('--register',check=False)
+    assert response.returncode == 0, "An application should return 0"
+    stdout = response.stdout.decode("UTF-8")
```
ptoscano, line 38: ditto

ptoscano (changes requested):
Thanks @jstavel for the new tests, they look like a good start!

----

One change overall is to reformat the codebase using `black`: this will be in place really soon, so simply run `black` on your newly introduced file.

----

Adding labels for the assertions is good; OTOH we do not use them, as generally looking at the failures gives the assertion that fails (with the expanded variables). So i"d say to drop them.

----

Thanks!

integration-tests/test_register.py
jstavel: I'd like for a test to be more descriptive. At least when registering is the main point of the test. I want to describe behaviour of register task - and return code is a part of this behaviour.

integration-tests/test_register.py
jstavel: A way I wrote it is more descriptive and no hidden behavior is used. ie.  the test is rather a sort of documentation.

integration-tests/test_register.py
jstavel: removed.

integration-tests/test_register.py
```diff
@@ -0,0 +1,73 @@
...
+    ), "An application should inform that a system was already registered"
+    assert (
+        "Successfully uploaded report" in response.stdout
+    ), "An application should inform that report was successfully uploaded"
+    assert (
+        "View the Red Hat Insights console" in response.stdout
+    ), "An application should inform about Insights console"
+    assert (
+        insights_client.is_registered
+    ), "Current state of registration should be 'registered'"
```
ptoscano, line 54: this test is now covered by `test_double_registration` (provided by #191); if extra checks are needed, I'd add them to that test

ptoscano (changes requested):
thanks for the updates!

please note that, since #191 was merged, the tests related to the registration should go to `test_registration.py`; hence, please move your tests to that file

integration-tests/test_register.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+def test_register_using_noauth_proxy(insights_client, test_config):
+    proxy_config = functools.partial(test_config.get, "noauth_proxy")
```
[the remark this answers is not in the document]
jstavel: I think is is a matter of a few tests. ie. I think it should be applied just for a test. On the other hand it means that environmet settings are not true as it is - ie. if you take a look at environment section in .secrets.toml it comes to your mind that the current environment uses proxy. So it seems to be more consistent to have one config over all tests - ie. prod_with_auth_proxy, prod_with_noauth_proxy, prod. And let the right tests are run depending on the current environment. It is a matter of deeper discussion. I will ask team for the discussion about that.

integration-tests/test_register.py
jstavel: fixed. I used the logger to see what is the current proxy setting.

integration-tests/test_register.py
```diff
@@ -0,0 +1,73 @@
...
+    ), "An application should inform that Insights data will be collected"
+    assert (
+        "Successfully uploaded report" in response.stdout
+    ), "An application should inform that report was successfully uploaded"
+    assert (
+        "View the Red Hat Insights console" in response.stdout
+    ), "An application should inform about Insights console"
+    assert (
+        insights_client.is_registered
+    ), "Current state of registration should be 'registered'"
```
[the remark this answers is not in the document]
jstavel: I am trying to highlight the main sense of a verification. It helps in case the verification fails. If a string was changed - the assertation message helps me to decide whether the verification should be changed (ie. the behaviour comply the main sense of the verification) or it is the right failure.

integration-tests/test_register.py
jstavel: the test 'test_double_registration' plays with machine-id. It is the main point of the test - the machine-id is not changed anytime. A test in this PR verifies user interface behaviour rather. I think it makes sense to keep those two tests separated.

integration-tests/test_register.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+def test_register_using_noauth_proxy(insights_client, test_config):
+    proxy_config = functools.partial(test_config.get, "noauth_proxy")
+    proxy = f'http://{proxy_config("host")}:{proxy_config("port")}'
```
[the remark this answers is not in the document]
jstavel: I use http schema only. I think we do not have any test using https proxy anymore.

integration-tests/test_register.py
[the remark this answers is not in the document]
ptoscano: what you describe can be also done without that: the assertion will be shown, and you get the code around it anyway

integration-tests/test_register.py
ptoscano: > the test 'test_double_registration' plays with machine-id.

... because that was its first implementation, and there is nothing that limits it to machine-id checks; that tests is really about double registration, so it makes sense to have all the checks related to double registration attempts there: machine-id checks, output checks, etc

integration-tests/test_register.py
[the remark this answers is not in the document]
ptoscano: my point is that, if you need to use an https proxy for the tests (because it is configured as such), then you need a way to make the tests work with it

integration-tests/test_register.py
jstavel: merged.

integration-tests/test_register.py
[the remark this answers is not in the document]
jstavel: the message next to assertation provides you better error message in pytest overall report (in case test Failed).

integration-tests/test_register.py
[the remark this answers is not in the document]
jstavel: I've already merged those two files. 
I've finally merged test_register_twice with test_double_registration finally.

integration-tests/test_registration.py
```diff
@@ -1,22 +1,91 @@
...
     """`machine-id` is only present when insights-client is registered."""
     assert not insights_client.is_registered
     assert not os.path.exists(MACHINE_ID_FILE)
 
     res = insights_client.run(check=False)
     assert (
-        "This host has not been registered. Use --register to register this host."
-    ) in res.stdout
+        "This machine has not yet been registered. Use --register to register"
+    ) in res.stdout, "An application should ask a user to register a system"
```
ptoscano, line 88: the message I get when testing actually mentions "host", so I'm not sure why this assert was changed (the commit message does not mention anything about it)

integration-tests/test_register.py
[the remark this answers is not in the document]
ptoscano: as I said, these keys are not supported by pytest-client-tools at all ATM, so using them like this in the test is a no-go

integration-tests/test_registration.py
```diff
@@ -1,22 +1,91 @@
-import os
-
```
ptoscano, line 2: better avoid this kind of style change in the same commit -- simply add the other stdlib imports here, rather than reodering things

ptoscano (changes requested):
thanks for the updates; I left few inline notes

As I mentioned as well, the two new proxy tests (ie `test_register_using_noauth_proxy` and `test_register_using_auth_proxy`) try to use configuration bits that are not currently implemented/available in pytest-client-tools; while we discuss how to implement them, what would be their semantics and so on, in the spirit of not blocking the other two tests added here I recommend to split these new proxy-related tests in its own PR, and keep the other changes here. WDYT?

integration-tests/requirements.txt
```diff
@@ -1,2 +1,4 @@
 git+https://github.com/ptoscano/pytest-client-tools@main
 pyyaml
+black
+flake8
```
ptoscano, line 4: these modules are not needed for running the tests

ptoscano (changes requested):
See also my previous notes.

integration-tests/test_registration.py
```diff
@@ -1,22 +1,93 @@
...
+    """
+    Test `insights-client --register` when the client is registered
+    """
+    response = insights_client.run("--register", check=False)
+    assert (
+        response.returncode == 0
+    ), "An application returns 0 after successful registration"
+    assert (
+        "Successfully registered host" in response.stdout
+    ), "An application should inform a user that a system was registered"
```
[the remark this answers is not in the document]
Archana-PandeyM: I am removing this check as the line only appear in legacy_upload True

Archana-PandeyM:
@jstavel I have modified this PR to include changes related to only --register + All review comment. All the improvement/optimization for existing code should be done in a new PR.

m-horky (changes requested):
Could you please rebase over `master` and squash your commits into one? Let's ensure we have clean git history.

LGTM for content.

Archana-PandeyM:
@m-horky I have rebased it now and commit already squashed. Please confirm again. Let me know if I missed anything

m-horky (approved):

PR #194, opened 2024-03-11
feat: Debug initialization

Glutexo:
[CCT-372](https://issues.redhat.com/browse/CCT-372)

This pull request improves the debugging experience of the the egg selection and core initialization process by adding more verbosity. It is now clear what egg has been selected, what commands are used to run the core phases as well as what happend in case of an error. GPG verfification is verbosely logged, too.

A proper logger is initialized instead of using bare prints. It uses the same message format as the core and is controlled by the same CLI flags (`--verbose` and `--silent`) and corresponding environment variables. All log statements are called in a unified way. The temporary logger is torn down just before the core initializes its own one.

Steps to verify:

1. Run the client with a `--verbose` flag.
   ```
   # PYTHONPATH=src builddir/src/insights-client --verbose
   ```
3. See the messages formatted in a core-like way, including a line number.
   ```
   2024-05-06 08:57:04,787    DEBUG insights_client:109 Running command: PYTHONPATH=/var/lib/insights/last_stable.egg PATH=/sbin:/bin:/usr/sbin:/usr/bin /usr/bin/python3 -c from insights.client import InsightsClient; print(InsightsClient(None, False).version())
   ```
4. See the new messages and make sure it is clear what egg is being used.
   ```
   2024-05-06 08:57:06,304    DEBUG insights_client:460 InsightsClient initialized. Egg version: 3.3.19-1
   ...
   2024-05-06 08:57:06,326    DEBUG insights_client phase 'pre_update'; egg '/var/lib/insights/last_stable.egg'
   ```
5. Do the same with the environment variable.
   ```
   # PYTHONPATH=src INSIGHTS_VERBOSE=True builddir/src/insights-client 
   ```
6. Verify that the core log messages are not printed twice.
   ```
   2024-05-06 09:14:22,646    DEBUG insights.client.client Logging initialized
   2024-05-06 09:14:22,648    DEBUG insights.client.auto_config Trying to autoconfigure...
   2024-05-06 09:14:22,659    DEBUG rhsm.https Using standard libs to provide httplib and ssl
   ```
7. Run the client with a `--silent` flag or an `INSIGHTS_SILENT=True` environment variable.
   ```
   # PYTHONPATH=src builddir/src/insights-client --silent
   ```
8. See that no log messages are printed, not even registration warnings or collection status.
9. Run the client both with a verbose and silent flag in all possible flag/variable combinations.
10. See that the silent flag always has precedence and no messages are printed.

### DIFF
--- a/src/insights_client/__init__.py
+++ b/src/insights_client/__init__.py
@@ -16,23 +16,7 @@ import tempfile
 from distutils.version import LooseVersion
 
 
-def get_logging_config():
-    config = {}
-
-    for arg in ["silent", "verbose"]:
-        environ_variable = f"INSIGHTS_{arg.upper()}"
-        environ_value = os.environ.get(environ_variable, "")
-
-        if environ_value.lower() == "true":
-            config[arg] = True
-        else:
-            cli_flag = f"--{arg}"
-            config[arg] = cli_flag in sys.argv
-
-    return config
-
-
-LOGGING_CONFIG = get_logging_config()
+LOG_FORMAT = "%(asctime)s %(levelname)8s %(name)s:%(lineno)s %(message)s"
 NO_COLOR = os.environ.get("NO_COLOR") is not None
 
 BYPASS_GPG = os.environ.get("BYPASS_GPG", "").lower() == "true"
@@ -54,33 +38,40 @@ TEMPORARY_GPG_HOME_PARENT_DIRECTORY = "/var/lib/insights/"
 logger = logging.getLogger(__name__)
 
 
-def client_debug(message):
-    """Debug a log message when logging isn't available yet.
+def get_logging_config():
+    config = {}
 
-    Set 'INSIGHTS_DEBUG' variable to enable this method.
+    for arg in ["silent", "verbose"]:
+        environ_variable = f"INSIGHTS_{arg.upper()}"
+        environ_value = os.environ.get(environ_variable, "")
 
-    :param message: Text to display
-    :type message: str
-    """
-    if LOGGING_CONFIG["silent"]:
+        if environ_value.lower() == "true":
+            config[arg] = True
+        else:
+            cli_flag = f"--{arg}"
+            config[arg] = cli_flag in sys.argv
+
+    return config
+
+
+def set_up_logging(logging_config):
+    if logging_config["silent"]:
         logger.setLevel(logging.FATAL)
         return
-    elif not LOGGING_CONFIG["verbose"]:
+    elif not logging_config["verbose"]:
         return
 
-    prefix = "insights-client debug"
-    suffix = ""
-    if NO_COLOR or not sys.stdout.isatty():
-        prefix += ":"
-    else:
-        prefix = "\033[40m" + prefix + "\033[0m\033[33m"
-        suffix = "\033[0m"
+    logger.setLevel(logging.DEBUG)
 
-    print(prefix + " " + message + suffix, file=sys.stderr)
+    handler = logging.StreamHandler(sys.stdout)
+    formatter = logging.Formatter(LOG_FORMAT)
+    handler.setFormatter(formatter)
+    logger.addHandler(handler)
 
 
-def log(msg):
-    print(msg, file=sys.stderr)
+def tear_down_logging():
+    for handler in logger.handlers:
+        logger.removeHandler(handler)
 
 
 def join_path(parts):
@@ -148,15 +139,14 @@ def _remove_gpg_home(home):
     )
     stdout, stderr = shutdown_process.communicate()
     if shutdown_process.returncode != 0:
-        log(
-            "Could not kill the GPG agent, got return code {rc}".format(
-                rc=shutdown_process.returncode
-            )
+        logger.debug(
+            "Could not kill the GPG agent, got return code %d",
+            shutdown_process.returncode,
         )
         if stdout:
-            log(stdout)
+            logger.debug(stdout)
         if stderr:
-            log(stderr)
+            logger.debug(stderr)
 
     # Delete the temporary directory
     shutil.rmtree(home)
@@ -175,26 +165,19 @@ def gpg_validate(path):
     """
     # EGG= may be None or an invalid path
     if not path or not os.path.exists(path):
-        client_debug(
-            "Path '{path}' does not exist and cannot be GPG validated.".format(
-                path=path
-            )
-        )
+        logger.debug("Path '%s' does not exist and cannot be GPG validated.", path)
         return False
 
     # EGG may be a path to a directory (which cannot be signed)
     if BYPASS_GPG:
-        client_debug(
-            "'BYPASS_GPG' is set, pretending the GPG validation of '{path}' succeeded.".format(
-                path=path
-            )
+        logger.debug(
+            "'BYPASS_GPG' is set, pretending the GPG validation of '%s' succeeded.",
+            path,
         )
         return True
 
     if not os.path.exists(path + ".asc"):
-        client_debug(
-            "Path '{path}' does not have an associated '.asc' file.".format(path=path)
-        )
+        logger.debug("Path '%s' does not have an associated '.asc' file.", path)
         return False
 
     # The /var/lib/insights/ directory is used instead of /tmp/ because
@@ -222,11 +205,10 @@ def gpg_validate(path):
     verify_process.communicate()
     _remove_gpg_home(home)
 
-    client_debug(
-        "The GPG verification of '{path}' returned status code {code}.".format(
-            path=path,
-            code=verify_process.returncode,
-        )
+    logger.debug(
+        "The GPG verification of '%s' returned status code %d.",
+        path,
+        verify_process.returncode,
     )
     return verify_process.returncode == 0
 
@@ -245,23 +227,23 @@ def run_phase(phase, client, validated_eggs):
     all_eggs = [NEW_EGG] + validated_eggs
     if ENV_EGG is not None:
         all_eggs = [ENV_EGG] + all_eggs
-        client_debug("Environment egg defined as %s." % ENV_EGG)
+        logger.debug("Environment egg defined as %s." % ENV_EGG)
 
     for egg in all_eggs:
         if config["gpg"]:
             if not os.path.isfile(egg):
-                client_debug("%s is not a file, can't GPG verify. Skipping." % (egg,))
+                logger.debug("%s is not a file, can't GPG verify. Skipping.", egg)
                 continue
-            client_debug("Verifying %s..." % egg)
+            logger.debug("Verifying %s..." % egg)
             if not client.verify(egg)["gpg"]:
-                client_debug(
-                    "WARNING: GPG verification failed. Not loading egg: %s" % egg
+                logger.debug(
+                    "WARNING: GPG verification failed. Not loading egg: %s", egg
                 )
                 continue
         else:
-            client_debug("GPG disabled by --no-gpg, not verifying %s." % egg)
+            logger.debug("GPG disabled by --no-gpg, not verifying %s.", egg)
 
-        client_debug("phase '%s'; egg '%s'" % (phase["name"], egg))
+        logger.debug("phase '%s'; egg '%s'", phase["name"], egg)
 
         # prepare the environment
         pythonpath = str(egg)
@@ -279,14 +261,12 @@ def run_phase(phase, client, validated_eggs):
         stdout, stderr = process.communicate()
         if process.returncode == 0:
             # phase successful, don't try another egg
-            client_debug("phase '{phase}' successful".format(phase=phase["name"]))
+            logger.debug("phase '%s' successful", phase["name"])
             update_motd_message()
             return
 
-        client_debug(
-            "phase '{phase}' failed with return code {rc}".format(
-                phase=phase["name"], rc=process.returncode
-            )
+        logger.debug(
+            "phase '%s' failed with return code %d", phase["name"], process.returncode
         )
         if process.returncode == 1:
             # egg hit an error, try the next
@@ -320,9 +300,8 @@ def update_motd_message():
     """
     if not os.path.exists(os.path.dirname(MOTD_FILE)):
         logger.debug(
-            "directory '{dir}' does not exist, ignoring MOTD update request".format(
-                dir=os.path.dirname(MOTD_FILE)
-            )
+            "directory '%s' does not exist, ignoring MOTD update request",
+            os.path.dirname(MOTD_FILE),
         )
         return
 
@@ -339,47 +318,44 @@ def update_motd_message():
         if not os.path.lexists(MOTD_FILE):
             logger.debug(
                 ".registered and .unregistered do not exist; "
-                "pointing the MOTD file '{source}' to '{motd}'".format(
-                    source=MOTD_SRC, motd=MOTD_FILE
-                )
+                "pointing the MOTD file '%s' to '%s'",
+                MOTD_SRC,
+                MOTD_FILE,
             )
             try:
                 os.symlink(MOTD_SRC, MOTD_FILE)
             except OSError as exc:
                 logger.debug(
-                    "could not point the MOTD file '{source}' to '{motd}': {exc}".format(
-                        source=MOTD_SRC, motd=MOTD_FILE, exc=exc
-                    )
+                    "could not point the MOTD file '%s' to '%s': %s",
+                    MOTD_SRC,
+                    MOTD_FILE,
+                    exc,
                 )
         else:
             logger.debug(
                 ".registered and .unregistered do not exist; "
-                "file '{source}' correctly points to '{motd}'".format(
-                    source=MOTD_SRC, motd=MOTD_FILE
-                )
+                "file '%s' correctly points to '%s'",
+                MOTD_SRC,
+                MOTD_FILE,
             )
 
     else:
         # .registered or .unregistered exist, MOTD should not be displayed
         if os.path.lexists(MOTD_FILE):
             logger.debug(
-                ".registered or .unregistered exist; removing the MOTD file '{path}'".format(
-                    path=MOTD_FILE
-                )
+                ".registered or .unregistered exist; removing the MOTD file '{path}'",
+                MOTD_FILE,
             )
             try:
                 os.remove(MOTD_FILE)
             except OSError as exc:
                 logger.debug(
-                    "could not remove the MOTD file '{path}': {exc}".format(
-                        path=MOTD_FILE, exc=exc
-                    )
+                    "could not remove the MOTD file '{path}': {exc}", MOTD_FILE, exc
                 )
         else:
             logger.debug(
-                ".registered or .unregistered exist; file '{motd}' correctly does not exist".format(
-                    motd=MOTD_FILE
-                )
+                ".registered or .unregistered exist; file '{motd}' correctly does not exist",
+                MOTD_FILE,
             )
 
 
@@ -389,6 +365,9 @@ def _main():
     attempt to collect and upload with new, then current, then rpm
     if an egg fails a phase never try it again
     """
+    logging_config = get_logging_config()
+    set_up_logging(logging_config)
+
     # sort rpm and stable eggs after verification
     validated_eggs = sorted_eggs(list(filter(gpg_validate, [STABLE_EGG, RPM_EGG])))
     # if ENV_EGG was specified and it's valid, add that to front of sys.path
@@ -404,15 +383,14 @@ def _main():
 
     # ENV egg comes first
     all_valid_eggs = valid_env_egg + validated_eggs
-    client_debug("Using eggs: %s", join_path(all_valid_eggs))
+    logger.debug("Using eggs: %s", join_path(all_valid_eggs))
     sys.path = all_valid_eggs + sys.path
 
     try:
         # flake8 complains because these imports aren't at the top
         import insights
 
-        initial_egg_path = egg_path(insights)
-        client_debug("Loaded initial egg: %s" % (initial_egg_path,))
+        logger.debug("Loaded initial egg: %s", egg_path(insights))
 
         from insights.client import InsightsClient
         from insights.client.phase.v1 import get_phases
@@ -443,11 +421,10 @@ def _main():
         # handle client instantation here so that it isn't done multiple times in __init__
         # The config can be passed now by parameter
         client = InsightsClient(config, False)  # read config, but dont setup logging
-        client_debug(
-            "InsightsClient initialized. Egg version: %s" % (client.version(),)
-        )
+        logger.debug("InsightsClient initialized. Egg version: %s", client.version())
 
-        # we now have access to the clients logging mechanism instead of using print
+        # we now have access to the clients logging mechanism
+        tear_down_logging()
         client.set_up_logging()
 
         for p in get_phases():
--- a/src/insights_client/__init__.py
+++ b/src/insights_client/__init__.py
@@ -16,7 +16,23 @@ import tempfile
 from distutils.version import LooseVersion
 
 
-INSIGHTS_DEBUG = os.environ.get("INSIGHTS_DEBUG", "").lower() == "true"
+def get_logging_config():
+    config = {}
+
+    for arg in ["silent", "verbose"]:
+        environ_variable = f"INSIGHTS_{arg.upper()}"
+        environ_value = os.environ.get(environ_variable, "")
+
+        if environ_value.lower() == "true":
+            config[arg] = True
+        else:
+            cli_flag = f"--{arg}"
+            config[arg] = cli_flag in sys.argv
+
+    return config
+
+
+LOGGING_CONFIG = get_logging_config()
 NO_COLOR = os.environ.get("NO_COLOR") is not None
 
 BYPASS_GPG = os.environ.get("BYPASS_GPG", "").lower() == "true"
@@ -46,7 +62,10 @@ def client_debug(message):
     :param message: Text to display
     :type message: str
     """
-    if not INSIGHTS_DEBUG:
+    if LOGGING_CONFIG["silent"]:
+        logger.setLevel(logging.FATAL)
+        return
+    elif not LOGGING_CONFIG["verbose"]:
         return
 
     prefix = "insights-client debug"
@@ -370,7 +389,6 @@ def _main():
     attempt to collect and upload with new, then current, then rpm
     if an egg fails a phase never try it again
     """
-
     # sort rpm and stable eggs after verification
     validated_eggs = sorted_eggs(list(filter(gpg_validate, [STABLE_EGG, RPM_EGG])))
     # if ENV_EGG was specified and it's valid, add that to front of sys.path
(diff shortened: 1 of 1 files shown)


Glutexo:
Do we want to use _client_debug_ after the Client logger has been initialized? It would make sense to switch to `logging.[root.]debug`. I was thinking about logging something in the lines of:

> Setting up regular logging. Use --verbose to see the debug messages from here on.

There are already regular log calls and it is confusing where to use `client_debug` and when to use `logging.debug`. I originally thought the purpose of `client_debug` is to have debug statements when we don’t have a working logger yet.

src/insights_client/__init__.py
```diff
@@ -147,58 +140,55 @@ def gpg_validate(path):
...
         return False
 
     # EGG may be a path to a directory (which cannot be signed)
     if BYPASS_GPG:
-        client_debug(
-            "'BYPASS_GPG' is set, pretending the GPG validation of '{path}' succeeded.".format(
-                path=path
-            )
+        logger.debug(
+            "'BYPASS_GPG' is set, pretending the GPG validation of '%s' succeeded."
```
m-horky, line 152: Invalid syntax. Should be `.format(path)` or similar.

src/insights_client/__init__.py
```diff
@@ -16,8 +16,8 @@
 from distutils.version import LooseVersion
 
 
-INSIGHTS_DEBUG = os.environ.get("INSIGHTS_DEBUG", "").lower() == "true"
-NO_COLOR = os.environ.get("NO_COLOR") is not None
+LOG_FORMAT = "%(asctime)s %(levelname)8s %(name)s:%(lineno)s %(message)s"
+INSIGHTS_VERBOSE = os.environ.get("INSIGHTS_VERBOSE", "").lower() == "true"
```
m-horky, line 20: Since the Core detects both this and `--verbose` flag, should we do the same here?

```py
INSIGHTS_VERBOSE = os.environ.get(...) == "true" or "--verbose" in sys.argv
```

src/insights_client/__init__.py
```diff
@@ -38,30 +38,21 @@
...
+    handler = logging.StreamHandler(sys.stdout)
+    if INSIGHTS_VERBOSE:
+        logger.setLevel(logging.DEBUG)
+        log_format = LOG_FORMAT
     else:
-        prefix = "\033[40m" + prefix + "\033[0m\033[33m"
-        suffix = "\033[0m"
-
-    print(prefix + " " + message + suffix, file=sys.stderr)
+        log_format = "%(message)s"
```
m-horky, line 47: If we know we don't want to log anything from this module, should this function just `return None` directly?

src/insights_client/__init__.py
```diff
@@ -147,58 +140,55 @@ def gpg_validate(path):
...
+            path
         )
         return False
 
     # The /var/lib/insights/ directory is used instead of /tmp/ because
     # GPG needs to have RW permissions in it, and existing SELinux rules only
     # allow access here.
+    logger.debug("Creating temporary directory in %s...", TEMPORARY_GPG_HOME_PARENT_DIRECTORY)
     home = tempfile.mkdtemp(dir=TEMPORARY_GPG_HOME_PARENT_DIRECTORY)
+    logger.debug("Created temporary directory: %s", home)
```
m-horky, line 169: I don't think we have to log every statement in the happy path, it will be too verbose. The logger statement two lines above is enough.

src/insights_client/__init__.py
```diff
@@ -214,16 +204,26 @@ def run_phase(phase, client, validated_eggs):
     ] + sys.argv[1:]
     config = client.get_conf()
 
-    all_eggs = [ENV_EGG, NEW_EGG] + validated_eggs
+    all_eggs = [NEW_EGG] + validated_eggs
+    if ENV_EGG is not None:
+        all_eggs = [ENV_EGG] + all_eggs
+        logger.debug("ENV_EGG defined as %s." % ENV_EGG)
+    else:
+        logger.debug("ENV_EGG not defined, skipping.")
```
m-horky, line 212: I don't think we need to log this.

src/insights_client/__init__.py
```diff
@@ -214,16 +204,26 @@ def run_phase(phase, client, validated_eggs):
     ] + sys.argv[1:]
     config = client.get_conf()
 
-    all_eggs = [ENV_EGG, NEW_EGG] + validated_eggs
+    all_eggs = [NEW_EGG] + validated_eggs
+    if ENV_EGG is not None:
+        all_eggs = [ENV_EGG] + all_eggs
+        logger.debug("ENV_EGG defined as %s." % ENV_EGG)
```
m-horky, line 210: The name of the variable is implementation detail. I'd do something like

```py
logger.debug("Using environment egg '%s'.".format(ENV_EGG))
```

instead.

src/insights_client/__init__.py
```diff
@@ -214,16 +204,26 @@ def run_phase(phase, client, validated_eggs):
...
+    else:
+        logger.debug("ENV_EGG not defined, skipping.")
+
+    for egg in all_eggs:
+        if config["gpg"]:
+            if not os.path.isfile(egg):
+                logger.debug("%s is not a file, can't GPG verify. Skipping." % (egg,))
+                continue
+            logger.debug("Verifying %s..." % egg)
```
m-horky, line 219: We shouldn't need this. We get the egg information in both the happy and sad path.
- If the validation fails, it's logged on line 221.
- If it succeeds, we'll get the log on line 226.

src/insights_client/__init__.py
```diff
@@ -301,47 +300,46 @@ def update_motd_message():
         if not os.path.lexists(MOTD_FILE):
             logger.debug(
                 ".registered and .unregistered do not exist; "
-                "pointing the MOTD file '{source}' to '{motd}'".format(
-                    source=MOTD_SRC, motd=MOTD_FILE
-                )
+                "pointing the MOTD file '%s' to '%s'",
+                MOTD_SRC,
+                MOTD_FILE
```
m-horky, line 305: What does this change achieve?

src/insights_client/__init__.py
```diff
@@ -301,47 +300,46 @@ def update_motd_message():
...
                 os.symlink(MOTD_SRC, MOTD_FILE)
             except OSError as exc:
                 logger.debug(
-                    "could not point the MOTD file '{source}' to '{motd}': {exc}".format(
-                        source=MOTD_SRC, motd=MOTD_FILE, exc=exc
-                    )
+                    "could not point the MOTD file '%s' to '%s': %s",
+                    MOTD_SRC,
+                    MOTD_FILE,
+                    exc
```
m-horky, line 314: Same as above

src/insights_client/__init__.py
```diff
@@ -351,37 +349,62 @@ def _main():
     attempt to collect and upload with new, then current, then rpm
     if an egg fails a phase never try it again
     """
+    set_up_logging()
+    logger.debug("Verifying initial eggs to load config.")
```
m-horky, line 353: I don't think this log statement carries any value.

src/insights_client/__init__.py
```diff
@@ -351,37 +349,62 @@ def _main():
...
+
+    validated_eggs = []
+    for egg in [STABLE_EGG, RPM_EGG]:
+        logger.debug("GPG verifying initial egg %s..." % (egg,))
+        if gpg_validate(egg):
+            logger.debug("GPG verification successful. Loading initial egg: %s" % (egg,))
```
m-horky, line 359: We're not loading it, it will be considered when we run the code. We're getting the information about it later as well.

src/insights_client/__init__.py
```diff
@@ -351,37 +349,62 @@ def _main():
...
+            logger.debug("WARNING: GPG verification failed. Not loading initial egg: %s" % (egg,))
     # sort rpm and stable eggs after verification
-    validated_eggs = sorted_eggs(list(filter(gpg_validate, [STABLE_EGG, RPM_EGG])))
+    validated_eggs = sorted_eggs(validated_eggs)
     # if ENV_EGG was specified and it's valid, add that to front of sys.path
     #  so it can be loaded initially. keep it in its own var so we don't
     #  pass it to run_phase where we load it again
+    logger.debug("Verifying initial ENV_EGG: %s..." % (ENV_EGG,))
     if gpg_validate(ENV_EGG):
+        logger.debug("GPG verification successful. Loading initial egg: %s" % (ENV_EGG,))
```
m-horky, line 370: Same as above. We're not loading it.

src/insights_client/__init__.py
```diff
@@ -351,37 +349,62 @@ def _main():
...
+    logger.debug("Verifying initial eggs to load config.")
+
+    validated_eggs = []
+    for egg in [STABLE_EGG, RPM_EGG]:
+        logger.debug("GPG verifying initial egg %s..." % (egg,))
+        if gpg_validate(egg):
+            logger.debug("GPG verification successful. Loading initial egg: %s" % (egg,))
+            validated_eggs.append(egg)
+        else:
+            logger.debug("WARNING: GPG verification failed. Not loading initial egg: %s" % (egg,))
```
m-horky, line 362: The `gpg_validate` already logs if the gpg passed successfully or not.

src/insights_client/__init__.py
```diff
@@ -351,37 +349,62 @@ def _main():
...
     if not validated_eggs and not valid_env_egg:
-        sys.exit("No GPG-verified eggs can be found")
+        sys.exit("No GPG-verified initial eggs can be found")
 
     # ENV egg comes first
     sys.path = valid_env_egg + validated_eggs + sys.path
 
     try:
+        sys_path = ":".join(sys.path)
+        logger.debug("Importing the insights module from %s..." % (sys_path,))
```
m-horky, line 384: I have my doubts about the benefit of logging full PATH, but I am not against either.

Non-blocking comment.

src/insights_client/__init__.py
```diff
@@ -351,37 +349,62 @@ def _main():
...
+        logger.debug("Loaded initial egg: %s" % ( egg_path,))
+
         from insights.client import InsightsClient
         from insights.client.phase.v1 import get_phases
         from insights.client.config import InsightsConfig
 
         # Add the insights-config here
         try:
-            config = InsightsConfig(_print_errors=True).load_all()
+            logger.debug("Initializing InsightsConfig.")
```
m-horky, line 396: I don't think this is interesting log information.

src/insights_client/__init__.py
```diff
@@ -351,37 +349,62 @@ def _main():
...
+            config_args = {}
+            if INSIGHTS_VERBOSE:
+                config_args["verbose"] = True
+            config = InsightsConfig(_print_errors=True, **config_args).load_all()
         except ValueError as e:
             sys.stderr.write("ERROR: " + str(e) + "\n")
             sys.exit("Unable to load Insights Config")
 
         if config["version"]:
+            logger.debug("Only printing version, skipping regular Client run.")
```
m-horky, line 407: From the log standpoint this is not very interesting to know.

src/insights_client/__init__.py
```diff
@@ -398,13 +421,17 @@ class InsightsConstants(object):
 
         # handle client instantation here so that it isn't done multiple times in __init__
         # The config can be passed now by parameter
+        logger.debug("Initializing InsightsClient with the loaded config...")
         client = InsightsClient(config, False)  # read config, but dont setup logging
+        logger.debug("InsightsClient initialized. Egg version: %s" % (client.version(),))
```
m-horky, line 426: From these I would only log the second one.

```py
logger.debug("Client initialized with egg '%s'.", client.version())
```

src/insights_client/__init__.py
```diff
@@ -398,13 +421,17 @@ class InsightsConstants(object):
...
 
         # we now have access to the clients logging mechanism instead of using print
+        logger.debug("Setting up regular logging.")
```
m-horky, line 429: This is not interesting enough to log.

src/insights_client/__init__.py
```diff
@@ -398,13 +421,17 @@ class InsightsConstants(object):
...
+        tear_down_logging()
         client.set_up_logging()
-        logging.root.debug("Loaded initial egg: %s", os.path.dirname(insights.__file__))
 
         for p in get_phases():
+            logger.debug("Running phase %s." % (p["name"],))
```
m-horky, line 434: Nitpick: Why do you keep mixing two different ways of formatting things?
```py
logger.debug("Running phase %s." % (p["name"],))
logger.debug("Running phase %s.", p["name"])

logger.debug("Running phase '{phase}'.".format(phase=p["name"]))
```

m-horky (changes requested):

Glutexo:
Thanks for your review! I pushed some more commits and would consider the code feature complete. I am now going to address your comments, update the branch, squash the commits and fill in commit messages and PR description. Removing the draft flag from the PR.

src/insights_client/__init__.py
Glutexo: Agreed. It makes sense to not setup stuff if we don’t want to use it. Also we’d get rid of this explicit re-stating of the default format.

src/insights_client/__init__.py
Glutexo: Although it’s not the most interesting piece of information, I think it still brings some value. I can imagine a case when you think you have your ENV_EGG set, but it doesn’t propagate to the actual environment the Client is run in.

src/insights_client/__init__.py
Glutexo: Or even better

```suggestion
        logger.debug("Using environment egg %s.", ENV_EGG)
```

The Python variable name is an implementation detail, but the environment variable name isn’t – that’s an API. But that’s EGG, not ENV_EGG and “EGG defined as %s.” sounds rather cryptic.

Still, however, even a Python variable name may make sense in a debug message if its audience is the developers. But it’s “verbose”, not “debug”, so it sounds more like for power users rather than developers. 👍🏻

src/insights_client/__init__.py
Glutexo: Although logging before and after may look redundant, I see the advantage of having a pre-message like this that in case of 💥, it is the last thing printed. A failure message may not be logged in case the 💥 is too bad so that it’s not rescued from gracefully. 

The success message is probably more useless, but it nicely pairs with the pre-message. Especially for heavy operations like spawning a subprocess.

src/insights_client/__init__.py
Glutexo: It is maybe a bit verbose. But when debugging why was a module loaded from a certain location, the full PATH may be invaluable.

src/insights_client/__init__.py
Glutexo: 🤔  Well, the InsightsConfig initialization is pretty heavy. It provides an information that the modules are imported and we’re going to actually initialize the class that parses environment values, a config file and command line arguments, merges them etc. Again, it’s the last message you’ll see if the config initialization fails.

src/insights_client/__init__.py
Glutexo: Probably. 🤷🏻‍♂️ It can be just a comment, or nothing at all.

src/insights_client/__init__.py
Glutexo: From a developer perspective, it may be. But it probably doesn’t deserve to be printed just for _verbose_.

src/insights_client/__init__.py
Glutexo: I wanted to unify this, leveraging the native logging interpolation. For some reason, I didn’t do it consistenly. Good catch! 🙇🏻‍♂️

src/insights_client/__init__.py
Glutexo: It unifies the log formatting syntax, questioned by your later [comment](https://github.com/RedHatInsights/insights-client/pull/194#discussion_r1549333579). It also leverages the native logging module interpolation.

src/insights_client/__init__.py
Glutexo: Right. The information is redundant. Only the latter part of both log messages brings new information.

src/insights_client/__init__.py
Glutexo: Do you think the first message is not useful? _InsightsClient_ initialization is a big thing and in case of failure, it may be nice to see the first message as the pointer to the last place where things we ok.

src/insights_client/__init__.py
```diff
@@ -208,47 +230,62 @@ def run_phase(phase, client, validated_eggs):
...
         env.update(insights_env)
 
+        current_path = os.path.dirname(__file__)
+        logger.debug("Current path: %s", current_path)
+
+        insights_command = [
+            sys.executable,
+            os.path.join(current_path, "run.py"),
+        ] + sys.argv[1:]
+        logger.debug("Running command: %s", debug_command(insights_command, env))
```
m-horky, line 278: I would drop the environment, or maybe the whole line. It should be clear from other logs what we end up running.

src/insights_client/__init__.py
```diff
@@ -208,47 +230,62 @@ def run_phase(phase, client, validated_eggs):
...
             "INSIGHTS_PHASE": str(phase["name"]),
             "PYTHONPATH": pythonpath,
         }
-        env = os.environ
+        logger.debug("Adding to the environment: %s", debug_environ(insights_env))
+
         env.update(insights_env)
 
+        current_path = os.path.dirname(__file__)
+        logger.debug("Current path: %s", current_path)
```
m-horky, line 272: Is this log statement telling us something useful?

src/insights_client/__init__.py
```diff
@@ -208,47 +230,62 @@ def run_phase(phase, client, validated_eggs):
...
-        client_debug("phase '%s'; egg '%s'" % (phase["name"], egg))
+        logger.debug("phase '%s'; egg '%s'", phase["name"], egg)
 
         # prepare the environment
+        env = os.environ
+        logger.debug("Existing enviroment: %s", debug_environ(env))
+
         pythonpath = str(egg)
-        env_pythonpath = os.environ.get("PYTHONPATH", "")  # type: str
+        logger.debug("Adding to PYTHONPATH: %s", pythonpath)
```
m-horky, line 258: This is implementation detail, we don't have to log this. It will simply not work without it.

src/insights_client/__init__.py
```diff
@@ -208,47 +230,62 @@ def run_phase(phase, client, validated_eggs):
...
+        env_pythonpath = env.get("PYTHONPATH", "")  # type: str
+
         if env_pythonpath:
             pythonpath += ":" + env_pythonpath
         insights_env = {
             "INSIGHTS_PHASE": str(phase["name"]),
             "PYTHONPATH": pythonpath,
         }
-        env = os.environ
+        logger.debug("Adding to the environment: %s", debug_environ(insights_env))
```
m-horky, line 267: Again, this should be implicit from previous log statements. However, if we wanted to log the values of those two variables, it should be done before we subprocess to run `run.py`.

src/insights_client/__init__.py
```diff
@@ -399,12 +458,14 @@ class InsightsConstants(object):
...
+        logger.debug("InsightsClient initialized. Egg version: %s", client.version())
 
-        # we now have access to the clients logging mechanism instead of using print
+        # we now have access to the clients logging mechanism
+        tear_down_logging()
         client.set_up_logging()
-        logging.root.debug("Loaded initial egg: %s", os.path.dirname(insights.__file__))
 
         for p in get_phases():
+            logger.debug("Running phase %s.", p["name"])
```
m-horky, line 468: This information is already logged on line 251.

src/insights_client/__init__.py
```diff
@@ -208,47 +230,62 @@ def run_phase(phase, client, validated_eggs):
...
             update_motd_message()
             return
 
-        client_debug(
-            "phase '{phase}' failed with return code {rc}".format(
-                phase=phase["name"], rc=process.returncode
-            )
+        logger.debug(
+            "phase '%s' failed with return code %d", phase["name"], process.returncode
         )
```
m-horky, line 289: Actually, looking at Core, return code 100 means "OK, do not continue". There should be a condition that separates 0 and 100 from other exit codes.
```py
if process.returncode not in (0, 100):
    logger.debug(...)
```

m-horky (changes requested):

src/insights_client/__init__.py
```diff
@@ -208,47 +230,62 @@ def run_phase(phase, client, validated_eggs):
...
-            continue
-        if config["gpg"] and not client.verify(egg)["gpg"]:
-            client_debug("WARNING: GPG verification failed. Not loading egg: %s" % egg)
-            continue
-        client_debug("phase '%s'; egg '%s'" % (phase["name"], egg))
+        logger.debug("phase '%s'; egg '%s'", phase["name"], egg)
 
         # prepare the environment
+        env = os.environ
+        logger.debug("Existing enviroment: %s", debug_environ(env))
```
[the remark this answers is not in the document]
Glutexo: It indeed is noisy. Yet it may not be straightforward to find out what the full environment is. A command line is the most straightforward, but still there are things like _sudo_ filtering, which variables can go throguh. Not to say when the Client is run by a manager like systemd.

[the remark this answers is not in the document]
Glutexo: Removed.

src/insights_client/__init__.py
Glutexo: Removed.

src/insights_client/__init__.py
Glutexo: It prints the path of the _insights\_client/\_\_init\_\_.py_, which can be useful whenever the _\_\_init\_\_.py_ file is not run directly. Most commonly this is from the _insights-client_ command script, which, too, can be run from _PATH_ instead of specifying the exact location. It can easily discern the RPM-installed client from the one in a development folder.

Glutexo: Removed.

src/insights_client/__init__.py
Glutexo: Interesting! I honestly found this log message one of the most useful: having the full command including the environment with PATH and PYTHONPATH at hand.

src/insights_client/__init__.py
Glutexo: Don’t you consider the actual return code a useful piece of information? With a condition like this we don’t know whether the phase exitted with 0 or 100. Actually, we can know that if we inspect the source code, because in case of 0, the “phase '%s' successful” is logged but for 100 it’s not. 🤷🏻‍♂️

Glutexo: Added the condition you suggested.

— A little side note: I used a list instead of a tuple. Although I know that tuples are more efficient and convey a meaning of an immutable object, I’d rather use them semantically as a real _tuple_ – a fixed set of coupled values, a structure if you will. In this case, we _list_ out the exit code for the condition.

Glutexo (commented):
Addressed the review comments. Force-pushed. Re-requesting a review.

src/insights_client/__init__.py
Glutexo: Removed.

src/insights_client/__init__.py
Glutexo: I still found it useful, because it’s logged right before entering the _get_phase_ function with its egg selection machinery. `phase '%s'; egg '%s'` is a new message saying that an egg has been selected for the phase.

src/insights_client/__init__.py
```diff
@@ -351,32 +384,58 @@ def _main():
...
     #  pass it to run_phase where we load it again
+    logger.debug("Verifying initial ENV_EGG: %s...", ENV_EGG)
     if gpg_validate(ENV_EGG):
+        logger.debug("GPG verification successful. Using initial egg: %s", ENV_EGG)
         valid_env_egg = [ENV_EGG]
     else:
+        logger.debug(
+            "WARNING: GPG verification failed. Not using initial egg: %s", ENV_EGG
+        )
         valid_env_egg = []
```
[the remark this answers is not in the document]
Glutexo: It sure is quite verbose. I replaced it with a single logging statement printing the paths of the verified eggs in order. That is what’s prepended to the _sys.path_, which is no longer logged, before importing.

Glutexo:
Force-pushed the last change and re-formatted with black. Stylish is no longer complaining.

Glutexo:
Fixed a few weak points, did slight refactoring to reduce logic repetition, and squashed into a fewer number of commits. Re-requesting a review.

src/insights_client/__init__.py
Glutexo: The introduced refactoring resolved this concern.

Glutexo:
Updated the PR description and added steps to verify.

src/insights_client/__init__.py
```diff
@@ -362,21 +413,27 @@ def _main():
...
+        sys.exit("No GPG-verified initial eggs can be found")
 
     # ENV egg comes first
-    sys.path = valid_env_egg + validated_eggs + sys.path
+    all_valid_eggs = valid_env_egg + validated_eggs
+    logger.debug("Using eggs: %s", join_path(all_valid_eggs))
+    sys.path = all_valid_eggs + sys.path
 
     try:
+        logger.debug("Importing the insights module from %s...", join_path(sys.path))
```
m-horky, line 424: Information on this line is completely covered by the `Using eggs:` log message:
```
2024-05-15 11:19:10,159    DEBUG __main__:420 Using eggs: /.../insights-core:/var/lib/insights/last_stable.egg
2024-05-15 11:19:10,160    DEBUG __main__:424 Importing the insights module from /.../insights-core:/var/lib/insights/last_stable.egg:/.../insights-client/src/insights_client:/.../insights-client/src:/usr/lib64/python312.zip:/usr/lib64/python3.12:/usr/lib64/python3.12/lib-dynload:/usr/lib64/python3.12/site-packages:/usr/lib/python3.12/site-packages...
```

src/insights_client/__init__.py
m-horky: This line is fully covered by previous log statement:
```
2024-05-15 11:19:10,279    DEBUG __main__:275 phase 'pre_update'; egg '/.../insights-core'
2024-05-15 11:19:10,279    DEBUG __main__:279 Adding to PYTHONPATH: /.../insights-core
```

m-horky (changes requested):
It's great to see this in combination with https://github.com/RedHatInsights/insights-core/pull/4094. This PR has come a long way, I only have two small requests, I'll be happy to merge this after they are resolved.

src/insights_client/__init__.py
Glutexo: And partially with the “Loaded initial egg” message directly after. We agreed that we don’t need one message before an action and another one after it.

src/insights_client/__init__.py
Glutexo: > phase '%s'; egg '%s'

does not describe the implementation detail of adding to the PYTHONPATH, so I wouldn’t say it’s fully coveret. The egg path is present though, that’s true.

Glutexo: Removed.

src/insights_client/__init__.py
Glutexo: Removed.

Glutexo:
Addressed the review comments and rebased on the current master. Re-requesting a review.

m-horky (approved):
ACK

PR #196, opened 2024-03-13
Check the sources with flake8

ptoscano:
- remove extra/unused arguments in `format()` (fixes F522)
- remove an unused `import` (fixes F401)
- fix/improve the existing `.flake8` file according to the current sources
- run `flake8` in the existing `stylish` GHA workflow

See the messages of the commits for longer descriptions.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,3 +1,19 @@
 [flake8]
-ignore = E501,E126,E127,E128,E722
-exclude = bin,docs,include,lib,lib64,.git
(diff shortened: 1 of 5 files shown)


m-horky (approved):
LGTM

> Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: rbialon/flake8-annotations@v1.

The latest version is 1.1, I guess we can't do anything there.

PR #198, opened 2024-03-14
fix(test): use install_katello_rpm before subman_session

ptoscano:
This makes sure that the "install_katello_rpm" fixture is setup before "subman_session", and "install_katello_rpm" is tore down after "subman_session". This should ensure that the katello-ca-consumer RPM is uninstalled after "subman_session" unregisters the system.

### DIFF
--- a/integration-tests/conftest.py
+++ b/integration-tests/conftest.py
@@ -21,7 +21,7 @@ def install_katello_rpm(test_config):
 
 @pytest.fixture(scope="session")
 def register_subman(
-    external_candlepin, subman_session, test_config, install_katello_rpm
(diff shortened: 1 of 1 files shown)


zhangqianqian:
Verified that this PR makes the "install_katello_rpm" fixture is setup before "subman_session", and "install_katello_rpm" is tore down after "subman_session":


image

zhangqianqian:
> Verified that this PR makes the "install_katello_rpm" fixture is setup before "subman_session", and "install_katello_rpm" is tore down after "subman_session":

I think this PR can be merged.

m-horky (approved):
ACK

PR #199, opened 2024-03-14
Tests: Added Test to verify insights-client file permission and logro…

Archana-PandeyM:


integration-tests/test_client.py
```diff
@@ -0,0 +1,99 @@
...
+
+logger = logging.getLogger(__name__)
+
+pytestmark = pytest.mark.usefixtures("register_subman")
+
+
+def test_client_files_permission(insights_client):
+    """Verify that permission for following files is 0644:
+    - /etc/insights-client/.lastupload
+    - /etc/insights-client/.last-upload.results only created in legacy mode which
```
ptoscano, line 17: since we are not dealing with legacy upload, should this comment be here?

integration-tests/test_client.py
```diff
@@ -0,0 +1,99 @@
...
+    Verify the existence of mandatory files for insights-client rpm
+    """
+    if (
+        pytest.rhel_major_version in ("6", "7") or pytest.centos_major_version == "7"
+    ) and "motd" in file:
+        pytest.skip("motd feature not enabled in RHEL6/7 and centos7 RPMs")
+
+    elif pytest.rhel_major_version in ("6") and "systemd" in file:
+        pytest.skip("No systemd files in RHEL6 RPMs")
+
```
ptoscano, line 59: considering that
- these tests represent/test insights-client in the branch in which they are
- `master` is for EL 8 & 9

should we really include tests for older/other versions?

integration-tests/test_client.py
```diff
@@ -0,0 +1,99 @@
...
+def test_client_files_permission(insights_client):
+    """Verify that permission for following files is 0644:
+    - /etc/insights-client/.lastupload
+    - /etc/insights-client/.last-upload.results only created in legacy mode which
+    will be deprecated soon, hence skipping
+
+    Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1924990
+    """
+    file_ = "/etc/insights-client/.lastupload"
```
ptoscano, line 22: rather than `file_`, i'd give it a better name e.g. `last_upload_file`

ptoscano (commented):
some initial comments; not an exhaustive review, will continue tomorrow

integration-tests/test_client.py
```diff
@@ -0,0 +1,99 @@
...
+    if exists(file_):
+        os.remove(file_)  # performing a cleanup before test
```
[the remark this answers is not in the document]
jstavel: another pythonic hint to use:

> os.path.exists(file_) and os.remove(file_)


happy pythoning!

integration-tests/test_client.py
jstavel: I vote for the same - ie. it is better to recognize the main meaning of it.

integration-tests/test_client.py
```diff
@@ -0,0 +1,99 @@
...
+    ) and "motd" in file:
+        pytest.skip("motd feature not enabled in RHEL6/7 and centos7 RPMs")
+
+    elif pytest.rhel_major_version in ("6") and "systemd" in file:
+        pytest.skip("No systemd files in RHEL6 RPMs")
+
+    else:
+        cmd = ["rpm", "-ql", "insights-client"]
+        rpm_ql_output = subprocess.run(cmd, capture_output=True, text=True)
+        assert file in rpm_ql_output.stdout
```
ptoscano, line 63: while this works, it is a bit inefficient, as it will run `rpm -ql` for every combination of this parametrized test case

what I'd do instead is to save the split lines of `rpm -ql` as fixture for the test, so it is run only once

integration-tests/test_client.py
```diff
@@ -0,0 +1,99 @@
...
+def test_verify_logrotate_is_configured(insights_client):
+    """Verify that standard logrotate is configured via /etc/logrotate.d/insights-client
+    Ref : https://bugzilla.redhat.com/show_bug.cgi?id=1940267
+    """
+    logrotate_conf_file_path = "/etc/logrotate.d/insights-client"
+
+    assert os.path.exists(logrotate_conf_file_path), "logrotate is not configured"
+    f = open("/etc/logrotate.d/insights-client")
+    content = f.read()
+    logger.info(f"Content of log file is {content}")
```
ptoscano, line 99: this test in practice only checks that `/etc/logrotate.d/insights-client` exists; if there is nothing more to do for it, it seems to me it'd be better to add that path to the parameters of `test_client_rpm_mandatory_files` (`/etc/logrotate.d/insights-client` is shipped as RPM file)

integration-tests/test_client.py
Archana-PandeyM: I will remove these, I had different understanding that these tests are for all distros.

integration-tests/test_client.py
Archana-PandeyM: ok.

integration-tests/test_client.py
ptoscano: hm displaying the content of the file during the test is not exactly useful, as it will be hidden inside the test logs; in the end, the content of the file is what is shipped in the package, so it should be easy to fetch its content whenever needed

my vote here would be:
- add the existence of `/etc/logrotate.d/insights-client` to `test_client_rpm_mandatory_files`
- drop this test for now
- in case there are more in-depth checks for the logrotate enablement/configuration, add a brand-new test

integration-tests/test_client.py
```diff
@@ -0,0 +1,99 @@
...
+
+    Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1924990
+    """
+    file_ = "/etc/insights-client/.lastupload"
+    if exists(file_):
+        os.remove(file_)  # performing a cleanup before test
+
+    insights_client.register()
+    assert insights_client.is_registered
+    assert oct(os.stat(file_).st_mode)[-3:] == "644"
```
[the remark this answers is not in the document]
Archana-PandeyM: I think there is problem with suggestion, left side int right side is oct, how will comparison work

integration-tests/test_client.py
Archana-PandeyM: I am removing this test, will cover logging the content in new test that I am working on, validating logrotate feature.

integration-tests/test_client.py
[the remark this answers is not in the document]
ptoscano: oh right, you need to filter the actual permissions:
```python
    assert (os.stat(file_).st_mode & 0o777) == 0o644
```

integration-tests/test_client.py
```diff
@@ -0,0 +1,84 @@
...
+        filename in rpm_ql_insights_client
+    ), f"{filename} is not in insights-client package"
+
+
+def test_client_logfiles_mask(insights_client):
+    """Verify that files in /var/log/insights-client have the right mode: 0600
+
+    Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1955724
+    """
+    logfiles = glob.glob("/var/log/insights-client/*.log*")
```
ptoscano, line 65: the `insights_client` fixture ensures that, before each test, there are no log files around: because of this, `logfiles` will most likely be empty right at the beginning of the test; my suggestion is to move it below the `register()` call, so at least something will be logged, and thus files will be around

integration-tests/test_client.py
```diff
@@ -0,0 +1,84 @@
...
+
+def test_client_logdir_permissions():
+    """Verify that permissions on directory /var/log/insights-client
+    have the right mode: 0700
+    """
+    logdir_name = "/var/log/insights-client"
+    permissions = os.stat(logdir_name).st_mode
+    assert (
+        permissions == 0o40700
+    ), f"permissions on dir {logdir_name} must be: 0700, current one is: {permissions}"
```
ptoscano, line 84: this label for `assert` is very much redundant: on case of failure, pytest will print the "expected" and "got" values already; also, this test checks exactly one thing, so it is clear that, if it fails, that specific directory has wrong permissions

integration-tests/test_client.py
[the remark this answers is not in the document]
Archana-PandeyM: this when asserts shows integer values in assertion message , which makes the log difficult to interpret.
    _assert (os.stat(file_).st_mode & 0o777) == 0o644
E   AssertionError: assert (33152 & 511) == 420_
However if I use 
`assert oct(os.stat(file_).st_mode & 0o777) == '0o644'` its more readable
    assert oct(os.stat(file_).st_mode & 0o777) == '0o601'
E   AssertionError: assert '0o600' == '0o601'

integration-tests/test_client.py
```diff
@@ -0,0 +1,103 @@
...
+        1-Count the number of log file in beginning
+        2-Rotate logs by running logrotate command on CLI for insights-client
+        3-Verify that log is rotated and new log file is created
+    """
+
+    logrotate_conf_file_path = "/etc/logrotate.d/insights-client"
+    logdir = "/var/log/insights-client/"
+    assert os.path.exists(logrotate_conf_file_path), "logrotate is not configured"
+    logrotate_file = open(logrotate_conf_file_path)
+    logger.info(f"Content of log file is {logrotate_file.read()}")
```
ptoscano, line 95: not sure what this is for - i'd simply not print to the debug of the test the content of a file...

integration-tests/test_client.py
```diff
@@ -0,0 +1,103 @@
...
+
+    # It is necessary to perform some command using insights-client
+    # to populate logs
+    insights_client.register()
+    number_of_log_files = len(os.listdir(logdir))
+    subprocess.run(["logrotate", "-vf", logrotate_conf_file_path], check=True)
```
ptoscano, line 101: hint: `run(... check=True)` -> `check_call()`

ptoscano (changes requested):
thanks for the updates! this generally looks good to me.

One important thing: the new `test_verify_logrotate_feature` test uses logrotate, which thus must be installed (in my test systems I did not have it); please add `logrotate` as installed package in `systemtest/tests/integration/test.sh` (there is already a `dnf install` command there).

------

Some of the tests in the newly added file can also be run without registering; one possibility here could be to apply the `register_subman` fixture only to the tests that need to register, rather than the whole file -- for example:
```python
def test_client_files_permission(insights_client):
```
This way, at least few more tests would run upstream.

WDYT?

------

One minor style note: I see the RHBZ references as:
```
    Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1955724
```
```
    Ref : https://bugzilla.redhat.com/show_bug.cgi?id=1940267
```
maybe worth using one style? what about a simpler
```
    RHBZ: https://bugzilla.redhat.com/show_bug.cgi?id=1940267
```
?

Archana-PandeyM:
> WDYT
Makes sense. Thank you!

Archana-PandeyM:
> thanks for the updates! this generally looks good to me.
> 
> One important thing: the new `test_verify_logrotate_feature` test uses logrotate, which thus must be installed (in my test systems I did not have it); please add `logrotate` as installed package in `systemtest/tests/integration/test.sh` (there is already a `dnf install` command there).
> 
> Some of the tests in the newly added file can also be run without registering; one possibility here could be to apply the `register_subman` fixture only to the tests that need to register, rather than the whole file -- for example:
> 
> ```python
> @pytest.mark.usefixtures("register_subman")
> def test_client_files_permission(insights_client):
> ```
> 
> This way, at least few more tests would run upstream.
> 
> WDYT?
> 
> One minor style note: I see the RHBZ references as:
> 
> ```
>     Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1955724
> ```
> 
> ```
>     Ref : https://bugzilla.redhat.com/show_bug.cgi?id=1940267
> ```
> 
> maybe worth using one style? what about a simpler
> 
> ```
>     RHBZ: https://bugzilla.redhat.com/show_bug.cgi?id=1940267
> ```
> 
> ?

I think 'Ref' could be used because it need not always be Bugzilla going forward. It could be Jira also.
I will update code to be consistent.

integration-tests/test_client.py
```diff
@@ -0,0 +1,99 @@
...
+
+    logrotate_conf_file_path = "/etc/logrotate.d/insights-client"
+    logdir = "/var/log/insights-client/"
+    assert os.path.exists(logrotate_conf_file_path), "logrotate is not configured"
+    # It is necessary to perform some command using insights-client
+    # to populate logs
+    insights_client.register()
+    number_of_log_files = len(os.listdir(logdir))
+    cmd = "logrotate -vf " + logrotate_conf_file_path
+    subprocess.check_call(cmd, shell=True)
```
ptoscano, line 97: the command split as list without `shell=True` was correct :sweat_smile:

ptoscano (approved):
thank you, LGTM now

will merge once the last CI jobs complete successfully

PR #201, opened 2024-03-18
fix(test): Loop for statement

jstavel:
It should help in a situation where it is necessary to sleep for a while for a long running service to update its state.
I've applied the function in a test to present it.

### DIFF
--- a/integration-tests/conftest.py
+++ b/integration-tests/conftest.py
@@ -1,5 +1,6 @@
 import pytest
 import subprocess
+import time
 
 
 @pytest.fixture(scope="session")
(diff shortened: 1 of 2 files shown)


jstavel:
I have fixed a problem with verification of the second deregistering message.
I have run all tests in the file test_deregister.py - all are OK now.
is this the one you @Lorquas has requested?

integration-tests/conftest.py
```diff
@@ -34,3 +35,26 @@ def register_subman(
...
+to update its state.
+
+an example:
+
+   result = loop_until(lambda: insights_client.is_registered)
+   assert result == True
+
+   The loop function will retry to run predicate every 5secs
+   untill the total time exceeds timeout_sec.
+"""
```
Lorquas, line 51: This docstring should be moved inside the function under the `def` line, that way `help(conftest.loop_until)` will work.  See: https://peps.python.org/pep-0257/

Lorquas:
Overall good. Can you also add this to the double registration test (that's also broken with the same bug)?  Also you need to squash these commits.

Lorquas (changes requested):
Overall good. Can you also add this to the double registration test (that's also broken with the same bug)? Also you need to squash these commits.

PR #202, opened 2024-03-19
feat(test): Add tests for MOTD

m-horky:
* Card ID: CCT-414
* Card ID: CCT-415

### DIFF
--- a/integration-tests/test_motd.py
+++ b/integration-tests/test_motd.py
@@ -1,0 +1,72 @@
+import contextlib
+import os
+
+import pytest
+
+
+DOT_REGISTERED_PATH = "/etc/insights-client/.registered"
+DOT_UNREGISTERED_PATH = "/etc/insights-client/.unregistered"
+
+MOTD_PATH = "/etc/motd.d/insights-client"
+MOTD_SRC = "/etc/insights-client/insights-client.motd"
+
+
+pytestmark = pytest.mark.usefixtures("register_subman")
+
+
+@pytest.fixture
+def delete_special_files():
+    """Clean special files before and after test.
+
+    Since .registered and .unregistered files are usually ignored and
+    do not affect the behavior of insights-client, the MOTD case is
+    special and depends on them.
+
+    To ensure the tests here are not affected by them, we remove them
+    """
+
+    def delete_files():
+        with contextlib.suppress(FileNotFoundError):
+            os.remove(DOT_REGISTERED_PATH)
+        with contextlib.suppress(FileNotFoundError):
+            os.remove(DOT_UNREGISTERED_PATH)
+        with contextlib.suppress(FileNotFoundError):
+            os.remove(MOTD_PATH)
+
+    delete_files()
+    yield
+    delete_files()
+
+
+@pytest.mark.usefixtures("delete_special_files")
+def test_motd(insights_client):
+    """MOTD only exists on unregistered system without (.un)registered files."""
+    # If the system is not registered, the file should be present.
+    insights_client.run("--status")
+    assert os.path.exists(MOTD_PATH)
+
+    # After registration, the file should not exist.
+    insights_client.register()
+    assert not os.path.exists(MOTD_PATH)
+
+    # The system was unregistered. Because .unregistered file exists,
+    # the file should not be present.
+    insights_client.unregister()
+    assert not os.path.exists(MOTD_PATH)
+
+
+@pytest.mark.usefixtures("delete_special_files")
+def test_motd_dev_null(insights_client):
+    """MOTD should not be touched if it is a /dev/null symlink."""
+    with contextlib.ExitStack() as stack:
+        os.symlink(os.devnull, MOTD_PATH)
+        stack.callback(os.unlink, MOTD_PATH)
+
+        insights_client.run("--status")
+        assert os.path.samefile(os.devnull, MOTD_PATH)
+
+        insights_client.register()
+        assert os.path.samefile(os.devnull, MOTD_PATH)
+
+        insights_client.unregister()
+        assert os.path.samefile(os.devnull, MOTD_PATH)
--- a/src/insights_client/tests/test_motd.py
+++ b/src/insights_client/tests/test_motd.py
@@ -1,0 +1,113 @@
+import os
+import pathlib
+import tempfile
+import typing
+import unittest
+import unittest.mock
+
+import pytest
+
+import insights_client
+
+
+class MockFS:
+    """Mock filesystem.
+
+    Ensures that the files required for the MOTD manipulation are created
+    in a temporary directory instead of working with actual system files.
+    """
+
+    _dir: tempfile.TemporaryDirectory
+
+    @property
+    def _chroot(self) -> pathlib.Path:
+        return pathlib.Path(self._dir.name)
+
+    def touch(self, path: str):
+        (self._chroot / path).open("w").close()
+
+    def exists(self, path: str) -> bool:
+        return (self._chroot / path).exists()
+
+    def symlink_to(self, path: str, point_to: str):
+        (self._chroot / path).symlink_to(point_to)
+
+    def samefile(self, first: str, second: str) -> bool:
+        return (self._chroot / first).samefile(self._chroot / second)
+
+    def __init__(self):
+        self._dir = tempfile.TemporaryDirectory(prefix="test-chroot-")
+
+        paths: typing.Dict[str, str] = {
+            "insights_client.MOTD_SRC": "etc/insights-client/insights-client.motd",
+            "insights_client.MOTD_FILE": "etc/motd.d/insights-client",
+            "insights_client.REGISTERED_FILE": "etc/insights-client/.registered",
+            "insights_client.UNREGISTERED_FILE": "etc/insights-client/.unregistered",
+        }
+
+        self.patches: typing.List[unittest.mock.patch] = []
+        for target, path in paths.items():
+            mocked_path = self._chroot / path
+            mocked_path.parent.mkdir(parents=True, exist_ok=True)
+            patch = unittest.mock.patch(target, f"{mocked_path!s}")
+            self.patches.append(patch)
+            patch.start()
+
+        self.touch("etc/insights-client/insights-client.motd")
+
+    def __del__(self):
+        for patch in self.patches:
+            patch.stop()
+        self._dir.cleanup()
+
+
+@pytest.fixture(scope="function")
+def mock_fs():
+    fs = MockFS()
+    yield fs
+    del fs
+
+
+def test_present(mock_fs):
+    assert not mock_fs.exists("etc/motd.d/insights-client")
+
+    # The file gets created/symlinked when .registered & .unregistered do not exist
+    insights_client.update_motd_message()
+    assert mock_fs.exists("etc/motd.d/insights-client")
+    assert mock_fs.samefile(
+        "etc/motd.d/insights-client", "etc/insights-client/insights-client.motd"
+    )
+
+
+@pytest.mark.parametrize(
+    "filename",
+    [".registered", ".unregistered"],
+)
+def test_absent_on_dot(mock_fs, filename):
+    # The MOTD gets created by default...
+    insights_client.update_motd_message()
+    assert mock_fs.exists("etc/motd.d/insights-client")
+
+    # ...and gets removed when .registered/.unregistered exists.
+    mock_fs.touch(f"etc/insights-client/{filename}")
+
+    insights_client.update_motd_message()
+    assert not mock_fs.exists("etc/motd.d/insights-client")
+
+    # ...and it stays absent when run multiple times.
+    insights_client.update_motd_message()
+    assert not mock_fs.exists("etc/motd.d/insights-client")
+
+
+def test_ignored_on_dev_null(mock_fs):
+    # When the /etc/motd.d/insights-client is a symbolic link to /dev/null...
+    mock_fs.symlink_to("etc/motd.d/insights-client", os.devnull)
+
+    # ...it should not be overwritten...
+    insights_client.update_motd_message()
+    assert mock_fs.samefile("etc/motd.d/insights-client", os.devnull)
+
+    # ...whether the MOTD file should be present or not.
+    mock_fs.touch("etc/insights-client/.registered")
+    insights_client.update_motd_message()
+    assert mock_fs.samefile("etc/motd.d/insights-client", os.devnull)


integration-tests/test_motd.py
```diff
@@ -0,0 +1,52 @@
...
+
+@pytest.fixture
+def clean_system():
+    """Clean up the system before every test."""
+    with contextlib.suppress(OSError):
+        os.remove(DOT_REGISTERED_PATH)
+    with contextlib.suppress(OSError):
+        os.remove(DOT_UNREGISTERED_PATH)
+    with contextlib.suppress(OSError):
+        os.remove(MOTD_PATH)
```
ptoscano, line 25: maybe use `FileNotFoundError` instead of a generic `OSError`; if there are other issues than "the file does not exist", I'd like to see them rather than silently ignoring them

integration-tests/test_motd.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+def test_motd(clean_system, insights_client):
```
ptoscano, line 28: since `clean_system` is merely to do some pre-test work, maybe rather include it for the test without using it as parameter:
```python
@pytest.mark.usefixtures("clean_system")
def test_motd(insights_client):
```

integration-tests/test_motd.py
```diff
@@ -0,0 +1,52 @@
...
+    insights_client.register()
+    assert not os.path.exists(MOTD_PATH)
+
+    # The system was unregistered. Because .unregistered file exists,
+    # the file should not be present.
+    insights_client.unregister()
+    assert not os.path.exists(MOTD_PATH)
+
+
+def test_motd_dev_null(clean_system, insights_client):
```
ptoscano, line 44: ditto (`@pytest.mark.usefixtures`)

integration-tests/test_motd.py
```diff
@@ -0,0 +1,52 @@
...
+    """MOTD should not be touched if it is a /dev/null symlink."""
+    os.symlink(os.devnull, MOTD_PATH)
```
ptoscano, line 46: this symlink ought to be removed at the end; I'd recommend creating a `contextlib.ExitStack` variable, and pushing an helper function to `os.unlink()` the file

integration-tests/test_motd.py
```diff
@@ -0,0 +1,52 @@
...
+MOTD_PATH = "/etc/motd.d/insights-client"
+MOTD_SRC = "/etc/insights-client/insights-client.motd"
+
+
+pytestmark = pytest.mark.usefixtures("register_subman")
+
+
+@pytest.fixture
+def clean_system():
+    """Clean up the system before every test."""
```
ptoscano, line 19: maybe give this fixture a better name & comment on what it does, as IMHO `clean_system` a bit too generic

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,97 @@
...
+import unittest.mock
+
+import insights_client
+
+
+class TestMOTD(unittest.TestCase):
+    """Verify the behavior of MOTD file."""
+
+    def setUp(self):
+        self.dir = tempfile.TemporaryDirectory()
```
ptoscano, line 14: the other unit tests use the pytest API rather than `unittest.TestCase`, so it'd be better for the test cases here to follow the same approach

in that case, the `tmp_path` fixture will give you a new temporary directory for every test in which it is used

ptoscano (changes requested):

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,97 @@
...
+        assert os.path.samefile(insights_client.MOTD_FILE, insights_client.MOTD_SRC)
+
+    def test_absent_on_dot_registered(self):
+        # The MOTD file exists by default
+        insights_client.update_motd_message()
+        assert os.path.exists(insights_client.MOTD_FILE)
+
+        # It gets removed when .registered or .unregistered exists
+        with open(f"{self.dir.name}/reg", "w") as f:
+            f.write("")
```
ptoscano, line 62: if all you want to do here is create an empty file, then you can do it in a simpler way (IMHO):
```suggestion
        os.truncate(f"{self.dir.name}/reg", 0)
```
(the same applies to the other occurrences of this code)

m-horky: `truncate` operates on existing files, this raises `FileNotFoundError`s.

src/insights_client/tests/test_motd.py
m-horky: ACK, done with fixtures instead. Whether it is more or less readable I cannot say.

m-horky:
Rawhide test failure is addressed by #245

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,116 @@
...
+    Ensures that the files required for the MOTD manipulation are created
+    in a temporary directory instead of working with actual system files.
+    """
+
+    def __init__(self):
+        self.dir = tempfile.TemporaryDirectory()
+
+        # /etc/insights-client/insights-client.motd
+        self.dir_src = unittest.mock.patch(
+            "insights_client.MOTD_SRC", f"{self.dir.name}/src"
```
Glutexo, line 23: For safe path joining, use os.path.join instead of string interpolation.

```suggestion
            "insights_client.MOTD_SRC", os.path.join(f"{self.dir.name}/src"
```

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,116 @@
...
+        )
+        self.dir_reg.start()
+
+        # /etc/insights-client/.unregistered
+        self.dir_unreg = unittest.mock.patch(
+            "insights_client.UNREGISTERED_FILE", f"{self.dir.name}/unreg"
+        )
+        self.dir_unreg.start()
+
+        with open(f"{self.dir.name}/src", "w") as f:
```
Glutexo, line 45:
```suggestion
        with open(os.path.join(self.dir.name, "src"), "w") as f:
```

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,116 @@
...
+
+        # /etc/insights-client/insights-client.motd
+        self.dir_src = unittest.mock.patch(
+            "insights_client.MOTD_SRC", f"{self.dir.name}/src"
+        )
+        self.dir_src.start()
+
+        # /etc/motd.d/insights-client
+        self.dir_motd = unittest.mock.patch(
+            "insights_client.MOTD_FILE", f"{self.dir.name}/motd"
```
Glutexo, line 29:
```suggestion
            "insights_client.MOTD_FILE", os.path.join(self.dir.name, "motd")
```

```suggestion
            "insights_client.MOTD_FILE", f"{self.dir.name}{insights_client.MOTD_FILE}"
```

```suggestion
            "insights_client.MOTD_FILE", f"{self.dir.name}/etc--motd-d--insights-client"
```

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,116 @@
...
+        )
+        self.dir_motd.start()
+
+        # /etc/insights-client/.registered
+        self.dir_reg = unittest.mock.patch(
+            "insights_client.REGISTERED_FILE", f"{self.dir.name}/reg"
```
Glutexo, line 35:
```suggestion
            "insights_client.REGISTERED_FILE", os.path.join(self.dir.name, "reg")
```

```suggestion
            "insights_client.REGISTERED_FILE", f"{self.dir.name}{insights_client.REGISTERED_FILE}"
```

```suggestion
            "insights_client.REGISTERED_FILE", f"{self.dir.name}/etc--insights-client--registered"
```

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,116 @@
...
+        )
+        self.dir_reg.start()
+
+        # /etc/insights-client/.unregistered
+        self.dir_unreg = unittest.mock.patch(
+            "insights_client.UNREGISTERED_FILE", f"{self.dir.name}/unreg"
```
Glutexo, line 41:
```suggestion
            "insights_client.UNREGISTERED_FILE", os.path.join(self.dir.name, "unreg")
```

```suggestion
            "insights_client.UNREGISTERED_FILE", f"{self.dir.name}{insights_client.UNREGISTERED_FILE}"
```

```suggestion
            "insights_client.UNREGISTERED_FILE", f"{self.dir.name}/etc--insights-client--unregistered"
```

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,116 @@
...
+        )
+        self.dir_unreg.start()
+
+        with open(f"{self.dir.name}/src", "w") as f:
+            f.write("This is the content of the MOTD file.")
+
+    def __del__(self):
```
Glutexo, line 48: This is, however, nice: leveraging the descruction method. Gives me a Ruby-like feeling. ❤️

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,116 @@
...
+        self.dir_motd.stop()
+        self.dir_src.stop()
+        self.dir = None
+
+
+@pytest.fixture
+def mock_fs():
+    fs = MockFS()
+    yield fs
+    del fs
```
Glutexo, line 60: ❤️

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,116 @@
...
+        self.dir_reg = unittest.mock.patch(
+            "insights_client.REGISTERED_FILE", f"{self.dir.name}/reg"
+        )
+        self.dir_reg.start()
+
+        # /etc/insights-client/.unregistered
+        self.dir_unreg = unittest.mock.patch(
+            "insights_client.UNREGISTERED_FILE", f"{self.dir.name}/unreg"
+        )
+        self.dir_unreg.start()
```
Glutexo, line 43: There is a lot of repetition, which can be replaces by a for-cycle.

```suggestion
        DIRS = [
            ("insights_client.MOTD_SRC", "/src"),  # /etc/insights-client/insights-client.motd
            ("insights_client.MOTD_FILE", "/motd"),  # /etc/motd.d/insights-client
            ("insights_client.REGISTERED_FILE", "/reg"),  # /etc/insights-client/.registered
            ("insights_client.UNREGISTERED_FILE", "/unreg"),  # /etc/insights-client/.unregistered
        ]
        
        self.dirs = {}
        for target, value in DIRS:
            self.dirs[target] = unittest.mock.patch(target, f"{self.dir.name}{value}"
            self.dirs[target].start()
```

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,116 @@
...
+
+        with open(f"{self.dir.name}/src", "w") as f:
+            f.write("This is the content of the MOTD file.")
+
+    def __del__(self):
+        self.dir_unreg.stop()
+        self.dir_reg.stop()
+        self.dir_motd.stop()
+        self.dir_src.stop()
```
Glutexo, line 52:
```suggestion
        for dir in self.dirs:
            dir.stop()
```

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,116 @@
...
+        self.dir = None
```
Glutexo, line 53: Is this necessary? It doesn’t do anything and doesn’t even return the object to the pre-initialization state.

```suggestion
        del self.dir
```

may trigger at least some object destruction.

```suggestion
```
is however completely sufficent. The object is depleted and can’t be re-used, so there is no need to put it to a “clean” state.

src/insights_client/tests/test_motd.py
Glutexo, line 60: I can imagine doing this without the class. But as mentioned above, I like how it encapsulates the state and leverages the object deletion.

```suggestion
    fs = setup_fs()
    yield fs
    teardown_fs(fs)
```

Another option would be to put the function contents directly in here. With the suggested for-cycle, it’s not even that complicated piece of code to be inlined.

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,116 @@
...
+    in a temporary directory instead of working with actual system files.
+    """
+
+    def __init__(self):
+        self.dir = tempfile.TemporaryDirectory()
+
+        # /etc/insights-client/insights-client.motd
+        self.dir_src = unittest.mock.patch(
+            "insights_client.MOTD_SRC", f"{self.dir.name}/src"
+        )
```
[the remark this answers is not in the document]
m-horky: "chroot" wasn't a bad idea, it turned out quite elegant.

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,114 @@
...
+        with motd_src.open("w") as f:
+            f.write("This is the content of the MOTD file.")
+
+    @property
+    def chroot(self) -> str:
+        return self._dir.name
+
+    def __del__(self):
+        for mock in self.mocks.values():
```
Glutexo, line 49: This is the only place where _self.mocks_ is used and its keys are not used at all. If this was a list, the iteration wouldn’t need the `.values` call.

```suggestion
        for mock in self.mocks:
```

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,114 @@
...
+        paths: dict[str, str] = {
+            "insights_client.MOTD_SRC": "/etc/insights-client/insights-client.motd",
+            "insights_client.MOTD_FILE": "/etc/motd.d/insights-client",
+            "insights_client.REGISTERED_FILE": "/etc/insights-client/.registered",
+            "insights_client.UNREGISTERED_FILE": "/etc/insights-client/.unregistered",
+        }
+
+        self.mocks: dict[str, unittest.mock.patch] = {}
+        for label, path in paths.items():
+            mocked_path = f"{self.chroot}{path}"
```
Glutexo, line 33: Isn’t this a place for os.path.join rather than a string interpolation?

```suggestion
            mocked_path = os.path.join(self.chroot, path)
```

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,114 @@
...
+
+        paths: dict[str, str] = {
+            "insights_client.MOTD_SRC": "/etc/insights-client/insights-client.motd",
+            "insights_client.MOTD_FILE": "/etc/motd.d/insights-client",
+            "insights_client.REGISTERED_FILE": "/etc/insights-client/.registered",
+            "insights_client.UNREGISTERED_FILE": "/etc/insights-client/.unregistered",
+        }
+
+        self.mocks: dict[str, unittest.mock.patch] = {}
+        for label, path in paths.items():
```
Glutexo, line 32: This would work even if _paths_ was a list of tuples rather than a dictionary.

Glutexo, line 32: I was a bit confused by the variable name _label_. It gave me an impression that it’s some arbitrary naming string, but it fact, it’s the constant path and name to be patched. [unittest.mock](https://docs.python.org/3.6/library/unittest.mock.html) [calls](https://docs.python.org/3.6/library/unittest.mock.html#unittest.mock.patch) it “target”.

```suggestion
        for target, path in paths.items():
```

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,114 @@
...
+            mocked_path = f"{self.chroot}{path}"
+            pathlib.Path(mocked_path).parent.mkdir(parents=True, exist_ok=True)
+            self.mocks[label] = unittest.mock.patch(label, mocked_path)
```
Glutexo, line 35: When called “target” instead of “label”, it conveys the idea clearer.

```suggestion
            self.mocks[target] = unittest.mock.patch(target, mocked_path)
```

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,114 @@
...
+            self.mocks[label].start()
```
Glutexo, line 36: If self.mocks was a list, it would look like this:

```suggestion
            mock = unittest.mock.patch(label, mocked_path)
            mock.start()
            self.mocks.append(mock)
```

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,114 @@
...
+            "insights_client.MOTD_SRC": "/etc/insights-client/insights-client.motd",
+            "insights_client.MOTD_FILE": "/etc/motd.d/insights-client",
+            "insights_client.REGISTERED_FILE": "/etc/insights-client/.registered",
+            "insights_client.UNREGISTERED_FILE": "/etc/insights-client/.unregistered",
+        }
+
+        self.mocks: dict[str, unittest.mock.patch] = {}
+        for label, path in paths.items():
+            mocked_path = f"{self.chroot}{path}"
+            pathlib.Path(mocked_path).parent.mkdir(parents=True, exist_ok=True)
```
Glutexo, line 34: Although I like the object-oriented [pathlib](https://docs.python.org/3.6/library/pathlib.html?highlight=pathlib#module-pathlib) module, calls like these can easily become quite complex. Importing Path directly would help, but that may not be of your taste and may hinder consistency.

```suggestion
            from pathlib import Path
            Path(mocked_path).parent.mkdir(parents=True, exist_ok=True)
```

Using [os.path.dirname](https://docs.python.org/3.6/library/os.path.html#os.path.dirname) and [os.path.makedirs] may make the call even more concise:

```suggestion
            os.path.makedirs(os.path.dirname(mocked_path), exist_ok=True)
```

or even

```suggestion
            from os.path import dirname, makedirs
            makedirs(dirname(mocked_path), exist_ok=True)
```

if combined. To me it looks elegant.

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,114 @@
...
+    """Mock filesystem.
+
+    Ensures that the files required for the MOTD manipulation are created
+    in a temporary directory instead of working with actual system files.
+    """
+
+    _dir: tempfile.TemporaryDirectory
+
+    def __init__(self):
+        self._dir = tempfile.TemporaryDirectory(prefix="test-chroot-")
```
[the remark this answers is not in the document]
m-horky: Yes. When it's left laying around, or when you want to inspect the test, it's easier to find and identify.

src/insights_client/tests/test_motd.py
m-horky: I don't think it matters. It's a mapping from identifier to path, I think it is fine.

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,114 @@
...
+
+        paths: dict[str, str] = {
+            "insights_client.MOTD_SRC": "/etc/insights-client/insights-client.motd",
+            "insights_client.MOTD_FILE": "/etc/motd.d/insights-client",
+            "insights_client.REGISTERED_FILE": "/etc/insights-client/.registered",
+            "insights_client.UNREGISTERED_FILE": "/etc/insights-client/.unregistered",
+        }
+
+        self.mocks: dict[str, unittest.mock.patch] = {}
```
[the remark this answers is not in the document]
m-horky: Good point. I think the list is a bit easier to understand and track on first sight.

src/insights_client/tests/test_motd.py
m-horky: I like the object approach, I find it more descriptive than what you are proposing. Plus I dislike the `((value), arg=value)` it causes, it you have to keep more context in mind because one function is returning and other is ingesting. I know that is how some languages work, but in Python it just confuses me.

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,114 @@
...
+            self.mocks[label].start()
+
+        motd_src = pathlib.Path(
+            f"{self.chroot}/etc/insights-client/insights-client.motd"
+        )
+        with motd_src.open("w") as f:
+            f.write("This is the content of the MOTD file.")
+
+    @property
+    def chroot(self) -> str:
```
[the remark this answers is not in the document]
m-horky: Yeah. A small abstraction.

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,116 @@
...
+
+    # It gets removed when .registered exists
+    with open(f"{mock_fs.chroot}/etc/insights-client/.registered", "w") as f:
+        f.write("")
+    insights_client.update_motd_message()
+    assert not os.path.exists(insights_client.MOTD_FILE)
+
+    # It stays absent when run multiple times
+    insights_client.update_motd_message()
+    assert not os.path.exists(insights_client.MOTD_FILE)
```
Glutexo, line 85: This is a separate test case, verifying that the MOTD file stays absent if already absent and the `.registered` file exists.

```suggestion


def test_still_absent_on_dot_registered(mock_fs):
    with open(f"{mock_fs.chroot}/etc/insights-client/.registered", "w") as f:
        f.write("")
    insights_client.update_motd_message()
    assert not os.path.exists(insights_client.MOTD_FILE)
```

If the test names were action based, it could be called something like “test_not_created_on_registered”.

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,116 @@
...
+def test_absent_on_dot_registered(mock_fs):
+    # The MOTD file exists by default
+    insights_client.update_motd_message()
+    assert os.path.exists(insights_client.MOTD_FILE)
+
+    # It gets removed when .registered exists
+    with open(f"{mock_fs.chroot}/etc/insights-client/.registered", "w") as f:
+        f.write("")
+    insights_client.update_motd_message()
+    assert not os.path.exists(insights_client.MOTD_FILE)
```
Glutexo, line 81:
```suggestion
    assert not path.exists()
```

src/insights_client/tests/test_motd.py
Glutexo, line 85:
```suggestion
    assert not path.exists()
```

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,116 @@
...
+    Ensures that the files required for the MOTD manipulation are created
+    in a temporary directory instead of working with actual system files.
+    """
+
+    _dir: tempfile.TemporaryDirectory
+
+    def __init__(self):
+        self._dir = tempfile.TemporaryDirectory(prefix="test-chroot-")
+
+        paths: typing.Dict[str, str] = {
```
Glutexo, line 25: It would make sense for this to be a dictionary, if it was a class attribute and it was then accessed to provide the actual paths based on the keys.

src/insights_client/tests/test_motd.py
Glutexo: It seems like abusing a dictionary when not used for searching.

src/insights_client/tests/test_motd.py
m-horky: Yeah, that is equivalent, but I don't see how it is better.

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,116 @@
...
+    def chroot(self) -> str:
+        return self._dir.name
+
+    def __del__(self):
+        for patch in self.patches:
+            patch.stop()
+        self._dir.cleanup()
+
+
+@pytest.fixture(scope="function")
```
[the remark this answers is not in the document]
m-horky: They don't _require_ it, but as you say, they need to be present as they are in actual filesystem at runtime so we can detect misbehavior.

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,116 @@
...
+def mock_fs():
+    fs = MockFS()
+    yield fs
+    del fs
+
+
+def test_present(mock_fs):
```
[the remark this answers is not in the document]
m-horky: I think it does. If it's really unclear, I'd rather add a docstring, but I didn't feel it was necessary.

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,116 @@
...
+    assert not os.path.exists(insights_client.MOTD_FILE)
```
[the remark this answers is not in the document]
m-horky: I think it hardly matters, this is not production code. I just used whatever felt right on that place.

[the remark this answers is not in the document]
m-horky: The setup could change in the future. This test checks for the presence of this file multiple times, this may help us find an issue if we misconfigure the setup by accident.

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,116 @@
...
+
+    # The file gets created/symlinked when .registered & .unregistered do not exist
```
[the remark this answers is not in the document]
m-horky: I disagree, comments are significantly easier to read than `long_underscored_method_names`.

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,116 @@
...
+    insights_client.update_motd_message()
+    assert os.path.exists(insights_client.MOTD_FILE)
+    assert os.path.samefile(insights_client.MOTD_FILE, insights_client.MOTD_SRC)
```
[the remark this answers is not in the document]
m-horky: What you are losing by this change is the direct knowledge of what `path` is. You have to look that up, instead of seeing it directly on that line.

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+def test_absent_on_dot_registered(mock_fs):
+    # The MOTD file exists by default
+    insights_client.update_motd_message()
+    assert os.path.exists(insights_client.MOTD_FILE)
```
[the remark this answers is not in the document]
m-horky: I would rather let the code create the file, so if something changes there, this test's setup continues working. I'd rather have two tests failing because of one change in a production code than a test working by accident because we haven't kept it updated.

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,116 @@
...
+
+    # It gets removed when .registered exists
+    with open(f"{mock_fs.chroot}/etc/insights-client/.registered", "w") as f:
```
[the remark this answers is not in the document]
m-horky: I've actually decided to flip it around and refer to chroot-based paths everywhere. It better describes what is happening on the mock filesystem without you having to remember what the variable refers to.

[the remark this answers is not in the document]
m-horky: It doesn't make it more readable or better imo, I'll stick with my version.

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,116 @@
...
+        f.write("")
```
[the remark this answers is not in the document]
m-horky: This never occurred to me, and I am not sure if I like it. Let's try.

src/insights_client/tests/test_motd.py
m-horky: I wouldn't say it's a separate case. I've updated the separation of blocks to make it more clear.

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,110 @@
...
+        self._dir = tempfile.TemporaryDirectory(prefix="test-chroot-")
+
+        paths: typing.Dict[str, str] = {
+            "insights_client.MOTD_SRC": "/etc/insights-client/insights-client.motd",
+            "insights_client.MOTD_FILE": "/etc/motd.d/insights-client",
+            "insights_client.REGISTERED_FILE": "/etc/insights-client/.registered",
+            "insights_client.UNREGISTERED_FILE": "/etc/insights-client/.unregistered",
+        }
+
+        self.patches: typing.List[unittest.mock.patch] = []
```
Glutexo, line 32: I know I suggested using a list rather than a dictionary. I however found a place later in the code, where haveing a dictionary would help.

To create an MOTD file, `self.patches["insights_client.MOTD_FILE"].new` would get the path in the temporary directory without repeating the chroot path joining logic.

src/insights_client/tests/test_motd.py
m-horky: That's never happening. We're fine.

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,116 @@
...
+
+def test_present(mock_fs):
+    assert not (mock_fs.chroot / "etc/motd.d/insights-client").exists()
+
+    # The file gets created/symlinked when .registered & .unregistered do not exist
+    insights_client.update_motd_message()
+    assert (mock_fs.chroot / "etc/insights-client/insights-client.motd").samefile(
+        mock_fs.chroot / "etc/motd.d/insights-client"
+    )
+    assert (mock_fs.chroot / "etc/motd.d/insights-client").exists()
```
Glutexo, line 68:
```suggestion
    assert mock_fs.patches["insights_client.MOTD_FILE"].new.exists()
```

If the samefile assert above succeeds it implies that the file exists. In case of absence of either of the files, a `FileNotFoundError` would be raised.

```suggestion
```

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+def test_present(mock_fs):
+    assert not (mock_fs.chroot / "etc/motd.d/insights-client").exists()
+
+    # The file gets created/symlinked when .registered & .unregistered do not exist
+    insights_client.update_motd_message()
+    assert (mock_fs.chroot / "etc/insights-client/insights-client.motd").samefile(
+        mock_fs.chroot / "etc/motd.d/insights-client"
+    )
```
Glutexo, line 67:
```suggestion
    assert mock_fs.patches["insights_client.MOTD_SRC"].new.samefile(
        mock_fs.patches["insights_client.MOTD_FILE"].new
    )
```

It feels more natural to me to swap the two files. I mean, what I am verifying is the MOTD_FILE and I want to know whether it’s the same file as its template, the MOTD_SRC. Just as I’d write e.g. `motd_path == "/etc/motd.d/insights-client"` rather than `"/etc/motd.d/insights-client" == motd_path` or `actual_path == expected_path` rather than `expected_path == actual_path`. I know that some languages and frameworks write it the opposite way, but I think it’s not the case of Python.

```suggestion
    assert (mock_fs.chroot / "etc/motd.d/insights-client").samefile(
        mock_fs.chroot / "etc/insights-client/insights-client.motd"
    )
```

How would it look like if variables were used to make the assert shorter?

```suggestion
    motd_src = mock_fs.chroot / "etc/insights-client/insights-client.motd"
    motd_file = mock_fs.chroot / "etc/motd.d/insights-client"
    assert motd_src.samefile(motd_file)
```

src/insights_client/tests/test_motd.py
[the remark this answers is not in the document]
Glutexo: Should tests test themselves? 🤔 I suggested creating separate tests to verify correct working of the fixture. That should do the trick and in case of failure point to the problem.

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,116 @@
...
+    assert (mock_fs.chroot / "etc/motd.d/insights-client").exists()
+
+
+def test_absent_on_dot_registered(mock_fs):
+    # The MOTD file exists by default...
+    insights_client.update_motd_message()
```
Glutexo, line 73: _update_motd_message_ should not be used to create the MOTD symlink. There is already [test](https://github.com/m-horky/insights-client/blob/d5a2300b8eda44d6607892307a67107ad5dfc14b/src/insights_client/tests/test_motd.py#L60) for its creation. This test should focus solely on the removal and thus we should create the symlink manually.

```suggestion
    (mock_fs.chroot / "etc/motd.d"/insights-client").symlink_to(mock_fs.chroot / "etc/insights-client/insights-client.motd")
```

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,116 @@
...
+    assert (mock_fs.chroot / "etc/motd.d/insights-client").exists()
+    # ...and gets removed when .unregistered exists.
+    (mock_fs.chroot / "etc/insights-client/.unregistered").open("w").close()
+
+    insights_client.update_motd_message()
+    assert not (mock_fs.chroot / "etc/motd.d/insights-client").exists()
+
+    # It stays absent when run multiple times.
+    insights_client.update_motd_message()
+    assert not (mock_fs.chroot / "etc/motd.d/insights-client").exists()
```
Glutexo, line 98: This [too](https://github.com/RedHatInsights/insights-client/pull/202/files/d5a2300b8eda44d6607892307a67107ad5dfc14b#diff-cd2e501f6e57473e8f3436595c4f2eca1ca8ee2d50c9bac72b228fa219d42012) is rather a separate use case. Up to here, the test verifies that the MOTD file gets removed if _.uregistered_ file is present. That is remains absent (doesn’t get created) if it already doesn’t exist is a different scenario.

```suggestion

def test_still_absent_on_dot_unregistered(mock_fs):
    # It stays absent when run multiple times.
    (mock_fs.chroot / "etc/insights-client/.unregistered").open("w").close()
    insights_client.update_motd_message()
    assert not (mock_fs.chroot / "etc/motd.d/insights-client").exists()
```

Glutexo, line 98: To reduce repetition, these two casts can be parametrized:

```suggestion
@pytest.mark.parametrize(
    ["dot_file"],
    [
        ("etc/insights-client/.registered",),
        ("etc/insights-client/.unregistered",),
    ]
)
def test_absent_on_dot_file(mock_fs, dot_file):
    # The MOTD file exists by default...
    insights_client.update_motd_message()
    assert (mock_fs.chroot / "etc/motd.d/insights-client").exists()
    # ...and gets removed when the dot file exists.
    (mock_fs.chroot / dot_file).open("w").close()

    insights_client.update_motd_message()
    assert not (mock_fs.chroot / "etc/motd.d/insights-client").exists()

    # It stays absent when run multiple times.
    insights_client.update_motd_message()
    assert not (mock_fs.chroot / "etc/motd.d/insights-client").exists()
```

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+def test_ignored_on_dev_null(mock_fs):
+    # When the /etc/motd.d/insights-client is a symbolic link to /dev/null...
```
Glutexo, line 102: I am not sure whether the comment adds any information that isn’t already conveyed by the test name and the following [instruction](https://github.com/m-horky/insights-client/blob/d5a2300b8eda44d6607892307a67107ad5dfc14b/src/insights_client/tests/test_motd.py#L103).

```suggestion
```

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,116 @@
...
+    (mock_fs.chroot / "etc/motd.d/insights-client").symlink_to(pathlib.Path(os.devnull))
```
Glutexo, line 103:
```suggestion
    (mock_fs.chroot / "etc" / "motd.d" / "insights-client").symlink_to(pathlib.Path(os.devnull))
```

```suggestion
    mock_fs.chroot.patches["insights_client.MOTD_FILE"].new.symlink_to(pathlib.Path(os.devnull))
```

Glutexo, line 103: The [symlink_to](https://docs.python.org/ja/3.6/library/pathlib.html#pathlib.Path.symlink_to) call would work even with plain [os.devnull](https://docs.python.org/ja/3.6/library/os.html#os.devnull), which is a plain string.

```suggestion
    (mock_fs.chroot / "etc/motd.d/insights-client").symlink_to(os.devnull)
```

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,116 @@
...
+
+    # ...it should not be overwritten...
+    insights_client.update_motd_message()
+    assert (mock_fs.chroot / "etc/motd.d/insights-client").samefile(
```
Glutexo, line 107:
```suggestion
    assert (mock_fs.chroot / "etc" / "motd.d" / "insights-client").samefile(
```

```suggestion
    assert mock_fs.chroot.patches["insights_client.MOTD_FILE"].new.samefile(
```

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,116 @@
...
+        pathlib.Path(os.devnull)
```
Glutexo, line 108:
```suggestion
        os.devnull
```

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,116 @@
...
+    )
+
+    # ...whether the MOTD file should be present or not.
+    (mock_fs.chroot / "etc/insights-client/.registered").open("w").close()
```
Glutexo, line 112:
```suggestion
    (mock_fs.chroot / "etc" / "insights-client" / ".registered").open("w").close()
```

```suggestion
    mock_fs.chroot.patches["insights_client.REGISTERED_FILE"].new.open("w").close()
```

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,116 @@
...
+    insights_client.update_motd_message()
+    assert (mock_fs.chroot / "etc/motd.d/insights-client").samefile(
+        pathlib.Path(os.devnull)
```
Glutexo, line 115:
```suggestion
        os.devnull
```

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,116 @@
...
+    # ...it should not be overwritten...
+    insights_client.update_motd_message()
+    assert (mock_fs.chroot / "etc/motd.d/insights-client").samefile(
+        pathlib.Path(os.devnull)
+    )
+
+    # ...whether the MOTD file should be present or not.
+    (mock_fs.chroot / "etc/insights-client/.registered").open("w").close()
+    insights_client.update_motd_message()
+    assert (mock_fs.chroot / "etc/motd.d/insights-client").samefile(
```
Glutexo, line 114:
```suggestion
    assert (mock_fs.chroot / "etc" / "motd.d" / "insights-client").samefile(
```

```suggestion
    assert mock_fs.chroot.patches["insights_client.MOTD_FILE"].new.samefile(
```

Glutexo, line 114: A variable may come handy, saving not only repetition, but also simplifying syntax.

```suggestion
    motd_file = mock_fs.chroot / "etc/motd.d/insights-client"
    motd_file.symlink_to(pathlib.Path(os.devnull))

    # ...it should not be overwritten...
    insights_client.update_motd_message()
    assert motd_file.samefile(pathlib.Path(os.devnull)

    # ...whether the MOTD file should be present or not.
    (mock_fs.chroot / "etc/insights-client/.registered").open("w").close()
    insights_client.update_motd_message()
    assert motd_file.samefile(
```

src/insights_client/tests/test_motd.py
Glutexo, line 112: This only verifies the _.registered_ case, not the _.unregistered_ one. Parametrization can be utilized.

Also, is a different test scenario from the one where none of the dot files exists. Splitting those would make parametrization simpler.

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,116 @@
...
+        pathlib.Path(os.devnull)
+    )
```
Glutexo, line 116:
```suggestion
    (mock_fs.chroot / "etc/insights-client/.registered").open("w").close()
    insights_client.update_motd_message()
    assert (mock_fs.chroot / "etc/motd.d/insights-client").samefile(
        pathlib.Path(os.devnull)
    )

    (mock_fs.chroot / "etc/insights-client/.unregistered").open("w").close()
    insights_client.update_motd_message()
    assert (mock_fs.chroot / "etc/motd.d/insights-client").samefile(
        pathlib.Path(os.devnull)
    )
```

```suggestion
    for dot_file in [
        "etc/insights-client/.registered",
        "etc/insights-client/.unregistered"
    ]:
    (mock_fs.chroot / dot_file).open("w").close()
    insights_client.update_motd_message()
    assert (mock_fs.chroot / "etc/motd.d/insights-client").samefile(
        pathlib.Path(os.devnull)
    )
```

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,116 @@
...
+        paths: typing.Dict[str, str] = {
+            "insights_client.MOTD_SRC": "/etc/insights-client/insights-client.motd",
+            "insights_client.MOTD_FILE": "/etc/motd.d/insights-client",
+            "insights_client.REGISTERED_FILE": "/etc/insights-client/.registered",
+            "insights_client.UNREGISTERED_FILE": "/etc/insights-client/.unregistered",
+        }
+
+        self.patches: typing.List[unittest.mock.patch] = []
+        for target, path in paths.items():
+            mocked_path = f"{self.chroot}{path}"
```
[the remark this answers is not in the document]
m-horky: We still need to stringify it to make the patch work two lines below, but sure.

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,116 @@
...
+
+@pytest.fixture(scope="function")
+def mock_fs():
+    fs = MockFS()
+    yield fs
+    del fs
+
+
+def test_present(mock_fs):
+    assert not (mock_fs.chroot / "etc/motd.d/insights-client").exists()
```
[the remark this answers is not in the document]
m-horky: These lines are just verifying we transition from "does not exist" to "exists". Don't read too much into it. Since we know we are only manipulating symlinks, it doesn't make much sense to read the file, it should be enough it is present and is not broken symlink.

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,116 @@
...
+        self.patches: typing.List[unittest.mock.patch] = []
+        for target, path in paths.items():
+            mocked_path = f"{self.chroot}{path}"
+            pathlib.Path(mocked_path).parent.mkdir(parents=True, exist_ok=True)
+            patch = unittest.mock.patch(target, mocked_path)
+            self.patches.append(patch)
+            patch.start()
+
+        with (self.chroot / "etc/insights-client/insights-client.motd").open("w") as f:
+            f.write("This is the content of the MOTD file.")
```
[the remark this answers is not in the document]
m-horky: It doesn't matter, but we could make it end in newline.

src/insights_client/tests/test_motd.py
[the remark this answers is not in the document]
m-horky: I think we should keep the chrooted paths exposed to make the debugging easier, when something breaks.

src/insights_client/tests/test_motd.py
m-horky: Sure. It also makes it consistent with the `.samefile` asserts in the /dev/null tests.

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,116 @@
...
+    # The file gets created/symlinked when .registered & .unregistered do not exist
+    insights_client.update_motd_message()
+    assert (mock_fs.chroot / "etc/insights-client/insights-client.motd").samefile(
+        mock_fs.chroot / "etc/motd.d/insights-client"
+    )
+    assert (mock_fs.chroot / "etc/motd.d/insights-client").exists()
+
+
+def test_absent_on_dot_registered(mock_fs):
+    # The MOTD file exists by default...
```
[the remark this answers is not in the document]
m-horky: > Does it? It’s created by the call on the next line. I wouln’t necessarily call it “it exists by default”.

Rewording as "gets created by default".

> There is no assert that would verify the file is not there at the beginning, although there is one in the previous test.

The previous test ensures the file got created. These two verify it gets removed.

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,116 @@
...
+    insights_client.update_motd_message()
+    assert (mock_fs.chroot / "etc/motd.d/insights-client").exists()
```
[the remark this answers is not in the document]
m-horky: I'd call it a sanity check. We don't want to continue if something else went wrong already.

src/insights_client/tests/test_motd.py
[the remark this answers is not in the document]
m-horky: Good point.

src/insights_client/tests/test_motd.py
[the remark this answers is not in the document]
m-horky: As I said, it only makes it more confusing to debug.

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,116 @@
...
+    # ...and gets removed when .registered exists.
+    (mock_fs.chroot / "etc/insights-client/.registered").open("w").close()
+
+    insights_client.update_motd_message()
+    assert not (mock_fs.chroot / "etc/motd.d/insights-client").exists()
+
+    # It stays absent when run multiple times.
+    insights_client.update_motd_message()
+    assert not (mock_fs.chroot / "etc/motd.d/insights-client").exists()
```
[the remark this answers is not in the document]
m-horky: I disagree, it's still one single test. The method only reflects the state, it doesn't change anything by (not) creating the file.

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,116 @@
...
+
+    # The file gets created/symlinked when .registered & .unregistered do not exist
+    insights_client.update_motd_message()
+    assert (mock_fs.chroot / "etc/insights-client/insights-client.motd").samefile(
+        mock_fs.chroot / "etc/motd.d/insights-client"
+    )
+    assert (mock_fs.chroot / "etc/motd.d/insights-client").exists()
+
+
+def test_absent_on_dot_registered(mock_fs):
```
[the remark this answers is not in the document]
m-horky: True. I've decided to use state-based names instead here.

src/insights_client/tests/test_motd.py
m-horky: I disagree. It's a test setup, it can be done by the code we're testing.

src/insights_client/tests/test_motd.py
m-horky: Already done before I even got here :)

src/insights_client/tests/test_motd.py
m-horky: It makes it a one-liner, sure.

src/insights_client/tests/test_motd.py
m-horky: The test verifies that it is absent if other file exists. It is still one test and one use-case, only two paths to get there.

src/insights_client/tests/test_motd.py
m-horky: I like to be explicit.

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,101 @@
...
+            self.patches.append(patch)
+            patch.start()
+
+        with (self.chroot / "etc/insights-client/insights-client.motd").open("w") as f:
+            f.write("This is the content of the MOTD file.\n")
+
+    @property
+    def chroot(self) -> pathlib.Path:
+        return pathlib.Path(self._dir.name)
+
```
Glutexo, line 46: I thought up a way how to get the paths from the patches in a more cocsise may. I am not entirely sure about the semantics, but it doesn’t feel completely off either.

```suggestion


    def __getitem__(self, key):
        self.patches[key].new
```

A magic method like this would allow calls like `mock_fs["insights_client.MOTD_FILE"]`. Imagine:

```python
assert not mock_fs["insights_client.MOTD_FILE"].exists()
```

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,101 @@
...
+    (mock_fs.chroot / "etc/motd.d/insights-client").symlink_to(pathlib.Path(os.devnull))
+
+    # ...it should not be overwritten...
+    insights_client.update_motd_message()
+    assert (mock_fs.chroot / "etc/motd.d/insights-client").samefile(os.devnull)
+
+    # ...whether the MOTD file should be present or not.
+    (mock_fs.chroot / "etc/insights-client/.registered").open("w").close()
+    insights_client.update_motd_message()
+    assert (mock_fs.chroot / "etc/motd.d/insights-client").samefile(os.devnull)
```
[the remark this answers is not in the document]
m-horky: Yeah. I still don't agree. These tests are state-based, not action-based. The unit test ensures the files is (not) present, and we verify it through one or two paths to get there. Let's not go philosophical.

src/insights_client/tests/test_motd.py
m-horky: It's getting complicated. It is not a production code. I prefer readability over smartness in tests, it can save you significant amount of time when it goes wrong.

m-horky:
I've introduced further changes that should make the tests more readable. The MockFS now wraps the calls we're using (touch, exists) in a small interface that abstracts away the chroot completely.

src/insights_client/tests/test_motd.py
Glutexo: Which makes it worth separate test functions.

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,113 @@
...
+        paths: typing.Dict[str, str] = {
+            "insights_client.MOTD_SRC": "/etc/insights-client/insights-client.motd",
+            "insights_client.MOTD_FILE": "/etc/motd.d/insights-client",
+            "insights_client.REGISTERED_FILE": "/etc/insights-client/.registered",
+            "insights_client.UNREGISTERED_FILE": "/etc/insights-client/.unregistered",
+        }
+
+        self.patches: typing.List[unittest.mock.patch] = []
+        for target, path in paths.items():
+            mocked_path = f"{self._chroot}{path}"
```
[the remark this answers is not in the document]
m-horky: It is still part of MockFS, it is its `__init__`. I tried your suggestion, but it seems to create more confusion than it solves, I think it should be fine.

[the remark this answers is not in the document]
ptoscano: Oh... not sure why I thought it was in an actual test, sorry.

That said, there is still a simplification possible then for the current code:
```python
            mocked_path = f"{self._chroot}{path}"
            pathlib.Path(mocked_path).parent.mkdir(parents=True, exist_ok=True)
```
- `self._chroot` is a `Path`, which needs to be constructed from `self._dir` (so `str` -> `Path`)
- `mocked_path = f"{self._chroot}{path}"` creates a new path as string by stringifying the elements, including `self.chroot` (so `Path` -> `str` -- back to `str`)
- `pathlib.Path(mocked_path)` creates a new `Path` from the string (so `str` -> `Path` -- `Path` again)

let's simplifying things a bit? :)
```python
            mocked_path = self._chroot / path
            mocked_path.parent.mkdir(parents=True, exist_ok=True)
```

Archana-PandeyM:
motd tests failed downstream. Please analyse the failure.

integration-tests/test_motd.py
```diff
@@ -0,0 +1,72 @@
...
+    yield
+    delete_files()
+
+
+@pytest.mark.usefixtures("delete_special_files")
+def test_motd(insights_client):
+    """MOTD only exists on unregistered system without (.un)registered files."""
+    # If the system is not registered, the file should be present.
+    insights_client.run("--status")
+    assert os.path.exists(MOTD_PATH)
```
Archana-PandeyM, line 46: this would always fail, because delete_special_files would the motd path

m-horky: Yeah, it was a bad copy-paste. Fixed.

integration-tests/test_motd.py
```diff
@@ -0,0 +1,72 @@
...
+
+    # After registration, the file should not exist.
+    insights_client.register()
+    assert not os.path.exists(MOTD_PATH)
```
Archana-PandeyM, line 50: IMO this is not the right way to test it, forcefully deleting motd file and then performing registration does not mimic the right test scenario.
test case could be after registration the  motd file should be deleted but if we already deleted manually what are we testing

Archana-PandeyM: my suggestion is to not delete motd in delete_special_files

Archana-PandeyM:
LGTM. Downstream tests failed because those are running on older version of insights.

src/insights_client/tests/test_motd.py
Glutexo: Here, though, the internal name only serves as a string key to grab the path [defined](https://github.com/m-horky/insights-client/blob/d5a2300b8eda44d6607892307a67107ad5dfc14b/src/insights_client/tests/test_motd.py#L25) here in the tests.

src/insights_client/tests/test_motd.py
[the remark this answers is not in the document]
Glutexo: The philosophy behind the code is as important as the code itself. Otherwise it’s not possible to be correct and consistent.

src/insights_client/tests/test_motd.py
```diff
@@ -0,0 +1,113 @@
...
+    in a temporary directory instead of working with actual system files.
+    """
+
+    _dir: tempfile.TemporaryDirectory
+
+    @property
+    def _chroot(self) -> pathlib.Path:
+        return pathlib.Path(self._dir.name)
+
+    def touch(self, path: str):
```
Glutexo, line 26: Although the [touch](https://linux.die.net/man/1/touch) command is often (mis)used to create a file, it’s primary intended purpose is to set a timestamp. Since this method only mimics the side-effect and not the original function, the name is misleading.

src/insights_client/tests/test_motd.py
Glutexo: The whole [MockFS](https://github.com/RedHatInsights/insights-client/blob/4cab056f37c61d195e6f2567fa0363644caf235a/src/insights_client/tests/test_motd.py#L13) class with its seven methods brings already a great amount of smartness, and thus reducing streamlineliness.

src/insights_client/tests/test_motd.py
m-horky: from `touch (5)`:

> Update the access and modification times of each FILE to the current time.
> A FILE argument that does not exist is created empty,

The first paragraph documents this behavior. It is not misleading, `touch` is really common when you want to create an empty file on UNIX-like systems.

src/insights_client/tests/test_motd.py
ptoscano: > Here, though, the internal name only serves as a string key to grab the path [defined](https://github.com/m-horky/insights-client/blob/d5a2300b8eda44d6607892307a67107ad5dfc14b/src/insights_client/tests/test_motd.py#L25) here in the tests.

That's still missing the point of what I wrote above: the fact that there are mocks to redirect the file operation is a semi-internal implementation detail of the test. The point of the test is checking the resulting paths (that they exist, or what they are, etc), and thus it is correct to use their names in checks/assertions. Once the mocks are set up, from the POV of the tests they should "not be seen" (other than using mock_fs for checks).

So no to your proposal: using the names of the mocks in tests is logically wrong.

PR #204, opened 2024-03-21
Adding tests to insights-client/test_registration

zpetrace:
Edited one tests to better suit the CCT-161 latest changes and created 2 more tests to test CCT-161

### DIFF
--- a/integration-tests/test_registration.py
+++ b/integration-tests/test_registration.py
@@ -1,6 +1,7 @@
 import os
-import contextlib
 import pytest
+import contextlib
+from pytest_client_tools.util import Version
 
 
 pytestmark = pytest.mark.usefixtures("register_subman")
@@ -40,7 +41,11 @@ def test_machineid_changes_on_new_registration(insights_client):
     with open(MACHINE_ID_FILE, "r") as f:
         machine_id_new = f.read()
 
-    assert machine_id_new != machine_id_old
+    if insights_client.core_version >= Version(3, 3, 16):
+        """after the new changes to CCT-161 machine-id stays the same"""
+        assert machine_id_new == machine_id_old
+    else:
+        assert machine_id_new != machine_id_old
 
 
 def test_double_registration(insights_client):
@@ -93,3 +98,17 @@ def test_register_group_option(insights_client, legacy_upload_value):
         check=False,
     )
     assert register_group_option.returncode == 0
+
+
+def test_registered_and_unregistered_files_are_created_and_deleted(insights_client):
+    """'.registered and .unregistered file gets created and deleted"""
+    assert not insights_client.is_registered
+    assert not os.path.exists("/etc/insights-client/.registered")
+
+    insights_client.register()
+    assert os.path.exists("/etc/insights-client/.registered")
+    assert not os.path.exists("/etc/insights-client/.unregistered")
+
+    insights_client.unregister()
+    assert os.path.exists("/etc/insights-client/.unregistered")
+    assert not os.path.exists("/etc/insights-client/.registered")


integration-tests/test_registration.py
```diff
@@ -93,3 +98,38 @@ def test_register_group_option(insights_client, legacy_upload_value):
         check=False,
     )
     assert register_group_option.returncode == 0
+
+def test_machineid_changes_when_subman_is_unregistered_and_registered(insights_client, subman, test_config):
+    """machine-id should be different based on the auth method"""
+    if 'satellite' in test_config.environment:
+        pytest.skip("this test requires BASIC authentication")
+
+    insights_client.unregister()
```
Archana-PandeyM, line 107: insights_client.unregister() will fail at this point if already unregistered, you can try adding insights_client.unregister(check=False)

ptoscano: strictly speaking you don't need to explicitly `unregister()`, as using the `insights_client` fixture (as used elsewhere) ensures that `insights-client` is unregistered automatically at the end of each test

Archana-PandeyM: Like Pino said we don't need to unregister so it can be removed also.

integration-tests/test_registration.py
```diff
@@ -93,3 +98,38 @@ def test_register_group_option(insights_client, legacy_upload_value):
...
+    with open(MACHINE_ID_FILE, "r") as f:
+        machine_id_random = f.read()
+
+    assert machine_id_random != machine_id_cert
+
+
+def test_registered_and_unregistered_files_are_created_and_deleted(insights_client):
+    """""'.registered and .unregistered file gets created and deleted"""
+    assert not insights_client.is_registered
+
```
Archana-PandeyM, line 130: I would suggest to assert here that .registered doesn't exist before registration and similiar assertion for .unregister before unregistration

integration-tests/test_registration.py
```diff
@@ -93,3 +98,43 @@ def test_register_group_option(insights_client, legacy_upload_value):
...
+    insights_client.register()
+    with open(MACHINE_ID_FILE, "r") as f:
+        machine_id_cert = f.read()
+    insights_client.unregister()
+
+    insights_client.config.auto_config = False
+    insights_client.config.authmethod = "BASIC"
+    insights_client.config.username = test_config.get("candlepin", "username")
+    insights_client.config.password = test_config.get("candlepin", "password")
+    insights_client.config.save()
```
m-horky, line 120: After some testing, this doesn't look like a good way of doing it, because automatic configuration also configured prod/stage access.

Instead, I think we'll need to unregister and register sub-man here.

```py
try:
    subman.unregister()
    
    insights_client.register()
    ...
finally:
    subman.register()

assert ...
```

integration-tests/test_registration.py
```diff
@@ -40,7 +41,11 @@ def test_machineid_changes_on_new_registration(insights_client):
     with open(MACHINE_ID_FILE, "r") as f:
         machine_id_new = f.read()
 
-    assert machine_id_new != machine_id_old
+    if insights_client.core_version >= Version(3, 3, 13):
```
m-horky, line 44: I know you know this, but just for completion: the actual version needs to be changed before merging.

ptoscano (commented):
`test_machineid_changes_when_insights_auth_method_changes` exists with the idea of verifying that the UUID in the `machine-id` file is different when  registering with BASIC than when registering with CERT auth; this is also influenced by the recent changes in insights-core, i.e. https://github.com/RedHatInsights/insights-core/pull/4057, which makes the subscription-manager UUID also the insights-client UUID. Before the those insights-core changes, the UUID was always a random one, while after those, it is the subscription-manager one when the system is registered with subscription-manager; in both cases the way it is generated it is considered an implementation detail.

So now we have 3 potential scenarios in which fetch the UUID:

1. BASIC auth when the system is not registered with subscription-manager
1. BASIC auth when the system is registered with subscription-manager
1. CERT auth (the system is definitely registered with subscription-manager)

As @m-horky wrote, to support scenario (1) we would need to unregister with subscription-manager during the test, and that would require then re-registering again (which is what the other tests assume), complicating the sitation a bit.

Scenarios (2) and (3) behave in the same way, as the insights-core changes make use of the subscription-manager UUID no matter the auth method in use.

I though more about this, and the question I came up with is: what is this test actually testing? I do not understand what would be the point of checking that the UUIDs in the scenarios (1) and scenarios (2) & (3) are different. Why do we want to explicitly assert that, considering that the generation of the insights-client UUID is an implementation detail?

zpetrace:
I have deleted the test `test_machineid_changes_when_insights_auth_method_changes` in the latest updates as the changes are tested in other tests and there was no need for this test.

integration-tests/test_registration.py
zpetrace: fixed

PR #205, opened 2024-03-26
Tests for display-name option

jstavel:
There are two tests in the PR:
-  test_register_with_display_name
- test_register_twice_with_different_display_name

I've added a function 'fetch-from-inventory' to download inventory records from insights service.
I've started using subtests pytest plugin - see the second test in a list above.
subtests help me to separate logs for each part of a test. It is easier to troubleshoot a test - log messages are separated by 'subtest' message - at least.
another reason for using subtests is that a code is splitted into parts visually too.

### DIFF
--- a/integration-tests/test_display_name_option.py
+++ b/integration-tests/test_display_name_option.py
@@ -1,0 +1,135 @@
+import pytest
+import uuid
+import logging
+import fauxfactory
+import json
+
+from conftest import loop_until
+from constants import HOST_DETAILS
+
+logger = logging.getLogger(__name__)
+
+pytestmark = pytest.mark.usefixtures("register_subman")
+
+
+def read_host_details():
+    with open(HOST_DETAILS, "r") as data_file:
+        return json.load(data_file)
+
+
+def generate_unique_hostname():
+    return f"test-qa.{uuid.uuid4()}.csi-client-tools.example.com"
+
+
+def test_display_name(insights_client):
+    """Test insights-client --display-name"""
+    new_hostname = generate_unique_hostname()
+    insights_client.run("--register")
+    assert loop_until(lambda: insights_client.is_registered)
+
+    response = insights_client.run("--display-name", new_hostname)
+    logger.debug(f"response from console {response}")
+
+    assert f"Display name updated to {new_hostname}" in response.stdout
+
+    def display_name_changed():
+        insights_client.run("--check-results")
+        host_details = read_host_details()
+        logger.debug(f"host details {host_details}")
+        record = host_details["results"][0]
+        return new_hostname == record["display_name"]
+
+    assert loop_until(display_name_changed)
+
+
+def test_register_with_display_name(insights_client):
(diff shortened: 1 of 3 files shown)


jstavel:
there is a code block that varies depending on environment - a message from register command varies for 'stage' or 'prod' environment.

integration-tests/test_display_name_option.py
```diff
@@ -0,0 +1,80 @@
...
+
+    The current display name should appear in inventory
+    """
+    unique_hostname = f"test-qa.{uuid.uuid4()}.csi-client-tools.example.com"
+
+    status = insights_client.run("--register", "--display-name", unique_hostname)
+    assert loop_until(lambda: insights_client.is_registered)
+    assert (
+        "Successfully registered host" in status.stdout
+    ), "An application should inform about successfull registration"
```
ptoscano, line 22: this is already asserted by the basic registration tests, I don't think there is the need to repeat the checks here too

integration-tests/test_display_name_option.py
```diff
@@ -0,0 +1,80 @@
...
+    The `register` method does check if the host changed at all,
+    it only checks the machine_id
+    """
+    insights_id = None
+    unique_hostname = f"test-qa.{uuid.uuid4()}.csi-client-tools.example.com"
+    unique_hostname_02 = f"test-qa.{uuid.uuid4()}.csi-client-tools.example.com"
+
+    with subtests.test(msg="the first registration"):
+        status = insights_client.run("--register", "--display-name", unique_hostname)
+        assert "Successfully registered host" in status.stdout
```
ptoscano, line 52: ditto

ptoscano (changes requested):
I'm not sure what this "second registration" actually refers to: the existing `test_double_registration` already asserts that trying to register on an already registered system does nothing. This step here in practice is simply `insights-client --display-name=...`, so I'd do that only.

Not doing a "double registration" simplifies things a bit, and also removes the need to use the subtest plugin.

-----

Regarding the `fetch_from_inventory` fixture: while I see where this is coming, I think we need a more generic way to query the Insights inventory, rather than this ad-hoc fixture. I have a draft of work for this to be added in pytest-client-tools, so please hold this off for now.

-----

Thanks!

jstavel:
added another tests:
- test_display_name .. to verify that an application can set display-name
- test_invalid_display_name

integration-tests/test_display_name_option.py
jstavel: this is a case when user register a system. He tries to register the system once more, with different display-name. Even this case may not make sense - but it may happen. So I think it is worth automating it.

Lorquas: Lets do this: 

1.  Create a jira to create `fetch_from_inventory`  functionality in client-tools, and block this on that card.
2.  remove it from this PR
3. merge this once thats working.

I think the rest of this is fine as of the latest change.

PR #206, opened 2024-03-27
tests: add test cases for data redaction

zhangqianqian:


jstavel:
hi @zhangqianqian please consider a helper function conftest.loop_until. An example how to use it:
assert conftest.loop_until(lambda: insights_client.is_registered)  it could help you in cases when a synchronization process takes more than 5 seconds. see docstring for the function for details. instead of using sleep(5).

integration-tests/test_redaction.py
```diff
@@ -0,0 +1,163 @@
...
+        file_content = {"commands": [redaction_string]}
+    elif redaction_type == "file":
+        file_content = {"files": [redaction_string]}
+    with open(FILE_REDACTION_FILE, "w") as f:
+        yaml.dump(file_content, f, default_flow_style=False)
+    os.chmod(FILE_REDACTION_FILE, 0o600)
+
+
+def check_archive(insights_client, tested):
+    archive_result = insights_client.run("--no-upload")
```
Archana-PandeyM, line 142: I have a suggestion, instead of using regex why not use "insights-client --no-upload --output-file=  ". directly access the named archive .

zhangqianqian:
> hi @zhangqianqian please consider a helper function conftest.loop_until. 
Done

ptoscano (changes requested):
I see that `test_obfuscation` was moved from `test_redaction.py`; while I'm not for not against it, I'd recommend to do it as first commit, moving the file with `git mv`, and then adding more tests to it.

Also, in the commit message for this, please be a bit more verbose on what is the reason for this.

-------

Please fold the changes of the two `use conftest.loop_until in test_redaction.py` commits together with the commits that add/change those tests.

-------

I see that the new tests first register with `insights-client`, then they run a new collection that is not uploaded: this means that the collection is run twice per test. Since each data collection takes while a while (at least 3 minutes), this makes the new tests quite expensive.

OTOH, Archana is spot on here: for all these tests that need to check something in the collected data, I think we can simplify things a bit:
- do not register them, since there is no need to be registered to perform a data collection
- run the equivalent of `insights-client --no-upload --output-file=somefile.tar.gz` (remember to use the `tmp_path` fixture for the location of the temporary file)

This will also make it possible to run these tests offline, without an Insight server configured/usable.

(the same change can be applied to the existing `configure_redaction`/`test_redaction_on_pattern_hostname` -- let's keep this to a followup PR)

-------

Thanks!

zhangqianqian:
> I see that `test_obfuscation` was moved from `test_redaction.py`; while I'm not for not against it, I'd recommend to do it as first commit, moving the file with `git mv`, and then adding more tests to it.

https://github.com/RedHatInsights/insights-client/blob/master/integration-tests/test_obfuscation.py#L12 comes from https://gitlab.cee.redhat.com/insights-qe/iqe-insights-client-plugin/-/blob/master/iqe_insights_client/tests/test_satellite.py?ref_type=heads#L19, as you suggested to split the test cases in test_satellite.py.


When I began to migrate test_redaction.py in this PR, I found https://gitlab.cee.redhat.com/insights-qe/iqe-insights-client-plugin/-/blob/master/iqe_insights_client/tests/test_redaction.py?ref_type=heads is for file redaction(/etc/insights-client/file-redaction.yaml), and content redaction(/etc/insights-client/file-content-redaction.yaml) is tested by https://github.com/RedHatInsights/insights-client/blob/master/integration-tests/test_obfuscation.py#L12. So I move test_obfuscation from test_obfuscation.py to test_redaction.py and rename the test case as test_redaction_on_pattern_hostname. By the way, while waiting for merging this PR, I also started to migrate test_obfuscate.py https://gitlab.cee.redhat.com/insights-qe/iqe-insights-client-plugin/-/blob/master/iqe_insights_client/tests/test_obfuscate.py?ref_type=heads. So I did not use ‘git mv’.


> I see that the new tests first register with `insights-client`, then they run a new collection that is not uploaded: this means that the collection is run twice per test. Since each data collection takes while a while (at least 3 minutes), this makes the new tests quite expensive.
> 
> OTOH, Archana is spot on here: for all these tests that need to check something in the collected data, I think we can simplify things a bit:
> 
> * do not register them, since there is no need to be registered to perform a data collection
> * run the equivalent of `insights-client --no-upload --output-file=somefile.tar.gz` (remember to use the `tmp_path` fixture for the location of the temporary file)
> 
> This will also make it possible to run these tests offline, without an Insight server configured/usable.
> 
I agree to use --output-file=somefile.tar.gz with tmp_path fixture. But when we run 'insights-client --register', downloading latest egg happens which means: if we run `insights-client --no-upload --output-file=somefile.tar.gz` , we maybe not use the latest egg. So I prefer use "insights-client --register --no-upload --output-file=somefile.tar.gz".

integration-tests/test_obfuscation.py
```diff
@@ -1,71 +1,168 @@
...
+    """
+    # Obfuscate IP
+    insights_client.config.obfuscate = True
+    insights_client.config.save()
+
+    # Check if the package version is obfuscated by IP obfuscation
+    archive_name = "test_archive_" + str(uuid.uuid4()) + ".tar.gz"
+    archive_location = tmp_path / archive_name
+    insights_client.run(
+        "--register", "--no-upload", "--output-file=%s" % archive_location
```
Archana-PandeyM, line 109: no need for --no-upload, because --output-file runs in offline mode only.

Archana-PandeyM (changes requested):

integration-tests/test_obfuscation.py
zhangqianqian: Done

integration-tests/test_redaction.py
```diff
@@ -0,0 +1,136 @@
...
+
+    finally:
+        set_hostname(previous_hostname)
+        os.remove(CONTENT_REDACTION_FILE)
+
+
+def check_archive(insights_client, tmp_path, tested):
+    archive_name = "test_archive_" + str(uuid.uuid4()) + ".tar.gz"
+    archive_location = tmp_path / archive_name
+    logging.info(archive_location)
```
Archana-PandeyM, line 103: nit: we can remove this logging statement from here as it does not add significant value because tmp_path erases this path after test anyways

integration-tests/test_redaction.py
```diff
@@ -0,0 +1,136 @@
...
+    insights_client.run("--register", "--output-file=%s" % archive_location)
+    check_file = "/" + tested.split("/")[-1]
+    if check_file == "/ls":
+        check_file = "/ls_"
+    with tarfile.open(archive_location, "r") as tar:
+        check_tested = [
+            file_name for file_name in tar.getnames() if check_file in file_name
+        ]
+    logging.info(check_tested)
```
Archana-PandeyM, line 112: same.

integration-tests/test_redaction.py
```diff
@@ -0,0 +1,136 @@
...
+                    file_content = extracted_file.read().decode()
+                    assert hostname_to_test not in file_content
+
+    finally:
+        set_hostname(previous_hostname)
+        os.remove(CONTENT_REDACTION_FILE)
+
+
+def check_archive(insights_client, tmp_path, tested):
+    archive_name = "test_archive_" + str(uuid.uuid4()) + ".tar.gz"
```
Archana-PandeyM, line 101: nit: tmp_path is temporary so no need for uuid to get a unique path we can replace these two lines of code with something like:
archive_location = tmp_path / "test_archive.tar.gz"

zhangqianqian: Done

Archana-PandeyM:
LGTM. Please squash and clean commits.

integration-tests/test_redaction.py
```diff
@@ -0,0 +1,131 @@
...
+
+    """
+
+    # Record the current hostname
+    with open("/etc/hostname", "r") as f:
+        previous_hostname = f.read().strip()
+
+    try:
+        # Set a new hostname
+        hostname_to_test = "insights-client-satellite"
```
Archana-PandeyM, line 73: can we make it generic for hosted and satellite by removing 'satellite' word from here?
I think this test applies to both configurations.

Archana-PandeyM (changes requested):

integration-tests/test_redaction.py
zhangqianqian: Done

Archana-PandeyM (approved):

PR #211, opened 2024-04-15
feat(test): Migrating no_upload_test.py upstream

zpetrace:
Migrating no_upload_test.py upstream

### DIFF
--- a/integration-tests/test_client_options.py
+++ b/integration-tests/test_client_options.py
@@ -1,7 +1,11 @@
 import pytest
 import conftest
+import glob
+
 
 pytestmark = pytest.mark.usefixtures("register_subman")
+
+ARCHIVE_CACHE_DIRECTORY = "/var/cache/insights-client"
 
 
 def test_set_ansible_host_info(insights_client):
@@ -17,3 +21,22 @@ def test_set_ansible_host_info(insights_client):
     ret = insights_client.run("--ansible-host=foo.example.com", check=False)
     assert "Could not update Ansible hostname" not in ret.stdout
     assert ret.returncode == 0
+
+
+def test_no_upload(insights_client):
+    """Verify no traceback is returned when running '--no-upload' command
+    and the archive is created and saved on the system
+    """
+    archive_saved = "Archive saved at"
+    upload_message = "Successfully uploaded report"
+
+    insights_client.register()
+
+    archive_file_before = glob.glob(f"{ARCHIVE_CACHE_DIRECTORY}/*.tar.gz")
+
+    no_upload_output = insights_client.run("--no-upload")
+    assert archive_saved in no_upload_output.stdout
+    assert upload_message not in no_upload_output.stdout
+
+    archive_file_after = glob.glob(f"{ARCHIVE_CACHE_DIRECTORY}/*.tar.gz")
+    assert len(archive_file_after) > len(archive_file_before)


integration-tests/test_no_upload.py
```diff
@@ -0,0 +1,29 @@
+import pytest
+import glob
+import os
+import shutil
+
+
+pytestmark = pytest.mark.usefixtures("register_subman")
+
+
+def test_no_upload(insights_client):
```
Archana-PandeyM, line 10: My suggestion is to move this test to test_client_options.py for the reason that it is one of the insights-client options. or if you plan to have more no_upload tests in this class then we can discuss.

zpetrace: done

integration-tests/test_client_options.py
```diff
@@ -17,3 +21,25 @@ def test_set_ansible_host_info(insights_client):
...
+    """
+    archive_saved = "Archive saved at"
+    upload_message = "Successfully uploaded report"
+    archive_directory = "/var/cache/insights-client"
+
+    insights_client.register()
+
+    if os.listdir(archive_directory):
+        shutil.rmtree(archive_directory)
+        os.makedirs(archive_directory)
```
ptoscano, line 38: hmm removing files here IMHO is not a good idea, and I'd rather work regardless of the content of that directory; also, that directory is statically shipped by the RPM package, so there is no need to manually create it

what I'd do instead is the following:
- before the `--no-upload` invocation, `glob()` to get the content of the cache directory
- after `--no-upload`, `glob()` again to get the new content in the directory
- diff the "after" with the "before", and check that there is at least (or exactly) one element more; tip: convert the list to sets using `set()` and diff them easily

integration-tests/test_client_options.py
```diff
@@ -17,3 +21,25 @@ def test_set_ansible_host_info(insights_client):
...
+
+    no_upload_output = insights_client.run("--no-upload")
+    assert archive_saved in no_upload_output.stdout
+    assert upload_message not in no_upload_output.stdout
+
+    archive_file = glob.glob("/var/cache/insights-client/*.tar.gz")
```
ptoscano, line 44: you can use the existing `archive_directory` variable here

integration-tests/test_client_options.py
```diff
@@ -17,3 +21,25 @@ def test_set_ansible_host_info(insights_client):
...
     assert ret.returncode == 0
+
+
+def test_no_upload(insights_client):
+    """Verify no traceback is returned when running '--no-upload' command
+    and the archive is created and saved on the system
+    """
+    archive_saved = "Archive saved at"
+    upload_message = "Successfully uploaded report"
+    archive_directory = "/var/cache/insights-client"
```
ptoscano, line 32: since this variable seems to be useful also in other tests (potentially), I'd make it as a top-level variable, eg:
```python
ARCHIVE_CACHE_DIRECTORY = "/var/cache/insights-client"
```

ptoscano (changes requested):
generally OK, few things to improve -- thanks!

integration-tests/test_client_options.py
zpetrace: done

integration-tests/test_client_options.py
zpetrace: done

integration-tests/test_client_options.py
```diff
@@ -17,3 +21,22 @@ def test_set_ansible_host_info(insights_client):
...
+def test_no_upload(insights_client):
+    """Verify no traceback is returned when running '--no-upload' command
+    and the archive is created and saved on the system
+    """
+    archive_saved = "Archive saved at"
+    upload_message = "Successfully uploaded report"
+
+    insights_client.register()
+
+    archive_file_before = glob.glob(ARCHIVE_CACHE_DIRECTORY)
```
ptoscano, line 35: globbing a directory will always return 1, as the directory itself matches the glob exactly :D

```suggestion
    archive_file_before = glob.glob(f"{ARCHIVE_CACHE_DIRECTORY}/*.tar.gz")
```

integration-tests/test_client_options.py
```diff
@@ -17,3 +21,22 @@ def test_set_ansible_host_info(insights_client):
...
+
+    no_upload_output = insights_client.run("--no-upload")
+    assert archive_saved in no_upload_output.stdout
+    assert upload_message not in no_upload_output.stdout
+
+    archive_file_after = glob.glob(ARCHIVE_CACHE_DIRECTORY)
```
ptoscano, line 41: ditto

ptoscano (changes requested):
almost there! one more fix needed

integration-tests/test_client_options.py
zpetrace: done

ptoscano (approved):
thanks!

PR #212, opened 2024-04-16
Tests: added tests for --group and --support client option

Archana-PandeyM:
Added tests for --support and --group option. Basically migrated tests from iqe-insights-client-plugin .

### DIFF
--- a/integration-tests/test_client_options.py
+++ b/integration-tests/test_client_options.py
@@ -1,7 +1,8 @@
+import json
+import tarfile
 import pytest
 import conftest
 import glob
-
 
 pytestmark = pytest.mark.usefixtures("register_subman")
 
@@ -40,3 +41,49 @@ def test_no_upload(insights_client):
 
     archive_file_after = glob.glob(f"{ARCHIVE_CACHE_DIRECTORY}/*.tar.gz")
     assert len(archive_file_after) > len(archive_file_before)
+
+
+def test_group(insights_client, tmp_path):
+    """Test --group option
+    This test generates an archive in offline mode and verifies if the group
+    specified in --group was created and packed
+    """
+
+    group_name = "testing-group"
+    archive_name = "archive.tar.gz"
+    archive_location = tmp_path / archive_name
+
+    # Running insights-client in offline mode to generate archive
+    insights_client.run(
+        "--offline", f"--group={group_name}", f"--output-file={archive_location}"
+    )
+
+    with tarfile.open(archive_location, "r") as tar:
+        dir_name = tar.getnames()[0]
+        for m in tar.getmembers():
+            if m.name == f"{dir_name}/tags.json":
+                tag_file = tar.extractfile(m)
+                tag_file_content = json.load(tag_file)
+                break
+
+    assert len(tag_file_content) == 1
+
+    tag = tag_file_content[0]
+    assert tag["key"] == "group"
+    assert tag["namespace"] == "insights-client"
+    assert tag["value"] == group_name
+
+
+def test_support(insights_client):
+    """Test if --support option is giving expected information in output
+    and generates a support logfile for Red Hat Insights.
+    """
+    support_result = insights_client.run("--support")
+
+    assert "Insights version:" in support_result.stdout
+    assert "Registration check:" in support_result.stdout
+    assert "Last successful upload was" in support_result.stdout
+    assert "Connectivity tests" in support_result.stdout
+    assert "Running command:" in support_result.stdout
+    assert "Process output:" in support_result.stdout
+    assert "Support information collected in" in support_result.stdout


integration-tests/test_client_options.py
```diff
@@ -17,3 +22,53 @@ def test_set_ansible_host_info(insights_client):
...
+
+    # Running insights-client in offline mode to generate archive
+    insights_client.run(
+        "--offline", f"--group={group_name}", f"--output-file={archive_location}"
+    )
+
+    with tarfile.open(archive_location, "r") as tar:
+        for m in tar.getmembers():
+            insights_archive_name = os.path.dirname(m.name)
+            if m.name == f"{insights_archive_name}/tags.json":
```
ptoscano, line 46: rather than checking for the full path, wouldn't it be easier to simply check that the filename is simply `tags.json`? ie
```python
            if os.path.basename(m.name) == "tags.json":
```

integration-tests/test_client_options.py
```diff
@@ -17,3 +22,53 @@ def test_set_ansible_host_info(insights_client):
...
+                tag_file = tar.extractfile(m.name)
```
ptoscano, line 47: you can pass directly `m` to `extractfile()`

integration-tests/test_client_options.py
```diff
@@ -17,3 +22,53 @@ def test_set_ansible_host_info(insights_client):
...
+                content = tag_file.read()
```
ptoscano, line 48: you can keep `tag_file` here and avoid to `read()` from it here, so ...

integration-tests/test_client_options.py
```diff
@@ -17,3 +22,53 @@ def test_set_ansible_host_info(insights_client):
...
+                break
+
+    assert tag_file is not None, "Tag file not found"
+
+    tag_file_content = json.loads(content.decode())
```
ptoscano, line 53: ... so here you can pass `tag_file` directly to `json.load()`, which accepts an object on which you can `read()`; this way, you don't even need `decode()`

integration-tests/test_client_options.py
```diff
@@ -17,3 +22,53 @@ def test_set_ansible_host_info(insights_client):
...
+
+    tag = tag_file_content[0]
+    assert tag["key"] == "group"
+    assert tag["namespace"] == "insights-client"
+    assert tag["value"] == group_name
+
+
+def test_support(insights_client):
+    """Test if --support option is giving expected information in output
+    and generates a support logfile for RedHat Insights.
```
ptoscano, line 64: typo:
```suggestion
    and generates a support logfile for Red Hat Insights.
```

integration-tests/test_client_options.py
```diff
@@ -17,3 +22,53 @@ def test_set_ansible_host_info(insights_client):
...
+
+
+def test_group(insights_client, tmp_path):
+    """Test --group option
+    This test generates an archive in offline mode and verifies if the group
+    specified in --group was created and packed
+    """
+
+    group_name = f"RHIQE.{uuid.uuid4()}.testing-group"
+    archive_tmp_name = f"test_archive_{uuid.uuid4()}.tar.gz"
```
ptoscano, line 34: since the archive is in a temporary location created new every time the test is run, there is no need to use an unique filename for the archive; simply use `archive.tar.gz` directly

integration-tests/test_client_options.py
```diff
@@ -17,3 +22,53 @@ def test_set_ansible_host_info(insights_client):
...
     assert ret.returncode == 0
+
+
+def test_group(insights_client, tmp_path):
+    """Test --group option
+    This test generates an archive in offline mode and verifies if the group
+    specified in --group was created and packed
+    """
+
+    group_name = f"RHIQE.{uuid.uuid4()}.testing-group"
```
ptoscano, line 33: using an unique UUID is fine, although a bit complicated; maybe simply hardcode something like `testing-group` here?

ptoscano (changes requested):
generally good, there are few things to fix/improve -- thanks!

integration-tests/test_client_options.py
Archana-PandeyM: I thought of doing  so too but there are two tags.json in archive, i.e ../tags.json and ../data/tags.json

ptoscano: there needs to be a different way then, because the current check matches both of them, so in the end whichever is found first is used for the test of the test

integration-tests/test_client_options.py
Archana-PandeyM: using unique testing group name should be fine because these groups gets created in inventory and incase if test failure it might come handy to look for exact group

integration-tests/test_client_options.py
Archana-PandeyM: ok both are fine so I will change it to archive.tar.gz

integration-tests/test_client_options.py
ptoscano: to ensure that you can still use `tag_file` outside of the `for` loop: a possible idea here can be to move the json loading inside the context manager that opens the tarfile, ie
```python
    with tarfile.open(...) as tar:
        for m in tar.getmembers():
            ...
        assert tag_file
        content = json.load(tag_file)
```

integration-tests/test_client_options.py
Archana-PandeyM: ok I could use tar.getnames()[0] to get the dirname and check complete path.

integration-tests/test_client_options.py
```diff
@@ -17,3 +19,50 @@ def test_set_ansible_host_info(insights_client):
...
+    tag_file = None
+
+    # Running insights-client in offline mode to generate archive
+    insights_client.run(
+        "--offline", f"--group={group_name}", f"--output-file={archive_location}"
+    )
+
+    with tarfile.open(archive_location, "r") as tar:
+        for m in tar.getmembers():
+            if m.name == f"{tar.getnames()[0]}/tags.json":
```
ptoscano, line 42: calling `tar.getnames()` at every iteration is a bit expensive; would you please save the output of `tar.getnames()[0]` to a variable before the `for` loop is run?

integration-tests/test_client_options.py
```diff
@@ -17,3 +19,50 @@ def test_set_ansible_host_info(insights_client):
...
+def test_group(insights_client, tmp_path):
+    """Test --group option
+    This test generates an archive in offline mode and verifies if the group
+    specified in --group was created and packed
+    """
+
+    group_name = "testing-group"
+    archive_name = "archive.tar.gz"
+    archive_location = tmp_path / archive_name
+    tag_file = None
```
ptoscano, line 33: since `tag_file` is a class that lives only until the tarfile is open, then I'd use `tag_file_content` instead to determine whether the tags were found

ptoscano (changes requested):
much better now, thanks! there only a couple of minor issues to fix

integration-tests/test_client_options.py
Archana-PandeyM: done

integration-tests/test_client_options.py
Archana-PandeyM: done

integration-tests/test_client_options.py
```diff
@@ -17,3 +19,50 @@ def test_set_ansible_host_info(insights_client):
...
+    insights_client.run(
+        "--offline", f"--group={group_name}", f"--output-file={archive_location}"
+    )
+
+    with tarfile.open(archive_location, "r") as tar:
+        dir_name = tar.getnames()[0]
+        for m in tar.getmembers():
+            if m.name == f"{dir_name}/tags.json":
+                tag_file = tar.extractfile(m)
+                tag_file_content = json.loads(tag_file.read())
```
ptoscano, line 45: `read()` is not needed here, json will do it on its own if you use `load()` instead of `loads()`

ptoscano (changes requested):
thanks! a couple more changes (hopefully the last), and a needed rebase after other tests were merged in the same file

PR #213, opened 2024-04-16
feat(test): Migrating test_connection.py upstream

zpetrace:
Migrating test_connection.py upstream

### DIFF
--- a/integration-tests/test_connection.py
+++ b/integration-tests/test_connection.py
@@ -1,0 +1,95 @@
+import pytest
+
+# the tests need a valid connection to insights so therefore the subman registration
+pytestmark = pytest.mark.usefixtures("register_subman")
+
+
+def test_connection(insights_client):
+    """
+    Test insights-client --test-connection
+    `test_connection` method returns 0 if success, 1 if failure
+    """
+    url_test = "End Upload URL Connection Test: SUCCESS"
+    api_test = "End API URL Connection Test: SUCCESS"
+
+    test_connection = insights_client.run("--test-connection")
+    assert url_test in test_connection.stdout
+    assert api_test in test_connection.stdout
+
+
+def test_http_timeout(insights_client):
+    """
+    Test http_timeout config option
+    Set http_timeout to a very low value, run --test-connection and check
+    the set time
+    """
+    insights_client.config.http_timeout = 0.1
+    insights_client.config.save()
+
+    output = insights_client.run("--test-connection", check=False)
+    assert output.returncode == 1
+    assert "read timeout=0.1" in output.stdout
+
+
+def test_noauth_proxy_connection(insights_client, test_config):
+    """
+    Test no-auth proxy option with --test-connection option
+    """
+    url_test = "End Upload URL Connection Test: SUCCESS"
+    api_test = "End API URL Connection Test: SUCCESS"
+
+    no_auth_proxy: str = (
+        "http://"
+        + test_config.get("noauth_proxy", "host")
+        + ":"
+        + str(test_config.get("noauth_proxy", "port"))
+    )
+    insights_client.config.proxy = no_auth_proxy
+    insights_client.config.save()
+
+    test_connection = insights_client.run("--test-connection")
+    assert url_test in test_connection.stdout
+    assert api_test in test_connection.stdout
+
+
+def test_auth_proxy_connection(insights_client, test_config):
+    """
+    Test auth proxy option with --test-connection
+    """
+    url_test = "End Upload URL Connection Test: SUCCESS"
+    api_test = "End API URL Connection Test: SUCCESS"
+
+    auth_proxy: str = (
+        "http://"
+        + test_config.get("auth_proxy", "username")
+        + ":"
+        + test_config.get("auth_proxy", "password")
+        + "@"
+        + test_config.get("auth_proxy", "host")
+        + ":"
+        + str(test_config.get("auth_proxy", "port"))
+    )
+    insights_client.config.proxy = auth_proxy
+    insights_client.config.save()
+    test_connection = insights_client.run("--test-connection")
+    assert url_test in test_connection.stdout
+    assert api_test in test_connection.stdout
+
+
+def test_wrong_url_connection(insights_client):
+    """
+    Test wrong url option with --test-connection
+    """
+    end_upload_failure = "Failed to establish a new connection"
+    failed_connectivity_test = "Connectivity test failed!"
+
+    insights_client.config.auto_config = False
+    insights_client.config.auto_update = False
+    insights_client.config.base_url = "no-such-insights-url.example.com/something"
+    insights_client.config.authmethod = "CERT"
+    insights_client.config.save()
+
+    test_connection = insights_client.run("--test-connection", check=False)
+    assert test_connection.returncode == 1
+    assert end_upload_failure in test_connection.stdout
+    assert failed_connectivity_test in test_connection.stdout


integration-tests/test_connection.py
```diff
@@ -0,0 +1,107 @@
...
+
+def test_connection(insights_client):
+    """
+    Test insights-client --test-connection
+    `test_connection` method returns 0 if success, 1 if failure
+    """
+    url_test = "End Upload URL Connection Test: SUCCESS"
+    api_test = "End API URL Connection Test: SUCCESS"
+
+    insights_client.register()
```
Archana-PandeyM, line 16: Do we really need to register for this test?? I mean subman is register so certs would be available for connection test. In my opinion you could skip this step.

integration-tests/test_connection.py
```diff
@@ -0,0 +1,107 @@
...
+    assert api_test in test_connection.stdout
+
+
+def test_http_timeout(insights_client):
+    """
+    Test http_timeout config option
+    Set http_timeout to a very low value, run --test-connection and check
+    the set time
+    """
+    insights_client.register()
```
Archana-PandeyM, line 30: same.

integration-tests/test_connection.py
```diff
@@ -0,0 +1,105 @@
...
+
+def test_connection(insights_client):
+    """
+    Test insights-client --test-connection
+    `test_connection` method returns 0 if success, 1 if failure
+    """
+    url_test = "End Upload URL Connection Test: SUCCESS"
+    api_test = "End API URL Connection Test: SUCCESS"
+
+    assert conftest.loop_until(lambda: insights_client.is_registered)
```
ptoscano, line 16: this will actually fail, as each test starts with an unregistered `insights-client`; simply drop this

integration-tests/test_connection.py
```diff
@@ -0,0 +1,105 @@
...
+    assert api_test in test_connection.stdout
+
+
+def test_http_timeout(insights_client):
+    """
+    Test http_timeout config option
+    Set http_timeout to a very low value, run --test-connection and check
+    the set time
+    """
+    assert conftest.loop_until(lambda: insights_client.is_registered)
```
ptoscano, line 29: ditto

integration-tests/test_connection.py
```diff
@@ -0,0 +1,105 @@
...
+def test_noauth_proxy_connection(insights_client, test_config):
+    """
+    Test no-auth proxy option with --test-connection option
+    """
+    url_test = "End Upload URL Connection Test: SUCCESS"
+    api_test = "End API URL Connection Test: SUCCESS"
+
+    insights_client.register()
+    assert conftest.loop_until(lambda: insights_client.is_registered)
+
```
ptoscano, line 48: similar to what Archana already pointed out, there is no need to register for this test to run

integration-tests/test_connection.py
```diff
@@ -0,0 +1,105 @@
...
+def test_auth_proxy_connection(insights_client, test_config):
+    """
+    Test auth proxy option with --test-connection
+    """
+    url_test = "End Upload URL Connection Test: SUCCESS"
+    api_test = "End API URL Connection Test: SUCCESS"
+
+    insights_client.register()
+    assert conftest.loop_until(lambda: insights_client.is_registered)
```
ptoscano, line 71: ditto

integration-tests/test_connection.py
```diff
@@ -0,0 +1,105 @@
...
+
+def test_url_connection(insights_client):
+    """
+    Test wrong url option with --test-connection
+    """
+    url_test = "End Upload URL Connection Test: SUCCESS"
+    api_test = "End API URL Connection Test: SUCCESS"
+
+    insights_client.register()
+    assert conftest.loop_until(lambda: insights_client.is_registered)
```
ptoscano, line 98: ditto

integration-tests/test_connection.py
```diff
@@ -0,0 +1,105 @@
...
+        + str(test_config.get("auth_proxy", "port"))
+    )
+    insights_client.config.proxy = auth_proxy
+    insights_client.config.save()
+    test_connection = insights_client.run("--test-connection")
+    assert url_test in test_connection.stdout
+    assert api_test in test_connection.stdout
+
+
+def test_url_connection(insights_client):
```
ptoscano, line 90: since this test is about a wrong `base_url`, please tweak the test name so it mentions that explicitly

integration-tests/test_connection.py
```diff
@@ -0,0 +1,105 @@
...
+    """
+    Test wrong url option with --test-connection
+    """
+    url_test = "End Upload URL Connection Test: SUCCESS"
+    api_test = "End API URL Connection Test: SUCCESS"
+
+    insights_client.register()
+    assert conftest.loop_until(lambda: insights_client.is_registered)
+
+    insights_client.config.base_url = "cert-api.access.redhat.com:442/r/insights"
```
ptoscano, line 100: this URL is too similar to existing URLs; I'd suggest to use something less potentially easy to mistake, eg `no-such-insights-url.example.com/something`

integration-tests/test_connection.py
```diff
@@ -0,0 +1,105 @@
+import pytest
+import conftest
+
+
+pytestmark = pytest.mark.usefixtures("register_subman")
```
ptoscano, line 5: none of the tests actually need to register with `insights-client`, however they need to use a valid connection to Insights, and thus the certificate-based authentication

since IMHO it can be not clear in the long term to know why these test register with sub-man even if no actual registration with `insights-client`, then I recommend to add a comment before this mark to explain the situation

ptoscano (changes requested):
good start, there are few things to remove/change -- thanks!

integration-tests/test_connection.py
```diff
@@ -0,0 +1,94 @@
...
+def test_wrong_url_connection(insights_client):
+    """
+    Test wrong url option with --test-connection
+    """
+    wrong_url = (
+        "Max retries exceeded with url: //no-such-insights-url.example.com/something"
+    )
+    failed_connectivity = "Connectivity test failed!"
+
+    insights_client.config.auto_config = False
```
ptoscano, line 88: if `auto_config` is disabled, then, you also need to
- explicitly set the `authmethod` to "CERT", as the default in insights-core is still "BASIC"
- disable the `auto_update`, otherwise insights-core will try to fetch a new insights-core (failing miserably)

integration-tests/test_connection.py
```diff
@@ -0,0 +1,94 @@
...
+    insights_client.config.base_url = "no-such-insights-url.example.com/something"
+    insights_client.config.save()
+
+    test_connection = insights_client.run("--test-connection")
```
ptoscano, line 92: the return code of `insights-client` seems to be 1 in this situation, so you need to disable the automatic `check`, and then check for the return code

integration-tests/test_connection.py
```diff
@@ -0,0 +1,94 @@
...
+    assert url_test in test_connection.stdout
+    assert api_test in test_connection.stdout
+
+
+def test_wrong_url_connection(insights_client):
+    """
+    Test wrong url option with --test-connection
+    """
+    wrong_url = (
+        "Max retries exceeded with url: //no-such-insights-url.example.com/something"
```
ptoscano, line 84: in output lines with this text I do not see the hostname being attempted, only the paths; for example:
```
HTTPSConnectionPool(host='no-such-insights-url.example.com', port=443): Max retries exceeded with url: /something/uploads/ (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x7f751c80f1c0>: Failed to establish a new connection: [Errno -2] Name or service not known'))
HTTPSConnectionPool(host='no-such-insights-url.example.com', port=443): Max retries exceeded with url: / (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x7f751c3e0190>: Failed to establish a new connection: [Errno -2] Name or service not known'))
HTTPSConnectionPool(host='no-such-insights-url.example.com', port=443): Max retries exceeded with url: /r (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x7f751c80f760>: Failed to establish a new connection: [Errno -2] Name or service not known'))
HTTPSConnectionPool(host='no-such-insights-url.example.com', port=443): Max retries exceeded with url: /r/insights (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x7f751c3e0ac0>: Failed to establish a new connection: [Errno -2] Name or service not known'))
HTTPSConnectionPool(host='no-such-insights-url.example.com', port=443): Max retries exceeded with url: /r/insights (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x7f751c3e0ac0>: Failed to establish a new connection: [Errno -2] Name or service not known'))
```
also, "Max retries exceeded with url" is a message printed by `requests`, and thus may change depending on its version; I'd rather check for strings printed by insights-core itself (which can be kept static) rather than relying on what a 3rd party module prints as default exception representation

ptoscano (changes requested):
still some things to fix in `test_wrong_url_connection`

ptoscano (approved):
thanks!

PR #214, opened 2024-04-17
feat(test): Migrating test_reg.py upstream

zpetrace:
Migrating test_reg.py upstream with slight changes to the original test

### DIFF
--- a/integration-tests/test_client.py
+++ b/integration-tests/test_client.py
@@ -115,3 +115,24 @@ def test_insights_details_file_exists(insights_client):
     insights_client.run("--check-results")
     # Verify that insights-details.json gets re generated
     assert os.path.isfile(output_file)
+
+
+@pytest.mark.usefixtures("register_subman")
+def test_insights_directory_files(insights_client):
+    """
+    Test that /var/lib/insights have content available
+    """
+    directory = "/var/lib/insights"
+    registered_contents = [
+        "last_stable.egg",
+        "last_stable.egg.asc",
+    ]
+
+    insights_client.register()
+
+    dir_content_registered = [
+        entry.name for entry in os.scandir(directory) if entry.is_file()
+    ]
+
+    for item in registered_contents:
+        assert item in dir_content_registered, f"File '{item}' not found in directory."


integration-tests/test_client.py
```diff
@@ -97,3 +97,28 @@ def test_verify_logrotate_feature(insights_client):
...
+    directory = "/var/lib/insights"
+    registered_contents = [
+        "last_stable.egg",
+        "last_stable.egg.asc",
+    ]
+
+    insights_client.register()
+    dir_content_registered = [
+        os.path.basename(file) for file in glob.glob(f"{directory}/*")
+    ]
```
ptoscano, line 116: I think using `os.scandir()` to collect the filenames would be a better way to list the files here.

integration-tests/test_client.py
```diff
@@ -97,3 +97,28 @@ def test_verify_logrotate_feature(insights_client):
...
+@pytest.mark.usefixtures("register_subman")
+def test_insights_directory_files(insights_client):
+    """
+    Test that /var/lib/insights have the right files available
+    """
+    directory = "/var/lib/insights"
+    registered_contents = [
+        "last_stable.egg",
+        "last_stable.egg.asc",
+    ]
```
ptoscano, line 111: Note that `newest.egg` (+ `.asc`) are also files that may show up in that directory.

Note²: it seems the "malware" application in insights-core stores `.json` cache files in that directory.

ptoscano (changes requested):
To be honest, I'm a bit lost on what exactly is the test about: it seems to be asserting on the internals of `insights-core`, and in that case I'm not sure it is a good idea.

zpetrace:
To answer the question  why this test, @ptoscano  - the original test_reg.py from the iqe test suite was designed to unregister, delete /var/lib/insights , re-register and see if the folder got created again but I know we don't usually want to delete directories during tests, right? - I remember that from one of your comments in former PRs
so what I was trying to do is to just see if the files that are expected to be there are there but if you'd prefer to have it the original way - deleting and checking if it gets created again I can definitely do that

integration-tests/test_client.py
zpetrace: solved

integration-tests/test_client.py
zpetrace: done

integration-tests/test_client.py
```diff
@@ -115,3 +115,24 @@ def test_insights_details_file_exists(insights_client):
...
+@pytest.mark.usefixtures("register_subman")
+def test_insights_directory_files(insights_client):
+    """
+    Test that /var/lib/insights have content available
+    """
+    directory = "/var/lib/insights"
+    registered_contents = [
+        "last_stable.egg",
+        "last_stable.egg.asc",
+    ]
```
m-horky, line 129: Should this test also consider files like
- `/etc/insights-client/.lastupload`
- `/etc/insights-client/.registered`
 and so on? Those are always present on registered system as well.

zpetrace: Yeah but they are not in this directory, this test should focus just on that IMHO

m-horky (approved):
LGTM

PR #215, opened 2024-04-17
Tests : Added test for --diagnosis option

Archana-PandeyM:
Tests : Added test for --diagnosis option


ptoscano:
From a recent-ish discussion we had internally (CCT-226), it seems that there is lack of clarity on what exactly `--diagnosis` is for or does, both with and without the parameter for it.

My recommendation here is to hold this until we know exactly what `--diagnosis` is.

m-horky (approved):

m-horky:
> it seems that there is lack of clarity on what exactly --diagnosis is for or does, both with and without the parameter for it.

The conversation under CCT-226 is inconclusive, that's true. However, I believe we should model the new integration tests based on current behavior, and if CCT-226 is resolved, this test will be changed as well. I value regression tests, and this is a perfect example of behavior that should be tracked by QE tests.

PR #216, opened 2024-04-18
Tests: Added test to verify that --validate does not make network calls

Archana-PandeyM:
This test verifies that --validate does not try to connect to any network service.


integration-tests/test_client_options.py
```diff
@@ -87,3 +88,21 @@ def test_support(insights_client):
...
+
+def test_client_validate_no_network_call(insights_client):
+    """Verify that '--validate' option will not try to connect to any network service.
+
+    Ref : https://bugzilla.redhat.com/show_bug.cgi?id=2009864
+    """
+
+    # '/etc/insights-client/tags.yaml' file exists with any content
+    tags_filename = "/etc/insights-client/tags.yaml"
+    subprocess.run(["touch", tags_filename])
```
ptoscano, line 101: note that this file must be removed at the end of the test; what you can do is wrap the code of the test in a `try ... finally:` block, and in `finally` remove the file: this way, errors are propagated, and the file is removed in any case

integration-tests/test_client_options.py
```diff
@@ -87,3 +88,21 @@ def test_support(insights_client):
...
+    validate_result = insights_client.run("--validate")
+
+    # validating tags.yaml is loaded and no metric data in output
+    assert (
+        "/etc/insights-client/tags.yaml loaded successfully" in validate_result.stdout
+    )
+    assert "metrics Metrics:" not in validate_result.stdout
```
ptoscano, line 108: checking that the removed message is not present seems a good starting point; however, we cannot be sure that something else is not actually doing any network communication

I'd suggest something else instead, based on what done in #213 (for `test_wrong_url_connection`)
- disable `auto_update` & `auto_config`
- set a non-existing `base_url`
- run `insights-client --validate`, and expect it to work

this way, if anything attempts any communication to the Insights server, `insights-client` will fail

ptoscano (changes requested):
thanks for the initial approach, there are couple of things to fix/improve

integration-tests/test_client_options.py
Archana-PandeyM: I agree mostly but this is running when insights-client is not registered, if anything that needs network communication during --validate should fail even if base_url is correct.

integration-tests/test_client_options.py
Archana-PandeyM: done

integration-tests/test_client_options.py
Archana-PandeyM: done

ptoscano (approved):
thanks!

PR #217, opened 2024-04-18
feat(test): Migrating test_check_results.py upstream

zpetrace:
Migrating test_check_results.py upstream

### DIFF
--- a/integration-tests/test_client_options.py
+++ b/integration-tests/test_client_options.py
@@ -145,3 +145,35 @@ def test_client_diagnosis(insights_client):
     diagnosis_data = json.loads(diagnosis_result.stdout)
     # verify that diagnosis contains correct machine id
     assert diagnosis_data["insights_id"] == machine_id
+
+
+def test_check_show_results(insights_client):
+    """
+    Verify a remediation is returned when checking results from server (--check-results)
+    and then displaying then (--show-results)
+    First I will introduce a vulnerability to the system and at the
+    end of the test I will set the correct file permission again
+    """
+    insights_client.register()
+    assert conftest.loop_until(lambda: insights_client.is_registered)
+
+    try:
+        os.chmod("/etc/ssh/sshd_config", 0o777)
+
+        insights_client.run()
+
+        insights_client.run("--check-results")
+        show_results = insights_client.run("--show-results")
+
+        assert (
+            "hardening_ssh_config_perms|OPENSSH_HARDENING_CONFIG_PERMS"
+            in show_results.stdout
+        )
+        assert "Decreased security: OpenSSH config permissions" in show_results.stdout
+        assert (
+            "examine the following detected issues in OpenSSH settings:"
+            in show_results.stdout
+        )
+        assert "OPENSSH_HARDENING_CONFIG_PERMS" in show_results.stdout
+    finally:
+        os.chmod("/etc/ssh/sshd_config", 0o600)


integration-tests/test_client_options.py
```diff
@@ -87,3 +88,26 @@ def test_support(insights_client):
     assert "Running command:" in support_result.stdout
     assert "Process output:" in support_result.stdout
     assert "Support information collected in" in support_result.stdout
+
+
+def test_check_show_results(insights_client):
+    """
+    Verify a remediation is returned when checking results from server (--check-results)
+    and then displaying then (--show-results)
+    First I will intorudce a vulnerability to the system and at the
```
ptoscano, line 97: typo:
```suggestion
    First I will introduce a vulnerability to the system and at the
```

integration-tests/test_client_options.py
```diff
@@ -87,3 +88,26 @@ def test_support(insights_client):
...
+    end of the test I will set the correct chnod again
```
ptoscano, line 98: what does "chnod" mean?

integration-tests/test_client_options.py
```diff
@@ -87,3 +88,26 @@ def test_support(insights_client):
...
+    """
+    insights_client.register()
+    assert conftest.loop_until(lambda: insights_client.is_registered)
+
+    os.chmod("/etc/ssh/sshd_config", 0o777)
```
ptoscano, line 103: Not sure what it is doing, however I don't see anything related to it in the output of `--check-results` when I run it locally on my own RHEL 9 VM

ptoscano, line 103: note that the changes to this file must be revert at the end of the test; what you can do is wrap the code of the test in a `try ... finally:` block, and in `finally` revert the file changes: this way, errors are propagated, and the file is restored in any case

ptoscano (changes requested):
thanks for the initial version! please also rebase the branch according to the recent changes to avoid the conflicts

integration-tests/test_client_options.py
Archana-PandeyM: typo 'chmod'. May be you could write 'file permission' to make it more readable.

integration-tests/test_client_options.py
zpetrace: The change to the file was reverted with the `os.chmod("/etc/ssh/sshd_config", 0o600)` where I set the file permission back to the state it should be, although I edited it to be in a try - finally block.

zpetrace: we are introducing a vulnerability to the system to make sure that there will be an output using the --show-results - it doesn't really matter what vulnerability is being introduced  if the vulnerability is reversed at the end / which it does:)

zpetrace:
I have made the changes and rebased the branch so it has no conflicts.

integration-tests/test_client_options.py
```diff
@@ -118,3 +118,27 @@ def test_client_validate_no_network_call(insights_client):
...
+    insights_client.register()
+    assert conftest.loop_until(lambda: insights_client.is_registered)
+
+    try:
+        os.chmod("/etc/ssh/sshd_config", 0o777)
+
+        insights_client.run("--check-results")
+        show_results = insights_client.run("--show-results")
+
+        assert "rule" in show_results.stdout
```
Archana-PandeyM, line 139: Instead of directly looking for text in stdout, I would suggest to look in the json for specific "Problem" related to sshd_config permissions that you introduced. Reason for doing so is to ensure that desired issue is being reported in --show-results(ofcourse there could be multiple advisory but we want to look for the issue  we deliberately introduced):

suggestion:
look for "Update File Permission" , "OPENSSH_HARDENING_CONFIG_PERMS" etc.

zpetrace: done

integration-tests/test_client_options.py
```diff
@@ -145,3 +145,36 @@ def test_client_diagnosis(insights_client):
...
+    and then displaying then (--show-results)
+    First I will introduce a vulnerability to the system and at the
+    end of the test I will set the correct file permission again
+    """
+    insights_client.register()
+    assert conftest.loop_until(lambda: insights_client.is_registered)
+
+    try:
+        os.chmod("/etc/ssh/sshd_config", 0o777)
+
```
Archana-PandeyM, line 162: So when you introduce vulnerability in system we also need to make inventory aware about this issue. We can do so by running insights-client to upload the fresh data and then we run --check-results to verify that correct data is uploded.

Archana-PandeyM, line 162: So when you introduce vulnerability in system we also need to make inventory aware about this issue. We can do so by running insights-client to upload the fresh data and then we run --check-results to verify that correct data is uploded.

This test is currently failing when run. Please fix .

Archana-PandeyM (changes requested):

integration-tests/test_client_options.py
zpetrace: done

integration-tests/test_client_options.py
```diff
@@ -145,3 +145,38 @@ def test_client_diagnosis(insights_client):
...
+        insights_client.run("--check-results")
+        show_results = insights_client.run("--show-results")
+
+        assert (
+            "hardening_ssh_config_perms|OPENSSH_HARDENING_CONFIG_PERMS"
+            in show_results.stdout
+        )
+        assert "Decreased security: OpenSSH config permissions" in show_results.stdout
+        assert (
+            "security practices for configuring OpenSSH server are not being followed"
```
Archana-PandeyM, line 174: you can remove this assertion, its a generic recommendation.

Archana-PandeyM: suggested  verification to add-

error_key": "OPENSSH_HARDENING_CONFIG_PERMS"

"resolution_risk": { "name": "Update File Permission"

"rule_id": "hardening_ssh_config_perms|OPENSSH_HARDENING_CONFIG_PERMS",

"impact": { "name": "Decreased Security"

Archana-PandeyM:
LGTM

m-horky (approved):
LGTM

PR #218, opened 2024-04-19
Tests : added test case for --checkin with --offline client options

Archana-PandeyM:
This test verifies that running --checkin with --offline, a friendly message is logged, and it exists with a failure code


m-horky (approved):
LGTM

PR #219, opened 2024-04-19
Test : Added test for ultra-light checkin, --checkin operation

Archana-PandeyM:
when performing --checkin operation, it updates the "stale_timestamp" and "updated" host values.


m-horky (approved):
LGTM

PR #220, opened 2024-04-22
fix: updated expected error message when unregistering on already unregist…

Archana-PandeyM:
…ered system

This test is failing because the expected message differs in legacy mode and now we no more test on legacy mode, so updating the expected error message.


m-horky (approved):
LGTM

PR #222, opened 2024-04-22
test : Added test for --payload option to upload archives

Archana-PandeyM:
1- test to verify that a pre-collected insights-archive can be uploaded 
2- test to verify wrong content-type
3- test to verify upload of a very large file

### DIFF
--- a/integration-tests/test_upload.py
+++ b/integration-tests/test_upload.py
@@ -1,0 +1,78 @@
+import pytest
+import conftest
+
+pytestmark = pytest.mark.usefixtures("register_subman")
+
+
+def test_upload_pre_collected_archive(insights_client, tmp_path):
+    """This test verifies that a pre-collect insights-archive can be uploaded
+    using --payload operation.
+    """
+
+    archive_name = "archive.tar.gz"
+    archive_location = tmp_path / archive_name
+
+    # Registering the client because upload can happen on registered system
+    insights_client.register()
+    assert conftest.loop_until(lambda: insights_client.is_registered)
+
+    # Running insights-client in offline mode to generate archive and save at tmp dir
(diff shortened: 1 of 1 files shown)


integration-tests/test_upload.py
```diff
@@ -0,0 +1,78 @@
...
+    insights_client.register()
+    assert conftest.loop_until(lambda: insights_client.is_registered)
+
+    # Running insights-client in offline mode to generate archive and save at tmp dir
+    insights_client.run(f"--output-file={archive_location}")
+
+    # Running insights-client --payload with --content-type to upload archive
+    # collected in previous step
+    upload_result = insights_client.run(
+        f"--payload={archive_location}", "--content-type=gz"
```
m-horky, line 25: Maybe we should use the proper Advisor content type, as in
https://github.com/RedHatInsights/insights-core/blob/7069995dfc83783666bd967559b3f5810ffb8149/insights/client/connection.py#L372

```py
"--content-type=application/vnd.redhat.advisor.collection+tgz"
```

if we want to ensure the archive will be treated as such.

m-horky (commented):
No major notes, looks good.

integration-tests/test_upload.py
Archana-PandeyM: I understand your point here, but the test are automated in a way customer would use the tool. I found in Insights-client user guides :
--content-type=TYPE
Set the content-type for the PAYLOAD file. Type can be gz, bz2, xz, and none. The TYPE must match the --compressor used with the PAYLOAD.

Also using the types gz, xz and bz2 processes the archive successfully. Let me know if I am missing something here.

m-horky (approved):
LGTM

PR #224, opened 2024-04-24
test: Added test for --compressor

Archana-PandeyM:
This test verifies that valid compressor types(gz,bz2,xzwq) can be used to generate archive and upload using --payload option.

### DIFF
--- a/integration-tests/test_upload.py
+++ b/integration-tests/test_upload.py
@@ -1,3 +1,5 @@
+import os
+
 import pytest
 import conftest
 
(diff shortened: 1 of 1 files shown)


integration-tests/test_upload.py
```diff
@@ -76,3 +78,38 @@ def test_upload_too_large_archive(insights_client, tmp_path):
...
     assert "Archive is too large to upload" in upload_result.stdout
     assert "Upload failed." in upload_result.stdout
+
+
+@pytest.mark.parametrize(
+    "compressor,expected_extension",
+    [
+        ("gz", ".gz"),
+        ("bz2", ".bz2"),
+        ("xz", ".xz"),
```
m-horky, line 88: Would it make sense to expect `.tar.gz` instead of just `.gz`?

Archana-PandeyM: the archives that get created are compressed tar file. for example archive.tar.gz is gzip compreseed tar file and so the extension of this file is gz and not tar.gz.
But I think to validate that basename of file i.e archive.tar I can add more assertion.

PR #225, opened 2024-04-24
feat(test): Migrating test_common_specs.py upstream

zpetrace:
Migrating test_common_specs.py upstream

### DIFF
--- a/integration-tests/test_common_specs.py
+++ b/integration-tests/test_common_specs.py
@@ -1,0 +1,40 @@
+import pytest
+import re
+import subprocess
+
+pytestmark = pytest.mark.usefixtures("register_subman")
+
+
+def test_common_specs(insights_client):
+    common_specs = [
+        "insights.specs.Specs.date",
+        "insights.specs.Specs.dmidecode",
+        "insights.specs.Specs.fstab",
+        "insights.specs.Specs.hostname",
+        "insights.specs.Specs.hosts",
+        "insights.specs.Specs.installed_rpms",
+        "insights.specs.Specs.ip_addresses",
+        "insights.specs.Specs.ls_dev",
+        "insights.specs.Specs.lscpu",
+        "insights.specs.Specs.lspci",
+        "insights.specs.Specs.meminfo",
+        "insights.specs.Specs.mount",
+        "insights.specs.Specs.mountinfo",
+        "insights.specs.Specs.ps_auxcww",
+        "insights.specs.Specs.redhat_release",
+        "insights.specs.Specs.uname",
+        "insights.specs.Specs.yum_repos_d",
+    ]
+
+    insights_client.register()
+    output = insights_client.run("--no-upload")
+
+    pattern = r"Archive saved at (.+/[^/]+\.tar\.gz)$"
+    match = re.search(pattern, output.stdout, re.MULTILINE)
+    archive_path = match.group(1)
+
+    content = subprocess.run(
+        "tar", "-xvzf", archive_path, text=True, capture_output=True
+    )
+    for spec in common_specs:
+        assert spec in content.stdout


zpetrace:
PR number #217 has to be merged first due to merge conflicts

integration-tests/test_client_options.py
```diff
@@ -177,3 +179,43 @@ def test_check_show_results(insights_client):
...
+    ]
+
+    insights_client.register()
+    output = insights_client.run("--no-upload")
+
+    pattern = r"Archive saved at (.+/[^/]+\.tar\.gz)$"
+    match = re.search(pattern, output.stdout, re.MULTILINE)
+    if match:
+        archive_path = match.group(1)
+        return archive_path
```
m-horky, line 212: What are you trying to achieve here? Should it be `assert` instead? Pytest is complaining about it:

> `/usr/local/lib/python3.9/site-packages/_pytest/python.py:199: PytestReturnNotNoneWarning: Expected None, but integration-tests/test_client_options.py::test_common_specs` returned `'/var/cache/insights-client/insights-insights-client-integration-tests-20240425115452.tar.gz'`, which will be an error in a future version of pytest.  Did you mean to use `assert` instead of `return`?
```

Comment (zpetrace) [reply integration-tests/test_client_options.py:212]: yeah there was no need for returning the value, my bad there, solved

File: integration-tests/test_client_options.py [hunk 2]
```diff
@@ -177,3 +179,40 @@ def test_check_show_results(insights_client):
...
+        "insights.specs.Specs.yum_repos_d",
+    ]
+
+    insights_client.register()
+    output = insights_client.run("--no-upload")
+
+    pattern = r"Archive saved at (.+/[^/]+\.tar\.gz)$"
+    match = re.search(pattern, output.stdout, re.MULTILINE)
+    if match:
+        archive_path = match.group(1)
```
m-horky, line 211: What should happen if `match` is False-y? Currently you'd get an error two lines below that `archive_path` is not defined. I think we can assume the `match` will contain the data and drop the if;
```py
match = re.search(pattern, output.stdout, re.MULTILINE)
archive_path = match.group(1)
```

zpetrace: done

integration-tests/test_client_options.py
```diff
@@ -177,3 +179,39 @@ def test_check_show_results(insights_client):
...
+    ]
+
+    insights_client.register()
+    output = insights_client.run("--no-upload")
+
+    pattern = r"Archive saved at (.+/[^/]+\.tar\.gz)$"
+    match = re.search(pattern, output.stdout, re.MULTILINE)
+    archive_path = match.group(1)
+
+    subprocess.run(["tar", "-xvzf", archive_path, ">", "/var/tmp/archive.txt"])
```
m-horky, line 212: You cannot pipe content in subprocesses like that. There is an easier way to do this:

```py
content = subprocess.run(["tar", "-xvzf", archive_path], text=True, capture_output=True)
for spec in common_specs:
    assert spec in content.stdout
```

zpetrace: done, thanks for help!

integration-tests/test_client_options.py
```diff
@@ -177,3 +179,38 @@ def test_check_show_results(insights_client):
         assert "OPENSSH_HARDENING_CONFIG_PERMS" in show_results.stdout
     finally:
         os.chmod("/etc/ssh/sshd_config", 0o600)
+
+
+def test_common_specs(insights_client):
```
Archana-PandeyM, line 184: We need to be careful with grouping of tests. This test should not be in test_client_options.py because its not validating any insights-client options, please move it to a new file test_common_specs.py.

Archana-PandeyM (changes requested):

integration-tests/test_client_options.py
zpetrace: done

m-horky:
@zpetrace Please squash those two commits into one. We don't need that original commit in git history; otherwise it is ready to be merged.

zpetrace:
squashed

Archana-PandeyM (approved):

PR #226, opened 2024-04-25
test: added test for --retry option during failed uploads

Archana-PandeyM:
Adding two tests to verify that retry is attempted when upload fails and not attempted when failure is due to unrecoverable reasons.


m-horky (approved):
LGTM

PR #227, opened 2024-04-25
test : added a test to validate that insights-details.json gets recre…

Archana-PandeyM:
…ated on calling --check-results


m-horky (approved):
LGTM

PR #228, opened 2024-04-26
test : Migrating test_file_workflow.py

Archana-PandeyM:
Adding test to verify file workflow when uploading archive without cannonical facts and with only one cannoical fact


xiangce (approved):
@Archana-PandeyM  - these 2 tests look good to me

PR #232, opened 2024-05-02
refactor: how to manage shared variables (aka HOST_DETAILS)

jstavel:
that variables will be used in other tests. 
So I've moved the var into shared file.

### DIFF
--- a/integration-tests/test_checkin.py
+++ b/integration-tests/test_checkin.py
@@ -2,13 +2,12 @@ import contextlib
 import json
 
 import pytest
+from constants import HOST_DETAILS
 
 pytestmark = pytest.mark.usefixtures("register_subman")
 
-HOST_DETAILS = "/var/lib/insights/host-details.json"
 
-
-def test_ultralight_checkin(insights_client):
+def test_ultralight_checkin(insights_client, test_config):
     """test --checkin
     Sends nothing but canonical facts to update the host stale and updated timestamp
     """
--- a/integration-tests/constants.py
+++ b/integration-tests/constants.py
@@ -1,0 +1,1 @@
+HOST_DETAILS: str = "/var/lib/insights/host-details.json"


integration-tests/conftest.py
```diff
@@ -2,6 +2,8 @@
 import subprocess
 import time
 
+HOST_DETAILS: str = "/var/lib/insights/host-details.json"
```
subpop, line 5: Should this use a hard-coded prefix? Or should the build tool construct this path from its prefix and [other default build directories](https://mesonbuild.com/Builtin-options.html#directories)? I would imagine that because we're testing software that's immediately been built, we could incorporate these files into the build system too?

jstavel: do you mean something like this:   HOST_DETAILS = os.path.join(INSIGHTS_LIB_DIR, 'host-details.json')?

jstavel: I can imagine this scenario:
builder of the package creates properties.toml file with varibles aka: build.lib, buils.sysconfig, ....

testware (uses DYNACONF to load properties) loads the properties.
testware offers the properties to tests to create HOST_DETAILS, ... properties.

is it the case you think @subpop ?

jstavel:
I have solved this situation by moving the variable from python file into dynaconf setting file. even the var is static - it is a matter of some configuration processes - I mean it can be changed by build process. so the build process can change .secrets.toml by new value.

Archana-PandeyM:
In my opinion this var should either go to constants.py or let it be the way it is right now. this is not a configuration option that could go in config file.

integration-tests/test_checkin.py
```diff
@@ -5,17 +5,16 @@
...
+def test_ultralight_checkin(insights_client, test_config):
     """test --checkin
     Sends nothing but canonical facts to update the host stale and updated timestamp
     """
     insights_client.register()
-    # Performing check-results operation provides latest host data in host-details.json
+    # Performing check-results operation provi des latest host data in host-details.json
     insights_client.run("--check-results")
-    with open(HOST_DETAILS, "r") as data_file:
+    host_details = test_config.get("insights", "host_details")
```
Archana-PandeyM, line 16: In my opinion this var should either go to constants.py or let it be the way it is right now. this is not a configuration option that could go in config file.

Archana-PandeyM (changes requested):

integration-tests/test_checkin.py
jstavel: updated to your request.

Archana-PandeyM:
Looks good! Please squash the commits and change commit message to reflect the work.

integration-tests/test_checkin.py
```diff
@@ -2,18 +2,17 @@
...
 
-
-def test_ultralight_checkin(insights_client):
+def test_ultralight_checkin(insights_client, test_config):
     """test --checkin
     Sends nothing but canonical facts to update the host stale and updated timestamp
     """
     insights_client.register()
-    # Performing check-results operation provides latest host data in host-details.json
+    # Performing check-results operation provi des latest host data in host-details.json
```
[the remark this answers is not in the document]
jstavel: fixed.

PR #234, opened 2024-05-07
feat(test): Migrating tests from test_collection.py upstream

zpetrace:
migrating test_cmd_timeout, test_archive_structure and test_branch_info

### DIFF
--- a/integration-tests/test_collection.py
+++ b/integration-tests/test_collection.py
@@ -1,6 +1,11 @@
 import contextlib
 import os
 import tarfile
+import glob
+import json
+import pytest
+
+ARCHIVE_CACHE_DIRECTORY = "/var/cache/insights-client"
 
 
 def test_output_file_valid_parameters(insights_client, tmp_path):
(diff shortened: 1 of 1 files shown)


Lorquas:

integration-tests/test_collection.py
```diff
@@ -106,3 +111,101 @@ def test_output_file_already_exists(insights_client, tmp_path):
...
+        "proc",
+        "run",
+        "sys",
+        "usr",
+        "var",
+    ]
+
+    insights_client.register()
+
+    insights_client.config.core_collect = True
```
Archana-PandeyM, line 193: ditto

integration-tests/test_collection.py
```diff
@@ -106,3 +111,101 @@ def test_output_file_already_exists(insights_client, tmp_path):
...
+    Test cmd_timeout config option
+    Client executes all commands with timeout. If the command doesn't finish within
+    a specified amount of seconds, it is killed.
+    Test that setting a custom amount of seconds to cmd_timeout config option works.
+    """
+    cmd_output_message = "Executing: [['timeout', '-s', '9', '10'"
+
+    insights_client.register()
+
+    insights_client.config.core_collect = True
```
Archana-PandeyM, line 128: Hi, we need not specifically set core_collect to true because for versions of the Insights Client RPM 3.1.0 and newer, core collection is enabled by default. So please remove this step.

integration-tests/test_collection.py
```diff
@@ -106,3 +111,101 @@ def test_output_file_already_exists(insights_client, tmp_path):
...
+@pytest.mark.usefixtures("register_subman")
+def test_branch_info(insights_client):
+    """
+    Test that branch_info includes all required information
+    branch_info is used to identify the host when it is connectec via Satellite.
+    It uses branch_id and leaf_id to identify
+    """
+    insights_client.register()
+
+    insights_client.config.core_collect = True
```
Archana-PandeyM, line 145: ditto.

Archana-PandeyM (changes requested):

integration-tests/test_collection.py
zpetrace: oh okay, understood, done

integration-tests/test_collection.py
```diff
@@ -106,3 +111,94 @@ def test_output_file_already_exists(insights_client, tmp_path):
...
+        directory_name = latest_file.replace(".tar.gz", "")
+
+    extracted_dirs_files = os.listdir(directory_name)
+    missing_dirs = [d for d in dirs_list if d not in extracted_dirs_files]
+    assert not missing_dirs, f"Missing directories {missing_dirs}"
+
+    data_dir_path = os.path.join(directory_name, "data")
+    data_subdirs = os.listdir(data_dir_path)
+    missing_subdirs = [d for d in subdirs_list if d not in data_subdirs]
+    assert not missing_subdirs, f"Missing subdirectory {missing_subdirs}"
```
m-horky, line 204: There is nothing wrong with these tests, but I wonder whether it is the client's job to test what was collected. It feels like it should not be our responsibility to inspect the content of the archive.

If this is meant as a sanity check, I have no objections.

This is non-blocking question.

m-horky (approved):

Archana-PandeyM (approved):

PR #235, opened 2024-05-07
feat: save test results as junit

Lorquas:
This will save the test results as a junit.  This will be needed for downstream reporting (datarouter/polarion/jenkins etc...).  Could also possibly be used for fancier reporting here?

This will be a draft PR while I work on the downstream components.


Lorquas:
test this please

PR #237, opened 2024-05-13
fix(test):added assertions to verify registered state of system

Archana-PandeyM:
This change will help mitigate syncup issues with
system state after registration.
Some tests intermitently fail due to slow sync.


Archana-PandeyM:
test this please

jstavel (approved):
thank you. It will help with out tmt downstream tess.

PR #238, opened 2024-05-13
Build 'master' branch in COPR with Packit

subpop:
Add a job that triggers a COPR build in the @yggdrasil/lastest project when a commit is pushed to the master branch.


m-horky (approved):
LGTM

PR #239, opened 2024-05-15
chore(ci): Add CentOS Stream 10 to pytest matrix

m-horky:
As of now, `fedora:latest` more or less matches c10s/el10, but they will start to diverge, and in six months they will contain different software. This is proactive measure to start testing on RHEL 10-like system.


Lorquas:

ptoscano:
Side note: I think it would be good to add `.github/workflows/pytest.yml` itself to the `paths` of the "pytest" workflow, so the unit tests are triggered whenever the CI bits for them are changed; as alternative option, drop the path limitation altogether and let the unit test always run in any PR.

ptoscano (approved):
This affects only the unit tests in GHA -> merging.

PR #247, opened 2024-05-31
feat: improves motd registration instructions

DuckBoss:
This patch aims to improve the registration
instructions provided to users in the insights motd. 
The new messages focus on directing the user
to registering with 'rhc' and related KAs.

Resolves: CCT-264


m-horky (approved):
LGTM

PR #249, opened 2024-06-04
chore(ci): Use Vault for CentOS Stream 8

m-horky:
CentOS Stream 8 reached EOL on 2024-05-31. This patch ensures we can still run EL8-equivalent tests on frozen version of Stream 8.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -30,6 +30,11 @@ jobs:
       - name: "Checkout the repository"
         uses: actions/checkout@v4
 
+      - name: "Use CentOS Vault"
+        if: matrix.name == 'CentOS Stream 8'
+        run: |
+          sed -i 's|#baseurl=http://mirror|baseurl=http://vault|' /etc/yum.repos.d/CentOS-Stream-*.repo
+
       - name: "Install dependencies"
         run: |
           dnf --setopt install_weak_deps=False install -y git-core python3-pip gpg


subpop:
Interesting approach you took. It is my understanding that the contents of vault will not change, so even while Red Hat continues support for RHEL 8, those bug fixes will not be made available in a CentOS Stream 8 compose. This means that over time, the drift in what's committed upstream to the CentOS GitLab repositories and what RPMs are accessible on vault.centos.org will increase. Is it really worth testing against a static version of CentOS. Eventually, some incompatibility will cause the test to fail, and because vault is static, our only option then will be to drop CentOS Stream 8, so why not drop it now?

We should continue testing on RHEL8 however, as those composes are constructed from the upstream CentOS GitLab commits.

m-horky:
Well, there is a drift and there is a "drift". I don't assume RHEL 8.10 will start differing from this frozen c8s in a significant way, that would go against the stability principle. No packages will be yanked, no major versions released.

I think that even in a frozen state this is a really good proxy for RHEL, and for running our unit tests it should be well enough. We still have the downstream RHEL tests to verify everything is OK.

The only actual problem that may occur here is pip packages dropping support for Python 3.6, I believe we've seen this in subscription-manager repository somewhere. In that case we have to freeze all dependencies -- but that is still quite easy to do.

subpop:
Fair enough.

ptoscano (approved):
This affects only the unit tests in GHA -> merging.

PR #250, opened 2024-06-05
add display_name case with auto_config=false and auto_update=false

zhangqianqian:


integration-tests/test_display_name_option.py
```diff
@@ -133,3 +133,59 @@ def test_invalid_display_name(invalid_display_name, insights_client):
     assert (
         origin_display_name == record["display_name"]
     ), "display-name should remain unchanged when new display-name is rejected"
+
+
+def test_display_name_disable_autoconfig_and_autoupdate(insights_client, test_config):
+    """
+    This is a satellite bug(https://issues.redhat.com/browse/ESSNTL-2230),
```
Archana-PandeyM, line 140: (nit) to maintain uniformity in all tests, we should replace it with something like : 
Ref :  https://issues.redhat.com/browse/ESSNTL-2230,

integration-tests/test_display_name_option.py
```diff
@@ -133,3 +133,59 @@ def test_invalid_display_name(invalid_display_name, insights_client):
...
+    determine test this case for both satellite and production
+    environment after discussion with Archana.
```
Archana-PandeyM, line 142: Also you can remove names from docstring. and details can be added in the Jira card that is tracking this automation task.

integration-tests/test_display_name_option.py
```diff
@@ -133,3 +133,59 @@ def test_invalid_display_name(invalid_display_name, insights_client):
...
+        2) Host visible in the c.r.c.
+    """
+    # configure auto_config and auto_update as False
+    if "satellite" in test_config.environment:
+        satellite_hostname = test_config.get("candlepin", "host")
+        satellite_port = test_config.get("candlepin", "port")
+        insights_client.config.base_url = (
+            satellite_hostname + ":" + str(satellite_port) + "/redhat_access/r/insights"
+        )
+    elif "prod" in test_config.environment:
```
Archana-PandeyM, line 172: it could be 'stage' or  'prod' or something else. lets make it 'else'

Archana-PandeyM (changes requested):

integration-tests/test_display_name_option.py
zhangqianqian: @Archana-PandeyM you mean name this case as "Hilti config + Satellite"? But @ptoscano suggested "please NEVER mention customer names in public stuff", and that's why I choose current test case name.

Archana-PandeyM: by name I meant my name, I see already removed now. Thanks!
No I would also suggest not to put any customer specific data in upstream tests. You might want to edit the comment and remove names form above comment as well.
What I am suggesting is to replace line 140  with "Ref : https://issues.redhat.com/browse/ESSNTL-2230"

Archana-PandeyM: LGTM, please sqash commit messages into one. Rebase branch.

integration-tests/test_display_name_option.py
```diff
@@ -139,3 +139,62 @@ def test_invalid_display_name(invalid_display_name, insights_client):
...
+    insights_client.config.cert_verify = True
+    insights_client.config.authmethod = "CERT"
+    unique_hostname = generate_unique_hostname()
+    insights_client.config.display_name = unique_hostname
+    insights_client.config.save()
+
+    # register insights
+    status = insights_client.run("--register")
+    assert unique_hostname in status.stdout
+
```
Archana-PandeyM, line 194: intermittently test fails on command next to register command.
Please add the assertion to check if system is fully registered like, "assert conftest.loop_until(lambda: insights_client.is_registered)"

zhangqianqian:
something wrong with pushing, close this PR and will submit another one.

PR #251, opened 2024-06-07
test: Added test to trigger insights-client via insights-client.timer

Archana-PandeyM:
Added a test case to run data collection and upload via systemd timer

### DIFF
--- a/integration-tests/test_client_systemd.py
+++ b/integration-tests/test_client_systemd.py
@@ -1,0 +1,60 @@
+import conftest
+from time import sleep
+import pytest
+import subprocess
+from pathlib import Path
+
+pytestmark = pytest.mark.usefixtures("register_subman")
+
+
+def test_data_upload_systemd_timer(insights_client, external_inventory):
+    """
+    This test aims to verify that insight-client uploads data via systemd timer
+    steps:
+        1- Register the system
+        2- Note the last_check_in in host details from inventory
+        3- Edit insights-client timer to run every 3 minutes (24 hours is
+        too long to wait for a test, 3 min appears be a decent wait)
+        4- Wait for upload to finish
+        5- Again note the last_check_in in host details from inventory
+        6- Validate step 5 last_check_in > step 2 last_check_in
+
+    """
+
+    insights_client.register()
+    assert conftest.loop_until(lambda: insights_client.is_registered)
+
+    machine_id = insights_client.uuid
+    # fetching host details to note last_check_in time
+    response = external_inventory.get(path=f"hosts?insights_id={machine_id}")
+    host_staleness = response.json()["results"][0]["per_reporter_staleness"]
+    last_checkin_time = host_staleness["puptoo"]["last_check_in"]
+
+    # override insights-client.timer to run every 3 min
+    override_path = Path("/etc/systemd/system/insights-client.timer.d/override.conf")
+    override_content = """
+        [Timer]
+        OnCalendar=*:0/3
+        Persistent=true
+        RandomizedDelaySec=5
+        """
+    try:
+        override_path.parent.mkdir()
+        override_path.write_text(override_content)
+
+        subprocess.run(["systemctl", "daemon-reload"])
+
+        sleep(60 * 5)  # sleep for the period when timer is triggered and upload happens
+
+        # remove the override.conf for insights-client.timer to avoid multiple uploads
+        subprocess.run(["systemctl", "revert", "insights-client.timer"])
+
+        # fetch the last_check_in time
+        response = external_inventory.this_system()
+        latest_checkin_time = response["per_reporter_staleness"]["puptoo"][
+            "last_check_in"
+        ]
+        assert last_checkin_time < latest_checkin_time
+
+    finally:
+        subprocess.run(["systemctl", "revert", "insights-client.timer"])


integration-tests/test_client_systemd.py
```diff
@@ -0,0 +1,62 @@
...
+
+    # remove the override.conf for insights-client.timer to avoid multiple uploads
+    subprocess.run(["systemctl", "revert", "insights-client.timer"])
+
+    # fetch the last_check_in time
+    machine_id = insights_client.uuid
+    response = external_inventory.get(path=f"hosts?insights_id={machine_id}")
+    latest_checkin_time = response.json()["results"][0]["per_reporter_staleness"][
+        "puptoo"
+    ]["last_check_in"]
```
m-horky, line 60: What's the likelihood the Inventory won't have the new data yet? Should we worry about it, or does all test suite assume the HBI record is updated basically immediately?

integration-tests/test_client_systemd.py
```diff
@@ -0,0 +1,62 @@
...
+        OnCalendar=*:0/3
+        Persistent=true
+        RandomizedDelaySec=5
+        """
+    os.makedirs(os.path.dirname(override_path), exist_ok=True)
+    with open(override_path, "w") as override_file:
+        override_file.write(override_content)
+
+    subprocess.run(["systemctl", "daemon-reload"])
+    # subprocess.run(['systemctl', 'restart', 'insights-client.timer'])
```
m-horky, line 48: Is it commented out for a reason?

m-horky (commented):
The test scenario looks good. I just have a few questions.

integration-tests/test_client_systemd.py
```diff
@@ -0,0 +1,62 @@
...
+
+    sleep(300)  # sleep for the period when timer is triggered and upload happens
+
+    # remove the override.conf for insights-client.timer to avoid multiple uploads
+    subprocess.run(["systemctl", "revert", "insights-client.timer"])
```
ptoscano, line 53: while this is a good option, this will not be done (leaving the override in place for other tests) in case there is any failure between the creation of the file and this point; what I recommend is the `ExitStack` context manager [1]:
```python
with contextlib.ExitStack() as stack:
    # create the override
    stack.callback(subprocess.run, ["systemctl", "revert", "insights-client.timer"])
    # continue with the test as needed, until you need to actually call the revert
```
this way it is executed no matter what, even in case of failures, and without using a custom `try: ... finally: ...` block

[1] https://docs.python.org/3/library/contextlib.html#contextlib.ExitStack

integration-tests/test_client_systemd.py
```diff
@@ -0,0 +1,62 @@
...
+        RandomizedDelaySec=5
+        """
+    os.makedirs(os.path.dirname(override_path), exist_ok=True)
+    with open(override_path, "w") as override_file:
+        override_file.write(override_content)
+
+    subprocess.run(["systemctl", "daemon-reload"])
+    # subprocess.run(['systemctl', 'restart', 'insights-client.timer'])
+
+    sleep(300)  # sleep for the period when timer is triggered and upload happens
```
ptoscano, line 50: readability note (at least for me): since this is about waiting for 5 minutes, i.e. a couple of minutes more than the new timer setup, I'd write this as multiplication with seconds:
```python
    sleep(60 * 5)  # sleep for ...
```

integration-tests/test_client_systemd.py
Archana-PandeyM: I would say likelihood of inventory not having new data is same as any new register/upload operation.
There is no fixed delay to expect so adding any delay in tests is just a wild guess. In this test I have kept a delay of 300secs.

integration-tests/test_client_systemd.py
Archana-PandeyM: I would need this cleanup to happen at this step only and not at the end of test, reason is that the restart timer has a low value and multiple triggers would happen before test finishes. So probably I could use try..fanally and rerun this step in finally as well?
your thoughts?

m-horky (approved):
LGTM

integration-tests/test_client_systemd.py
```diff
@@ -0,0 +1,61 @@
...
+
+        sleep(60 * 5)  # sleep for the period when timer is triggered and upload happens
+
+        # remove the override.conf for insights-client.timer to avoid multiple uploads
+        subprocess.run(["systemctl", "revert", "insights-client.timer"])
+
+        # fetch the last_check_in time
+        machine_id = insights_client.uuid
+        response = external_inventory.get(path=f"hosts?insights_id={machine_id}")
+        host_staleness = response.json()["results"][0]["per_reporter_staleness"]
```
ptoscano, line 55: this should be `external_inventory.this_system()`

integration-tests/test_client_systemd.py
ptoscano: since there is no exception handling, a contextmanager is a slightly better solution: bits pushed to the `ExitStack` are rolled back in the exact inverted order than they were pushed, simplifying the rollback a lot (no manual ordering needed)

in this case there is only one operation to do; still I think it'd be better to follow a more clean procedure, rather than (mis)using `try ... finally` blocks

m-horky (approved):

PR #252, opened 2024-06-10
fix(test): Removing use of fauxfactory from integration tests

Archana-PandeyM:
currently tests are failing to install fauxfactory on s390x and ppcle64. It is causing failure in downstream compose testing. Replacing the use of fauxfactory by using python inbuilt 'random' and 'string' modules.
Also reverting python3-cryptography that was added a package to support installation of fauxfactory. Python3-cryptography not required by any other tests


m-horky (approved):
LGTM

PR #253, opened 2024-06-11
test: Add integration tests for the Ansible playbook verifier

m-horky:
* Card ID: CCT-546

### DIFF
--- a/integration-tests/playbook_verifier/playbooks/README.md
+++ b/integration-tests/playbook_verifier/playbooks/README.md
@@ -1,0 +1,12 @@
+# Test data for the Ansible playbook verifier
+
+To verify the functionality of the verifier, these playbooks have been downloaded from git repository of config-manager.
+
+- [insights_setup.yml](https://github.com/RedHatInsights/config-manager/blob/master/playbooks/insights_setup.yml)
+- [compliance_openscap_setup.yml](https://github.com/RedHatInsights/config-manager/blob/master/playbooks/compliance_openscap_setup.yml)
+
+The list of playbooks should be extended to improve coverage:
+- lines ending with Windows-style line endings (`\r\n`),
+- comments with weird indentation,
+- playbook with comments stripped before verifying,
+- playbooks containing non-ASCII UTF-8 characters.
--- a/integration-tests/playbook_verifier/playbooks/compliance_openscap_setup.yml
+++ b/integration-tests/playbook_verifier/playbooks/compliance_openscap_setup.yml
@@ -1,0 +1,43 @@
+# This playbook will install the required OpenSCAP packages to get your systems ready to be used
+# with Insights Compliance service.
+#
+# After running this playbook, perform the following steps manually:
+# 1. Create a Compliance policy in cloud.redhat.com and associate the required systems.
+# 2. Run `insights-client --compliance` on your system to report Compliance to cloud.redhat.com
+
+- name: Compliance OpenSCAP Setup
+  hosts: localhost
+  become: yes
+  vars:
+    insights_signature_exclude: /hosts,/vars/insights_signature
+    insights_signature: !!binary |
+      TFMwdExTMUNSVWRKVGlCUVIxQWdVMGxIVGtGVVZWSkZMUzB0TFMwS1ZtVnljMmx2YmpvZ1IyNTFV
+      RWNnZGpFS0NtbFJTVlpCZDFWQldUbE1WREpqZG5jMU9FUXJhalZ3VGtGUmFIaDFRa0ZCWjA1cE9D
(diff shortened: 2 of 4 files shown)


ptoscano (commented):
The new test generally looks OK; that said, IMHO it seems it can be simplified a bit, removing extra "wrapping" and bits:
- the helper `verify_playbook()` seems only a mere wrapping around `subprocess.run()`; I'd simply fold it in `test_official_playbook()`, in case there is nothing more that needs to use that helper
- in `verify_playbook()`: some arguments are passed directly to `subprocess.run()` and some others are in a `extra_args` list; I'd fold `extra_args` directly in `subprocess.run()`
- in `verify_playbook()`: passing `check=True` to `subprocess.run()` avoids the need to `assert result.returncode == 0`, as a failing execution will raise CalledProcessError whose content can be inspected/logged
- in `verify_playbook()`: you don't need to use `shutil.which()`, `subprocess.run()` (or whatever it uses internally) will do that automatically; it will not make any difference anyway, since both ways (`subprocess.run()` and `shutil.which()`) use `$PATH` to locale the executable

m-horky:
> the helper `verify_playbook()` seems only a mere wrapping around `subprocess.run()`; I'd simply fold it in `test_official_playbook()`, in case there is nothing more that needs to use that helper

I had plans to add additional test(s) that would run it with non-Red Hat GPG key, but I had to reprioritize. Even if, don't you think this better separates test setup from the actual call? I can merge them if not.

> in `verify_playbook()`: some arguments are passed directly to `subprocess.run()` and some others are in a `extra_args` list; I'd fold `extra_args` directly in `subprocess.run()`

This was also to make it cleaner -- the extra arguments are not required for the functionality, they are just a result of how limited the CLI API is. Again, if it doesn't make sense, I can rewrite it.

> in `verify_playbook()`: passing `check=True` to `subprocess.run()` avoids the need to `assert result.returncode == 0`, as a failing execution will raise CalledProcessError whose content can be inspected/logged

That way you don't get to see the stderr of the command, pytest only reports that the subcommand exited with non-zero status. This makes it much easier to debug.

> in `verify_playbook()`: you don't need to use `shutil.which()`, `subprocess.run()` (or whatever it uses internally) will do that automatically; it will not make any difference anyway, since both ways (`subprocess.run()` and `shutil.which()`) use `$PATH` to locale the executable

Fair point. Will address this once we agreed on what to do with the rest of your notes.

ptoscano:
> Even if, don't you think this better separates test setup from the actual call?

If there is anything more planned, maybe; since so far the test needed seems "run the validation and check it works", I'd keep things simple for now. Refactoring is easy, in case there is a need later on.

> This was also to make it cleaner -- the extra arguments are not required for the functionality, they are just a result of how limited the CLI API is. Again, if it doesn't make sense, I can rewrite it.

More than "doesn't make sense", what I looked at is how this is used in other workers (so far only rhc-worker-playbook and rhc-worker-script), and they both call `insights-client` with the same set of parameters (the ones in the test). Hence, for me the verification call is made by all the parameters at the same time, with no separation in two "types".

> That way you don't get to see the stderr of the command, pytest only reports that the subcommand exited with non-zero status.

If stderr is collected, then the raisen CalledProcessError has stderr too, which will be printed when the exception is stringified.

m-horky:
> I'd keep things simple for now.
> the verification call is made by all the parameters at the same time

Ack.

> If stderr is collected, then the raisen CalledProcessError has stderr too, which will be printed when the exception is stringified.

That's not what I'm seeing. The returned `CalledProcessError` only displays the arguments and status.

ptoscano:
IMHO the newly added test generally work fine. I think the commit message and/or the test  would deserve a bit more verbosity on what's being added (and why), however it is not a hard stop from me.

On OSes with Python 3.12 the stderr is filled with warnings for code that is in insights-core:
```
/etc/insights-client/rpm.egg/insights/parsers/__init__.py:160: SyntaxWarning: invalid escape sequence '\ '
```
The new test explicitly requires that stderr is empty, which thus causes the failure. Is this (stderr as empty) a hard requirement for considering the playbook verifier working/successful?
Also, looking at the current two users of the verifier (rhc-worker-playbook and rhc-worker-script):
- none of them actually checks for stderr when the validation succeeds
- both of them seem to print stderr on failure

so maybe we can ignore stderr on success?

ptoscano:
Hmm there used to be
```python
@pytest.mark.xfail(
    condition=sys.version_info >= (3, 12),
    reason="Verification is known to be broken on Python 3.12+",
)
```
in previous versions of this PR, and indeed the new `test_official_playbook` test fails on OSes with Python >= 3.12 (so Fedora >= 39); should that mark be added back, until the verifier is fixed?

m-horky:
Good call, I was testing it with locally patched egg, I forgot to put it back.

ptoscano (approved):
LGTM, thanks!

PR #254, opened 2024-06-12
Enable redhat-access-insights only in RHEL < 10

ptoscano:
There is no need for it anymore, so keep it only for existing enterprise
distributions.

A new `redhat_access_insights` meson option (which is false by default) is used to enable that deprecated executable.

Card ID: CCT-419


subpop (approved):
LGTM!

subpop:
I verified that the RHEL 9 RPM does include `/usr/bin/redhat-access-insights`, while the Fedora 10 does not.

PR #255, opened 2024-06-14
fix(test): removing --register option when using ----output-file

Archana-PandeyM:
register option is not considered when using --output-file option because
 --output-file only runs in offline mode, so even if you pass register, system will not be registered to insights.


zhangqianqian:
LGTM

m-horky (approved):
LGTM.

However, should this be a separate test case then? "If you pass `--output-file` with `--register`, the system will not get registered?"

Archana-PandeyM:
@m-horky yes I will cover while working on coverage.

PR #257, opened 2024-06-17
fix: check lspci and dmidecode specs on x86_64

zhangqianqian:


integration-tests/test_common_specs.py
```diff
@@ -1,31 +1,54 @@
...
+            "insights.specs.Specs.ls_dev.json",
+            "insights.specs.Specs.lscpu.json",
+            "insights.specs.Specs.meminfo.json",
+            "insights.specs.Specs.mount.json",
+            "insights.specs.Specs.mountinfo.json",
+            "insights.specs.Specs.ps_auxcww.json",
+            "insights.specs.Specs.redhat_release.json",
+            "insights.specs.Specs.uname.json",
+            "insights.specs.Specs.yum_repos_d.json",
+        ]
```
xiangce, line 51: Looks good to me. Just a tiny suggestion, how about making the following change to make the difference between `x86_64` and other archs a bit clear?
```python
     common_specs = [
            "insights.specs.Specs.date.json",
            "insights.specs.Specs.fstab.json",
            "insights.specs.Specs.hostname.json",
            "insights.specs.Specs.hosts.json",
            "insights.specs.Specs.installed_rpms.json",
            "insights.specs.Specs.ip_addresses.json",
            "insights.specs.Specs.ls_dev.json",
            "insights.specs.Specs.lscpu.json",
            "insights.specs.Specs.meminfo.json",
            "insights.specs.Specs.mount.json",
            "insights.specs.Specs.mountinfo.json",
            "insights.specs.Specs.ps_auxcww.json",
            "insights.specs.Specs.redhat_release.json",
            "insights.specs.Specs.uname.json",
            "insights.specs.Specs.yum_repos_d.json"
    ]

    if platform.processor() == "x86_64":
        common_specs.extends([
            "insights.specs.Specs.dmidecode.json",
            "insights.specs.Specs.lspci.json",
        ])
```

zhangqianqian: Thanks. common_specs.extends => common_specs.extend

integration-tests/test_common_specs.py
```diff
@@ -26,6 +28,13 @@ def test_common_specs(insights_client, tmp_path):
         "insights.specs.Specs.uname.json",
         "insights.specs.Specs.yum_repos_d.json",
     ]
+    if platform.processor() == "x86_64":
+        common_specs.extend(
+            [
+                "insights.specs.Specs.dmidecode.json",
+                "insights.specs.Specs.lspci.json",
```
ptoscano, line 35: This is not correct: PCI is available on basically every mainstream and new architecture nowadays.

This is a problem in insights-core, which assumes an empty output of `lspci` is a bug.

integration-tests/test_common_specs.py
```diff
@@ -26,6 +28,13 @@ def test_common_specs(insights_client, tmp_path):
         "insights.specs.Specs.uname.json",
         "insights.specs.Specs.yum_repos_d.json",
     ]
+    if platform.processor() == "x86_64":
+        common_specs.extend(
+            [
+                "insights.specs.Specs.dmidecode.json",
```
ptoscano, line 34: Technically speaking, dmidecode is usable also on few more architectures other than x86_64; for example aarch64.

ptoscano (changes requested):
This partially papers over bugs in insights-core; see my latest reply to RHINENG-10737.

integration-tests/test_common_specs.py
xiangce: As talked in Slack, this `ContentException('Empty content')` is expected for insights-core. Although the empty output of `lspci` is valid for s390x, it (the empty lspci file in insights-archive) is useless for insights, even if it's empty on `x86_64`, such a `ContentException` will also be recorded in its meta_data JSON `errors` section. The `errors` section in the JSON file indicates why this file is not collected in the archive, but doesn't mean this is an `error` of the collection.

On s390x, in most cases (particular on the test VM machine), its output is always empty. so, here in this test, we can ignore check it.

ptoscano: > It is working exactly as what we expect in the design.

As I said in other conversations (since this was discussed outside of this PR, sigh): I consider that design broken, since it has expectations that are not actually what happens in real life systems. If the spec collection considers something that can actually happen (no PCI devices) as "error", then the collection ought to be fixed.

> Although the empty output of lspci is valid for s390x, it (the empty lspci file in insights-archive) is useless for insights, even if it's empty on x86_64, such a `ContentException` will also be recorded in its meta_data JSON errors section.

Then there are two ways to fix this:
- consider an empty `lspci` output an error ONLY on x86_64; on other architectures, do not collect the empty file and do not report any error
- simply send the empty file in case it is there, and let Insights/Advisory/whatever decide how to consider the situation (valid, invalid) based on the rest of the collected details of a system

koalakangaroo: @ptoscano I think you mixed the use of "error" with "exception" in your expressions. "Exception" does not equal to "error".  To insights-core, ContentException is designed for such cases like lspci and dmidcode. I don't think it is worth nor a good idea to force all Insights services to change their code logic using data collection with what the ContentException is designed for.

PR #258, opened 2024-06-17
feat: add code to check branch info against satellite

zhangqianqian:


Archana-PandeyM:
please merge commits.
LGTM.

integration-tests/test_collection.py
```diff
@@ -152,8 +159,16 @@ def test_branch_info(insights_client):
...
         data = json.load(file)
-        assert data["remote_branch"] == -1, "Incorrect remote_branch value"
-        assert data["remote_leaf"] == -1, "Incorrect remote_leaf value"
+        if "satellite" in test_config.environment:
+            assert data["product"]["type"] == "Satellite"
+            assert isinstance(uuid.UUID(data["remote_branch"]), uuid.UUID)
+            cmd = "subscription-manager identity"
+            output = subprocess.check_output(cmd, universal_newlines=True, shell=True)
+            system_identity = output.split("\n")[0].split(":")[1].strip()
+            assert data["remote_leaf"] == system_identity
```
ptoscano, line 168: I just added a `SubscriptionManager.uuid` property in pytest-client-tools, which returns a `uuid.UUID` object for a registered system; now you can simply use it as:
```python
def test_branch_info(insights_client, test_config, subman):
   ...
            assert data["remote_leaf"] == subman.uuid
```

ptoscano (changes requested):

integration-tests/test_collection.py
```diff
@@ -152,8 +159,13 @@ def test_branch_info(insights_client):
     branch_info_path = os.path.join(directory_name, "branch_info")
     with open(branch_info_path, "r") as file:
         data = json.load(file)
-        assert data["remote_branch"] == -1, "Incorrect remote_branch value"
-        assert data["remote_leaf"] == -1, "Incorrect remote_leaf value"
+        if "satellite" in test_config.environment:
+            assert data["product"]["type"] == "Satellite"
+            assert isinstance(uuid.UUID(data["remote_branch"]), uuid.UUID)
+            assert uuid.UUID(data["remote_leaf"]) == subman.uuid
```
ptoscano, line 165: - the first assertion is a tautology (always true): if the UUID string can be parsed from `data["remote_branch"]`, then the result is always a `uuid.UUID` object
- if `uuid.UUID()` cannot parse a UUID string, it raises an exception (good, will fail the test as expected)
- the second assertion builds again a `uuid.UUID` object

So I'd simplify these two lines a bit:
```python
remote_branch_uuid = uuid.UUID(data["remote_branch"])
assert remote_branch_uuid == subman.uuid
```
- a failing UUID string properly causes `uuid.UUID()` to raise an exception
- only one `uuid.UUID` object constructed
- compare against the sub-man UUID as already done

ptoscano (changes requested):

integration-tests/test_collection.py
zhangqianqian: data["remote_branch"] and data["remote_leaf"] represent two different things. 
A branch is an organization for Satellite 6, is a UUID id of the upstream consumer manifest for organization to which host is subscribed. A leaf means subscription manager system identity UUID.

Here I use "assert isinstance(uuid.UUID(data["remote_branch"]), uuid.UUID)" to check the branch info is a uuid object, but I don’t check the exact value of it as the manifest can be re-imported/changed if necessary.
So I just check the leaf value, use "assert uuid.UUID(data["remote_leaf"]) == subman.uuid"

zhangqianqian: pls refer https://docs.google.com/document/d/193mN5aBwxtzpP4U-vnL3yVawPxiG20n0zkzDYsam-eU/edit

integration-tests/test_collection.py
```diff
@@ -4,6 +4,9 @@
 import glob
 import json
 import pytest
+import subprocess
```
Archana-PandeyM, line 7: please remove unused imports

Archana-PandeyM (changes requested):

PR #262, opened 2024-06-24
fix(test): Updated logrotate test to verify payload log rotation too

Archana-PandeyM:
Test need to verify logration for insights-client-payload.log files as well.
	The test was failing recently because it was only expecting insights-client.log in the log dir
	which was an incorrect assumption because insights-client-payload.log files are
	written in same log dir when --payload option is used

### DIFF
--- a/integration-tests/test_client.py
+++ b/integration-tests/test_client.py
@@ -82,21 +82,50 @@ def test_verify_logrotate_feature(insights_client):
     Ref : https://bugzilla.redhat.com/show_bug.cgi?id=1940267
 
     Test Steps:
-        1-Count the number of log file in beginning
-        2-Rotate logs by running logrotate command on CLI for insights-client
-        3-Verify that log is rotated and new log file is created
+        1 - Perform register and payload operation to ensure both
+            insights-client.log and insights-client-payload.log files have logs
+        2 - Rotate logs by running logrotate command on CLI for insights-client
+        3 - Verify insights-client.log and insights-client-payload.log size is 0B
+        4 - Verify rotated files are created in log dir by comparing number of log files
     """
 
     logrotate_conf_file_path = "/etc/logrotate.d/insights-client"
     logdir = "/var/log/insights-client/"
+    logfile_insights = f"{logdir}/insights-client.log"
+    logfile_payload = f"{logdir}/insights-client-payload.log"
+
     assert os.path.exists(logrotate_conf_file_path), "logrotate is not configured"
-    # It is necessary to perform some command using insights-client
-    # to populate logs
-    insights_client.register()
-    number_of_log_files = len(os.listdir(logdir))
+    """
+     It is necessary to perform some command using insights-client
+     to populate logs.
+     Save the archive, to be used while register operation using --keep-archive.
+     for example-
+     [root@test ~]# insights-client --register --keep-archive
+     Automatic scheduling for Insights has been enabled.
+     Starting to collect Insights data for test
+     Writing RHSM facts to /etc/rhsm/facts/insights-client.facts ...
+     Uploading Insights data.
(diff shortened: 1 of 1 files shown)


ptoscano:
Hm I'm not sure I understand what is going on, and how the changes to the registration and the added upload would change the situation. Would you please explain in the detail what is the situation that causes the failure (i.e. what you notice it happens)?

Archana-PandeyM:
> Hm I'm not sure I understand what is going on, and how the changes to the registration and the added upload would change the situation. Would you please explain in the detail what is the situation that causes the failure (i.e. what you notice it happens)?

As I have mentioned in the description above, tests started failing when it noticed insights-client-payload.log files in /var/log/insights-client/ directory. It happened so because there was an incorrect assumption that log directory only has insights-client.log files
When you perform a manual upload using --payload option then logs are getting written in insights-client-payload.log and hence tests started failing because number of files in log directory increased.

Now the second part of it, which is changes to registration and upload operation- that is done to ensure that insights-client.log and insights-client-payload.logs are not empty because log rotation is disabled on empty files.

Insights-client.log gets logs during registration but payload logs are populated only when --payload is run.

integration-tests/test_client.py
```diff
@@ -82,21 +82,37 @@ def test_verify_logrotate_feature(insights_client):
...
     assert os.path.exists(logrotate_conf_file_path), "logrotate is not configured"
     # It is necessary to perform some command using insights-client
     # to populate logs
-    insights_client.register()
-    number_of_log_files = len(os.listdir(logdir))
+    # Save the archive, to be used while upload operation
+    reg_result = insights_client.run("--register", "--keep-archive")
+    assert conftest.loop_until(lambda: insights_client.is_registered)
+
+    archive_name = reg_result.stdout.split()[-1]
```
ptoscano, line 104: What's the exact output of `insights-client --register --keep-archive`? This line seems to assume the last line contains only the archive name -- if that is the case, I'd add a comment before this line with a sample of the output, so it's easier to know why this text extraction is done this way.

integration-tests/test_client.py
```diff
@@ -82,21 +82,37 @@ def test_verify_logrotate_feature(insights_client):
...
+    insights_client.run(
+        f"--payload={archive_name}",
+        "--content-type=gz",
+    )
+    number_of_log_files = len(os.listdir(logdir))  # count of files before rotation
+
     subprocess.check_call(["logrotate", "-vf", logrotate_conf_file_path])
+    assert os.stat(logfile_insights).st_size == 0
+    assert os.stat(logfile_payload).st_size == 0
```
ptoscano, line 113: Since we need only the size of these files, `os.path.getsize()` is a shorter way to get it.

integration-tests/test_client.py
```diff
@@ -82,21 +82,37 @@ def test_verify_logrotate_feature(insights_client):
...
+            insights-client.log and insights-client-payload.log files have logs
+        2 - Rotate logs by running logrotate command on CLI for insights-client
+        3 - Verify insights-client.log and insights-client-payload.log size is 0B
+        4 - Verify rotated files are created in log dir by comparing number of log files
     """
 
     logrotate_conf_file_path = "/etc/logrotate.d/insights-client"
     logdir = "/var/log/insights-client/"
+    logfile_insights = "/var/log/insights-client/insights-client.log"
+    logfile_payload = "/var/log/insights-client/insights-client-payload.log"
```
ptoscano, line 95: rather than duplicating "/var/log/insights-client", you can do something like:
```python
       logfile_insights = f"{logdir}/insights-client.log"
```

ptoscano (approved):
```
out: integration-tests/test_client.py::test_verify_logrotate_feature PASSED   [ 14%]
```
thanks!

PR #263, opened 2024-07-03
test: add test for motd message

zhangqianqian:


integration-tests/test_motd.py
```diff
@@ -70,3 +70,25 @@ def test_motd_dev_null(insights_client):
 
         insights_client.unregister()
         assert os.path.samefile(os.devnull, MOTD_PATH)
+
+
+@pytest.mark.usefixtures("delete_special_files")
+def test_motd_message():
```
Archana-PandeyM, line 76: this test is getting skipped upstream because of [fixture use](https://github.com/RedHatInsights/insights-client/pull/263/files#diff-15ad4e59aa16a6da740f65ca9411ab1ce78a05d9140703c5301c561182b4e4f9L14) 
This test neither needs system to be registered via sub-man nor by insights, I would suggest to update the other two test to specifically call fixture "resgiter_subman" instead of module wise use.
for example:
@pytest.mark.usefixtures("delete_special_files")
def test_motd(insights_client):

By making this change to other two tests, we will be able to run 'test_motd_message()' this upstream.

Archana-PandeyM (changes requested):

Archana-PandeyM:
LGTM, please clean and squash commit messages

PR #267, opened 2024-07-24
fix: use system temporary directory when running as non-root

ptoscano:
The temporary directory chosen to work properly with SELinux is `/var/lib/insights`, which can be accessed only by root.

`insights-client` is basically not usable as non-root, and it will fail with `Insights client must be run as root.`. The only exception here is `insights-client --version`, which is run only in the client without loading the core; the exception is that the client will still verify the available eggs using GPG by default.

Hence, use the default temporary directory when running as non-root: this way the available 
eggs can be validated, and the client will either
- print the version and exit
- keep exiting saying root is required

Followup of commit dd6293d608f61a44e562a1fda44cad82cf667e7d.


m-horky (approved):
LGTM

The downstream test is failing on unrelated `integration-tests/test_client_options.py::test_check_show_results` because the Advisor rule likely changed.

PR #268, opened 2024-07-24
test: drop manual import of GPG key

ptoscano:
Both the client itself and insights-core were fixed to properly do the GPG validation with new versions of gnupg; since the fixed version of insights-core is now part of the client sources, the workaround can be dropped altogether.


m-horky (approved):
LGTM

PR #270, opened 2024-08-01
feat(test): Adding betelgeuse docstrings to test_client.py

zpetrace:
Adding betelgeuse docstrings to test_client.py

### DIFF
--- a/integration-tests/test_client.py
+++ b/integration-tests/test_client.py
@@ -1,3 +1,11 @@
+"""
+:casecomponent: insights-client
+:requirement: RHSS-291297
+:subsystemteam: sst_csi_client_tools
+:caseautomation: Automated
+:upstream: Yes
+"""
+
 import contextlib
 import glob
 import os
@@ -8,9 +16,22 @@ import conftest
 
 @pytest.mark.usefixtures("register_subman")
 def test_client_files_permission(insights_client):
-    """Verify that permission for lastupload file is 0644:
-     file /etc/insights-client/.lastupload
-    Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1924990
+    """
+    :id: e793cf5e-1c25-4e31-93c9-6f0465f50cae
+    :title: Verify client files permission
+    :description:
+        Verify that the permission for the last upload file
+        /etc/insights-client/.lastupload is set to 0644
+    :reference: https://bugzilla.redhat.com/show_bug.cgi?id=1924990
+    :tier: Tier 1
+    :steps:
+        1. Remove /etc/insights-client/.lastupload if it exists
+        2. Register insights-client
+        3. Verify the file permissions
+    :expectedresults:
+        1. The file /etc/insights-client/.lastupload does not exist
+        2. The insights-client is registered
+        3. The permission of /etc/insights-client/.lastupload is set to 0644
     """
     file_last_upload = "/etc/insights-client/.lastupload"
     with contextlib.suppress(FileNotFoundError):
@@ -46,7 +67,16 @@ def rpm_ql_insights_client():
 )
 def test_client_rpm_mandatory_files(filename, rpm_ql_insights_client):
     """
-    Verify the existence of mandatory files for insights-client rpm
+    :id: c7d2edbe-ae78-47e0-9b3d-ae1634c0ac79
+    :title: Verify mandatory files for RPM
+    :description: Verify the existence of mandatory files for the insights-client RPM
+    :tier: Tier 1
+    :steps:
+        1. List all files in the insights-client RPM package
+        2. Check if each mandatory file exists in the package
+    :expectedresults:
+        1. A list of files is generated
+        2. All of the mandatory files are present in the RPM
     """
     assert (
         filename in rpm_ql_insights_client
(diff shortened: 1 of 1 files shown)


Lorquas (commented):
Looks good to me.  How are you generating the test cases and/or adding them to the junit.xml with betelgeuse? I would like to test this out directly...

ptoscano:
> I would like to test this out directly...

Indeed -- this definitely needs to be tested with a github workflow/job, to verify that the format of these various comments, otherwise IMHO it's too easy to screw them up accidentally.

Lorquas:
Other note, don't we need an `:id: <uuid>` field?

Lorquas:
Update from slack convo: 

- We're going to co-opt this PR and add a README for the various manual betelgeuse invocations
- This PR will be blocked by CCT-675 where we will attempt to add .github/workflow and other automations (using this branch as a starting point)

integration-tests/test_client.py
```diff
@@ -78,15 +122,24 @@ def test_client_logdir_permissions():
...
+        4. Verify that 2 new log files were created
+        5. Verify the size of insights-client.log
+        6. Verify the size of insights-client-payload.log
+    :expectedresults:
+        1. The logrotate config file exists
+        2. The insights-client registers successfully and logs are populated
+        3. The logrotate command is executed successfully
+        4. Two new log files were created
+        5. The size is 0B
+        6. The size is 0B
```
ptoscano, line 142: "the size" ... of what?

integration-tests/test_client.py
```diff
@@ -131,8 +184,20 @@ def test_verify_logrotate_feature(insights_client):
...
+    :title: Verify insights-client details file exists
+    :description: Verify that the file /var/lib/insights/insights-client.json exists
+    :tier: Tier 0
+    :steps:
+        1. Register the insights-client
+        2. Delete /var/lib/insights/insights-client.json if it exists
+        3. run the --check-results command
+        4. Verify the existence of /var/lib/insights/insights-client.json
+    :expectedresults:
+        1. Insights-client is registrated
```
ptoscano, line 197: typo "registrated"

integration-tests/test_client.py
```diff
@@ -55,9 +84,18 @@ def test_client_rpm_mandatory_files(filename, rpm_ql_insights_client):
 
 @pytest.mark.usefixtures("register_subman")
 def test_client_logfiles_mask(insights_client):
-    """Verify that files in /var/log/insights-client have the right mode: 0600
-
-    Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1955724
+    """
+    :id: 8f24500c-d0ff-4ab1-a2ae-3b99cbf68e36
+    :title: Verify Client logfiles Mask
```
ptoscano, line 89: this is not about a file/permission mask (like umask), but rather about the permissions themselves

ptoscano (commented):
There is a bit of inconsistency in the various steps & results that talk about registering with `insights-client`:

Steps:

>         2. Register insights-client
>         1. Register the insights-client to generate log files
>         2. Register insights-client and perform pazload operations
>         1. Register the insights-client

Results:

>         2. The insights-client registers successfully
>         1. Insights-client is registered and log files are generated
>         2. The insights-client registers successfully and logs are populated
>         1. Insights-client is registrated
>         1. Insights-client is registered

Ignoring the typos and the numbers, what I personally suggest is:
- keep the step & result to only mention the registration & its successfull result
- move other bits to own steps & results
- use a single wording, for example:
  - [step] Register insights-client
  - [result] Insights-client is registered successfully

This approach applies also to the testimony docstrings added in other PRs.

---------------

I see `testimony` validation failures for all the tests like the following:
```
test_client_files_permission:19
-------------------------------

* Tokens with invalid values:
  Tier: Tier 0
      type: choice
      case sensitive: False
      choices: ['tier 1', 'tier 2', 'tier 3']
```
I'm using the `testimony.yml` as currently available in #282, which has:
```yaml
Tier:
  casesensitive: false
  required: true
  type: choice
  choices:
    - Tier 1
    - Tier 2
    - Tier 3
```
Hm, I wonder whether this seems like a bug in `testimony`...

integration-tests/test_client.py
```diff
@@ -1,3 +1,12 @@
+"""Global Values
```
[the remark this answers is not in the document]
zpetrace: It's not, it convenieces me but we can omit that.

integration-tests/test_client.py
zpetrace: of insights-client.log and insights-client-payload.log

zpetrace:
I tried to get rid of the inconsistencies in wordings and also correct the typos etc. However I am not sure I understood the
> move other bits to own steps & results
part. If you can explain some more, please.

integration-tests/test_client.py
```diff
@@ -131,8 +189,20 @@ def test_verify_logrotate_feature(insights_client):
...
+    :description: Verify that the file /var/lib/insights/insights-client.json exists
+    :tier: Tier 1
+    :steps:
+        1. Register insights-client
+        2. Delete /var/lib/insights/insights-client.json if it exists
+        3. run the --check-results command
+        4. Verify the existence of /var/lib/insights/insights-client.json
+    :expectedresults:
+        1. Insights-client is registered
+        2. The file /var/lib/insights/insights-client.json does not exists
```
m-horky, line 203: `s/exists/exist`

integration-tests/test_client.py
```diff
@@ -131,8 +189,20 @@ def test_verify_logrotate_feature(insights_client):
...
-    This test verifies that /var/lib/insights/insights-details.json exists
-    when --check-results is called
+    :id: 2ccc8e00-0e76-47fd-bdb2-27998c0094ab
+    :title: Verify insights-client details file exists
+    :description: Verify that the file /var/lib/insights/insights-client.json exists
+    :tier: Tier 1
+    :steps:
+        1. Register insights-client
+        2. Delete /var/lib/insights/insights-client.json if it exists
+        3. run the --check-results command
```
m-horky, line 199: Should that start with capital R?

zpetrace:
I have changed the capitalization to the same format for every test as it was confusing. Hope that's better:)

ptoscano (approved):

PR #281, opened 2024-08-29
chore: wrap/split long lines

ptoscano:
- wrap/split few lines (code & comments) longer than 88
- drop no more needed per-file ignores from flake8 config

No behaviour changes, only formatting.


m-horky (approved):
LGTM

PR #282, opened 2024-08-30
Adding a Github workflow for testimony validation

zpetrace:
After more discussion this is a PR for testimony workflow only and should be merged after all the classes have the docstrings added.

Card ID: CCT-1152


zpetrace:
After the last commit the testimony should be set to expect tags instead of tiers.

zpetrace:
I have discovered, that the :reference: field cannot be left blank, so if the test doesn't have a reference, the field cannot be included and left blank - now the checks are passing, therefore opening this PR for a review:)

ptoscano:
> I have discovered, that the :reference: field cannot be left blank, so if the test doesn't have a reference, the field cannot be included and left blank - now the checks are passing, therefore opening this PR for a review:)

Yay! Not to push for more work: would it be possible to split the fixes to the docstrings in its own commit (even in this PR, I think it should be fine)? This way it should be easier to document, to review, and to navigate the git history about those changes.

Thanks!

ptoscano (approved):
thanks!

PR #289, opened 2024-09-20
test: Added test for file worflow in archive

Archana-PandeyM:
Test to verify that change in data of files in archive updates the host data in inventory when reuploaded. I have modified the test to run from integration test perspective instead of rewriting the existing test in iqe

### DIFF
--- a/integration-tests/test_file_workflow.py
+++ b/integration-tests/test_file_workflow.py
@@ -7,7 +7,11 @@
 """
 
 import json
+import random
+import string
+import subprocess
 import tarfile
+from time import sleep
 import pytest
 from constants import HOST_DETAILS
 from constants import MACHINE_ID_FILE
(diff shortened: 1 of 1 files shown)


integration-tests/test_file_workflow.py
```diff
@@ -114,6 +118,34 @@ def test_file_workflow_with_an_archive_without_canonical_facts(
...
+    """
+    insights_client.register()
+    assert conftest.loop_until(lambda: insights_client.is_registered)
+    current_hostname = subprocess.check_output("hostname", shell=True).decode().strip()
+    try:
+        # set a new hostname
+        new_hostname = set_hostname()
+        # collect data and upload again
+        insights_client.run()
+        sleep(30)  # small wait for data to get reflected in inventory
```
m-horky, line 141: I'm gonna assume there's no better way to do this.

Archana-PandeyM: unfortunately yes there is no other way at this point then being hopeful(sleep).

m-horky (approved):
Internal CI failed on unrelated errors. LGTM.

PR #298, opened 2024-09-27
feat(test): Adding docstrings to test_version.py

zpetrace:
Solves CCT-674


ptoscano:
Looks OK; would you please squash the two commits into one, with a proper commit message, and then rebase? Thanks!

PR #303, opened 2024-10-29
fix(test): Changing the custom fields to enable upload to Polarion

zpetrace:
Requirement is a built-in value in Polarion so while uploading the tests it failed to set up the custom value, therefore I deleted it and the upload to Polarion looks much better, more information in:
[CCT-616](https://issues.redhat.com/browse/CCT-616)


ptoscano:
hmm what about the various markers
```python
:requirement: RHSS-291297
```
added in docstrings so far?

zpetrace:
I have updated the commit message, hope it helps:)

m-horky (approved):
LGTM

Aftermath (approved):
This was required because polarion has `requirement` as a built-in value, rather than a custom-field. LGTM.

PR #305, opened 2024-10-30
ci: add workflow for codespell

ptoscano:
Add a simple workflow to run codespell as part of the checks for every PR.


m-horky (approved):
LGTM

PR #306, opened 2024-10-31
chore(test): Improve tests in test_collection.py

m-horky:
- It is not required to register insights-client to run the collection. This speeds up the test.
- There is a simpler way to get a content; '--output-dir' saves the content uncompressed into user-controlled location.
- The archive may contain less content than what the test originally required.

### DIFF
--- a/integration-tests/test_collection.py
+++ b/integration-tests/test_collection.py
@@ -331,8 +331,7 @@ def test_branch_info(insights_client, test_config, subman):
             assert data["remote_leaf"] == -1, "Incorrect remote_leaf value"
 
 
-@pytest.mark.usefixtures("register_subman")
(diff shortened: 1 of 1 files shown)


ptoscano (commented):
- please drop the `register_subman` fixture usage from `test_branch_info`, as it does not need to connect anymore
- please drop the `ARCHIVE_CACHE_DIRECTORY` constant from `test_collection.py`, as it is no more used

m-horky:
That `register_subman` fixture is still required for the Satellite subcase.

zpetrace (approved):
Yeah these changes will probably speed up the process a bit, I also think the fixture is still needed so LGTM, thanks!:)

ptoscano (approved):

PR #307, opened 2024-10-31
fix(test): Improve test_common_specs

m-horky:
* Card ID: CCT-934

Specs that run 'hostname' or try to access special system files (e.g. /etc/fstab) fail in containers. This patch ensures spec contents are not asserted if there's no way they were collected.


zpetrace (approved):
LGTM, thanks for the update @m-horky!

PR #310, opened 2024-11-05
fix(test): Skip hostname redaction in containers

m-horky:
* Card ID: CCT-934

Containers are not in control of their own hostnames, they are usually set from the host system. Because this test runs 'hostnamectl set-hostname', it has to be skipped when we're not running on a bare metal/VM system.


zpetrace (approved):
LGTM, thanks for the change @m-horky :)

PR #311, opened 2024-11-06
Feat/cct-144 lines # core_collect=True ...

grunwmar:
Lines # core_collect and its description above it
were removed from `insights-client.conf`.

CARD CCT-144


ptoscano (approved):
LGTM, thanks!

PR #312, opened 2024-11-07
feat: Added a fixture for integration tests to check selinux denials

Archana-PandeyM:


integration-tests/conftest.py
```diff
@@ -58,3 +56,18 @@ def loop_until(predicate, poll_sec=5, timeout_sec=120):
...
+    in the system after tests are run."""
+    yield
+    options = "ausearch -m avc -m user_avc -m selinux_err -i".split()
+    result = logged_run(options, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    if "<no matches>" not in str(result.stdout):
+        lines = str(result.stdout).split("\n")
+        for line in lines:
+            words = line.split()
+            if "denied" in words:
+                assert "permissive=1" in words, "SELinux AVC denials found"
```
m-horky, line 73: You could simplify this a bit by passing `text=True` to `logged_run()`. This would be an equivalent:

```py
result = logged_run(options, capture_output=True, text=True)
for line in result.stdout.split("\n"):
    if "denied" in line:
        assert "permissive=1" in line, "SELinux AVC denials found"
```

m-horky (commented):
This will make all test fail even if some other RHEL package has SELinux misconfigured and gets run. Which I guess is good, but it could poison our results for some time until it gets fixed.

Would it make sense to only check lines that might be related to our tools (e.g. insights-client, gpg, python, ...?)

ptoscano:
While the idea is good, I think this ought to be done rather in pytest-client-tools:
- the fixture can run potentially after fixtures, so missing commands
- pytest-client-tools, as it is a pytest plugin, knows already about the execution time of the test, and thus it is possible to properly restrict the lookup of the SELinux denials for each test specifically (see the checkpoints in ausearch)
- pytest-client-tools also already has logs for each test, and the log with SELinux denials would be an additional one

Archana-PandeyM:
> While the idea is good, I think this ought to be done rather in pytest-client-tools:
> 
> * the fixture can run potentially after fixtures, so missing commands
> * pytest-client-tools, as it is a pytest plugin, knows already about the execution time of the test, and thus it is possible to properly restrict the lookup of the SELinux denials for each test specifically (see the checkpoints in ausearch)
> * pytest-client-tools also already has logs for each test, and the log with SELinux denials would be an additional one

Thanks for feedback, I agree that we should put it in pytest_client_tools. Eventually we would add such checks in rhc repo as well. So would you suggest keeping this fixture in pytest_client_tools/plugin.py ?

Archana-PandeyM:
> This will make all test fail even if some other RHEL package has SELinux misconfigured and gets run. Which I guess is good, but it could poison our results for some time until it gets fixed.
> 
> Would it make sense to only check lines that might be related to our tools (e.g. insights-client, gpg, python, ...?)

Yes I planned to do so.

integration-tests/conftest.py
Archana-PandeyM: I am ok to use capture_output=True instead of subprocess.PIPE. Only reason for using the later was to avoid python version issues. I think capture_output is available in version > python 3.7. But it should not affect the tests, I will modify the command. Thanks!

ptoscano:
> I agree that we should put it in pytest_client_tools. Eventually we would add such checks in rhc repo as well. So would you suggest keeping this fixture in pytest_client_tools/plugin.py ?

Not as a fixture, no. There are already hooks that track certain parts of the tests execution flow, and the SELinux checks would need to be added there:
- before starting the test, a checkpoint needs to be created with ausearch
- when a test finishes, get the list of denials between the checkpoint and the test completion, logging the result

In pytest-client-tools there are already the bits to handle per-test stuff, so this should not be complicated to add.

integration-tests/conftest.py
Archana-PandeyM: @m-horky Like I suspsected 'capture_output' is failing on RHEL8 tests(python 3.6), so I am reverting the change back to original commit.

Archana-PandeyM:
> > I agree that we should put it in pytest_client_tools. Eventually we would add such checks in rhc repo as well. So would you suggest keeping this fixture in pytest_client_tools/plugin.py ?
> 
> Not as a fixture, no. There are already hooks that track certain parts of the tests execution flow, and the SELinux checks would need to be added there:
> 
> * before starting the test, a checkpoint needs to be created with ausearch
> * when a test finishes, get the list of denials between the checkpoint and the test completion, logging the result
> 
> In pytest-client-tools there are already the bits to handle per-test stuff, so this should not be complicated to add.

Would you suggest me to implement this in pytest_client_tools, Can you help me point to the bits that handle per-test stuff because I tried but could not figure out.

ptoscano:
This was recently implemented in pytest-client-tools, requiring `auditd` to be running and `auditctl` & `ausearch` to be installed (all of them are default in RHEL). The result is a new `selinux.log` artifact for each test in which there were denials logged.

Hence, I'm closing this PR. Thanks anyway!

PR #314, opened 2024-11-12
Feat/CCT-703: Skip tests not applicable to insights-core PR

grunwmar:
A string of path `/etc/yum.conf` was removed from list variable `TEST_FILE` in a file `integration-tests/test_redaction`.

CARD ID CCT-703


m-horky (changes requested):
The patch looks good.

In the commit message, could you please also say _why_ it was made? Just that the custom eggs don't have the filters.yaml file, and we are also running these tests with eggs built from master.

It will help us in the future when we are not sure why this change was made.

m-horky (approved):
LGTM

PR #316, opened 2024-11-28
spec: remove unused python3-magic dependency

yselkowitz:
The python file/libmagic bindings are now named python3-file-magic, while python3-magic is now provided by a separate package of a PyPI module.  In any case, it does not appear that either is used in insights-clients.

Related: https://bugzilla.redhat.com/show_bug.cgi?id=2020606

### DIFF
--- a/insights-client.spec
+++ b/insights-client.spec
@@ -24,7 +24,6 @@ Requires: pciutils
 %{?systemd_requires}
 Requires: python3-requests >= 2.6
 Requires: python3-PyYAML
-Requires: python3-magic
 Requires: python3-six
 Requires: python3dist(setuptools)
 Requires: coreutils


Lorquas:

m-horky:
Hi, @yselkowitz. This gets a bit more complicated. This `insights-client` repository is mostly an RPM-based wrapper around `insights-core` which does use the `magic` library; the client `Requires` it on its behalf.

If the `python3-file-magic` package is supposed to be used instead of the `python3-magic` one, which systems are affected? Which Fedora/CentOS Stream/RHEL versions?

m-horky (approved):

yselkowitz:
For the record, to answer your original question, the RPM in question has been properly named python3-file-magic since RHEL 9.

PR #319, opened 2024-12-17
[el10] test(ci): Drop CentOS 8 and 9 containers

m-horky:
'master' now targets RHEL 10/CentOS Stream 10, we don't need to test these.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -4,6 +4,7 @@ on:
   pull_request:
     paths:
       - "src/**"
+      - ".github/workflows/pytest.yml"
 
 jobs:
   pytest:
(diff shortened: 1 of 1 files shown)


m-horky:
```
Start: installing minimal buildroot with dnf
No matches found for the following disable plugin patterns: local, spacewalk, versionlock
Copr repository                                 2.2 kB/s | 133  B     00:00    
CentOS Stream 10 - BaseOS                       953 kB/s | 6.2 MB     00:06    
CentOS Stream 10 - AppStream                    2.5 MB/s | 2.4 MB     00:00    
CentOS Stream 10 - CRB                          239 kB/s | 646 kB     00:02    
CentOS Stream 10 - Extras packages              6.5 kB/s | 3.5 kB     00:00    
Error: 
 Problem: cannot install the best candidate for the job
  - nothing provides rpm = 4.19.1.1-9.el10 needed by rpm-build-4.19.1.1-9.el10.x86_64 from appstream
(try to add '--skip-broken' to skip uninstallable packages or '--nobest' to use not only best candidate packages)
```

It looks like there are some problems with the repositories in C10S. Perhaps we should wait a few days.

PR #320, opened 2024-12-17
[el8] test(ci): Drop CentOS Stream 10 and Fedora containers

m-horky:
This branch now only targets 'el8' and 'el9', it is not necessary to run on newer systems, as they start to diverge.


m-horky:
It looks like there are some problems with the repositories in C10S. Packit will be addressed in a separate card; CI run by GitHub Action passes.

PR #321, opened 2024-12-19
docs: Remove reference to BASIC authentication

pkoprda:
* Card ID: CCT-238

Since BASIC authentication has reached its EOL in insights services, it is no longer relevant or supported in the client tool. This update removes all references to BASIC authentication from the client configuration and documentation.


m-horky (approved):
LGTM.

PR #329, opened 2025-01-07
fix: Disable legacy API by default in RHEL 10

pkoprda:
* Card ID: RHEL-67937
* Card ID: CCT-1003

Ensure that client uses the non-legacy API URI `cert.cloud.redhat.com` by default. Previously, some commands temporarily switched to the non-legacy API URI, but this behavior was not consistent across all commands.

---

- [ ] `el8` (RHEL 8, RHEL 9)
- [ ] `el7` (RHEL 7)

### DIFF
--- a/data/insights-client.conf
+++ b/data/insights-client.conf
@@ -8,7 +8,7 @@
 #auto_config=True
 
 # Base URL for the Insights API
-#base_url=cert-api.access.redhat.com:443/r/insights
+#base_url=cert.cloud.redhat.com:443/api
 
 # URL for your proxy.  Example: http://user:pass@192.168.100.50:8080
 #proxy=
@@ -42,3 +42,7 @@
 
 # Location of the tags file for this system
 #tags_file=/etc/insights-client/tags.yaml
+
+# Use legacy API URLs
+# Warning: Legacy API is not supported in this RHEL version and may result in broken functionality.
+legacy_upload=False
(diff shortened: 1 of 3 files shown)


zpetrace (dismissed):
Seems good from the QE POV, the changes worked and pre-verification passed:)

data/insights-client.conf
```diff
@@ -42,3 +42,6 @@
 
 # Location of the tags file for this system
 #tags_file=/etc/insights-client/tags.yaml
+
+# Legacy HTTP configuration to perform the upload
+#legacy_upload=False
```
m-horky, line 47: Since we want to switch the value in all 10 installations, we'll have to uncomment this to take effect.

m-horky: I'd also change the description a bit.
```
# Use legacy API URLs
# Warning: Legacy API is not supported in this RHEL version and may result in broken functionality.
```

m-horky (changes requested):

data/insights-client.conf
pkoprda: Resolved

m-horky (approved):
Looks better. I'd like to get another verification from @zpetrace (see CCT card for details).

m-horky:
We'll likely need to go through different route, this change doesn't reliably flip the API. Closing.

PR #333, opened 2025-01-21
test: Enable Playbook Verifier tests on RHEL 10-like systems

m-horky:
* Card ID: CCT-1070

Since we have managed to resolve the problems the Verifier had in Python 3.12+, we can enable the integration tests for the playbooks shipped in this repository.

Since RHEL < 10 does not suffer from this issue, we don't need any backports.


ptoscano:
This likely needs an update of the egg shipped in the sources, otherwise the offline tests (eg those run via testing-farm) that run for EL 10 & Fedora fail:
```
integration-tests/playbook_verifier/test_verifier.py::test_official_playbook[bugs.yml] FAILED [  2%]
```

pkoprda (approved):
LGTM

PR #335, opened 2025-01-23
fix: Fix build after removing legacy collection

m-horky:


m-horky:
The Jenkins failure is caused by system teardown, the tests themselves have passed:
```
======== 135 passed, 1 xpassed, 2 subtests passed in 5606.64s (1:33:26) ========
```

PR #337, opened 2025-01-29
test: Status code indicates registration status

pkoprda:
* Card ID: CCT-1059

Status code returned by `insights-client --status` indicates whether a host is registered or not. This change updates the failing integration tests.

- `el9` (all of RHEL 9)
- `el8` (all of RHEL 8)

Since [insights-core#4344](https://github.com/RedHatInsights/insights-core/pull/4344) was merged, this PR is now ready for review.

### DIFF
--- a/integration-tests/test_motd.py
+++ b/integration-tests/test_motd.py
@@ -69,7 +69,7 @@ def test_motd(insights_client):
         5. The MOTD file is still not present
     """
     # If the system is not registered, the file should be present.
-    insights_client.run("--status")
+    insights_client.run("--status", check=False)
     assert os.path.exists(MOTD_PATH)
 
     # After registration, the file should not exist.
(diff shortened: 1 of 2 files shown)


integration-tests/test_motd.py
```diff
@@ -70,7 +70,8 @@ def test_motd(insights_client):
         5. The MOTD file is still not present
     """
     # If the system is not registered, the file should be present.
-    insights_client.run("--status")
+    with contextlib.suppress(subprocess.CalledProcessError):
+        insights_client.run("--status")
```
m-horky, line 74: This can be handled by `.run("--status", check=False)`.

m-horky (changes requested):

integration-tests/test_status.py
m-horky, line 1: Here we'll need to add Core version check, to account for both older and newer eggs:

```py
if insights_client.config.legacy_upload:
    ...  # can be kept as-is
else:
    if insights_client.core_version >= Version(3, 5, ...):
        assert registration_status.returncode == 1
    else:
        assert registration_status.returncode == 0
    assert "This host is unregistered.\n" == registration_status.stdout
```

m-horky (changes requested):

ptoscano (approved):
thanks!

zpetrace (approved):
This seems good, thanks for the update!

m-horky (approved):

PR #340, opened 2025-02-17
test: Check for bootc/image-mode deployments

Aftermath:
DNF commands without special options do not work in image-mode, and we will have pre-prepared images with needed packaging - so we will skip executing these bits for bootc images.

Card: CCT-1183


Lorquas (approved):
LGTM

PR #341, opened 2025-02-18
fix: display insights-client version when no egg is found

grunwmar:
Add print of insights-client version to 'if' block checking presence of valid eggs.

As whole try-except block is repeated from later code, this solution is a bit provisory.

Card ID: CCT-393

---

### DIFF
--- a/src/insights_client/__init__.py
+++ b/src/insights_client/__init__.py
@@ -431,7 +431,9 @@ def _main():
         valid_env_egg = []
 
     if not validated_eggs and not valid_env_egg:
-        sys.exit("No GPG-verified initial eggs can be found")
+        print("Client: %s" % InsightsConstants.version)
+        print("Core: not found")
+        return
 
     # ENV egg comes first
     all_valid_eggs = valid_env_egg + validated_eggs


tcnk-dev:
I would really appreciate for check if my logic of change was correct :)

src/insights_client/__init__.py
```diff
@@ -423,7 +423,16 @@ def _main():
...
+        try:
+            from insights_client.constants import InsightsConstants
+        except ImportError:
+            # The source file is build from 'constants.py.in'
+            # and is not available during development
+            class InsightsConstants(object):
+                version = "development"
+
+        print("Client: %s" % InsightsConstants.version)
+        sys.exit("No GPG-verified eggs can be found")
```
pkoprda, line 435: I think the expected output is (for development version):
```
Client: development
Core: not found
```
if there is no egg found on the system. So just change `sys.exit("No GPG-verified eggs can be found")` to:
```py
print("Core: not found")
return
```

pkoprda (changes requested):
Thank you @tcnk-dev! Looks good, just one nitpick.

And please rebase. Can you change the commit description, so it would say why did you do this change? Something like this:
```
Instead of showing only an error message, it should always display the client version while indicating that the core version is not found.
```
Also I would remove the sentence `As whole try-except block is repeated from later code, this solution is a bit provisory.`

src/insights_client/__init__.py
grunwmar: Thanks for check! :)

pkoprda: Now it prints correctly `Core: not found`, but there is no `return` after that, which may result in unexpected behavior. Can you add there `return` as I suggested in the previous comment?

pkoprda (approved):
LGTM, let's wait for QE

ptoscano:
Hmm there is a bit of code duplication here; I think we can avoid this, especially as fixing it was needed also for other work I had (see #317)

I just extracted that commit in its own, so it will be easier: please take a look at #353, thanks!

ptoscano (approved):
thanks! LGTM now

pkoprda (approved):

zpetrace (approved):
LGTM, verification passed so we can merge this:)

PR #342, opened 2025-02-18
feat: adding rhel 10 to main stream builds

Lorquas:
Adding rhel-10 to upstream builds to fix the PR jobs.


ptoscano:
Hey @Lorquas,

`targets` in build jobs are the distributions as available in copr, eg `rhel-N` -- see https://packit.dev/docs/configuration/upstream/copr_build#available-copr-build-targets

To enable RHEL in testing jobs, you need a bit more code: see the setup done in the `el9` branch.

PR #345, opened 2025-02-20
test: Unregistration should be always explicit

pkoprda:
* Card ID: CCT-1122

Previously, certain commands (e.g., `--status`) could unintentionally trigger system unregistration if the host was missing from the Inventory. This change adds the integration test which verifies that system unregistration does not happen implicitly.

---

- `el9` (all of RHEL 9)
- `el8` (all of RHEL 8)

Since [insights-core#4354](https://github.com/RedHatInsights/insights-core/pull/4354) was merged, this PR is now ready for review.

### DIFF
--- a/integration-tests/test_status.py
+++ b/integration-tests/test_status.py
@@ -9,11 +9,13 @@
 :upstream: Yes
 """
 
+import conftest
+from constants import REGISTERED_FILE, UNREGISTERED_FILE, MACHINE_ID_FILE
 import contextlib
+import os
 import pytest
 from pytest_client_tools.util import Version
 from time import sleep
-import conftest
 
 pytestmark = pytest.mark.usefixtures("register_subman")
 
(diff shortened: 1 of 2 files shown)


m-horky (approved):
LGTM

integration-tests/test_status.py
```diff
@@ -45,6 +47,29 @@ def test_status_registered(external_candlepin, insights_client):
         assert "This host is registered.\n" == registration_status.stdout
 
 
+def test_status_registered_only_locally(
+    external_candlepin, insights_client, external_inventory
+):
+    """
+    :title: Test insights-client --status when registered only locally
```
zpetrace, line 54: Hi, so the test looks good but there are some missing fields in the docstring here and that's why the testimony check fails, could you please add the rest of the docstrings? The necessary ones are in the failed testimony check:)

zpetrace (changes requested):

zpetrace (approved):
The tests failed on provisioning the beaker system, the changes are looking better and locally the test passes so I think we can merge it:)

PR #346, opened 2025-02-20
fix(test): Support running connection tests with proxy set 

m-horky:
When HTTP_PROXY environment variable is set, and is pointing at valid
proxy server, these two tests would fail because insights-client reports
different errors on network errors for proxy/non-proxy scenarios.

This code takes proxies into account, and asserts strings accordingly.

---

- `el9` (all of RHEL 9)
- `el8` (all of RHEL 8)

* Card ID: CCT-1199

### DIFF
--- a/integration-tests/test_connection.py
+++ b/integration-tests/test_connection.py
@@ -6,10 +6,41 @@
 :upstream: Yes
 """
 
+import configparser
+import contextlib
+import os
+import typing
+
 import pytest
+
+if typing.TYPE_CHECKING:
+    import pytest_client_tools.insights_client
 
 # the tests need a valid connection to insights so therefore the subman registration
 pytestmark = pytest.mark.usefixtures("register_subman")
+
+
+def _is_using_proxy(
+    insights_config: "pytest_client_tools.insights_client.InsightsClientConfig",
+) -> bool:
+    for key in os.environ.keys():
+        if key.lower() == "https_proxy":
+            return True
+
+    with contextlib.suppress(KeyError):
+        config_proxy = insights_config.proxy
+        if config_proxy != "":
+            return True
+
+    # sub-man fixture doesn't currently support reading the configuration file,
+    # let's parse it ourselves.
+    with contextlib.suppress(Exception):
+        rhsm_config = configparser.ConfigParser()
+        rhsm_config.read("/etc/rhsm/rhsm.conf")
+        if rhsm_config.get("server", "proxy_hostname", fallback=None) is not None:
+            return True
+
+    return False
 
 
 def test_connection_ok(insights_client):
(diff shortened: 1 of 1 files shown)


integration-tests/test_connection.py
```diff
@@ -6,13 +6,31 @@
...
+
 # the tests need a valid connection to insights so therefore the subman registration
 pytestmark = pytest.mark.usefixtures("register_subman")
 
 
-def test_connection(insights_client):
+def _is_using_proxy(
+    config: "pytest_client_tools.insights_client.InsightsClientConfig",
+) -> bool:
+    if os.getenv("HTTP_PROXY") is not None:
```
ptoscano, line 24: eh, I was setting a proxy in Unix would be this easy... I think it should be better to also check for `http_proxy` (lowercase) which (according to my experience) should be more commonly used

... and actually insights-core does not use `HTTP_PROXY` (which explicitly rejects), but `HTTPS_PROXY`:

https://github.com/RedHatInsights/insights-core/blob/23157b137ffa6ca3ee9f0d12ecf13e23f5f4a80d/insights/client/config.py#L553-L561

also, please keep into account that the proxy can be configured also in `rhsm.conf`, in case of autoconfig (which is true by default), so most likely you need to check that too, for added fun...

https://github.com/RedHatInsights/insights-core/blob/23157b137ffa6ca3ee9f0d12ecf13e23f5f4a80d/insights/client/connection.py#L334-L339

ptoscano (changes requested):

integration-tests/test_connection.py
```diff
@@ -6,12 +6,43 @@
...
+    for key in os.environ.keys():
+        if key.lower() == "https_proxy":
+            return True
+
+    try:
+        config_proxy = insights_config.proxy
+        if config_proxy != "":
+            return True
+    except KeyError:
+        return False
```
ptoscano, line 34: shouldn't this be a simple `pass` (or `contextlib.suppress()`)? otherwise a missing `proxy` entry in `insights-client.conf` (which is the default) will cause `KeyError` to be raised, and thus `False` to be returned unconditionally, without trying to read `rhsm.conf`

integration-tests/test_connection.py
```diff
@@ -6,12 +6,43 @@
...
+
+    # sub-man fixture doesn't currently support reading the configuration file,
+    # let's parse it ourselves.
+    rhsm_config = configparser.ConfigParser()
+    rhsm_config.read("/etc/rhsm/rhsm.conf")
+    if rhsm_config.get("server", "proxy_hostname", fallback=None) is not None:
+        return True
```
ptoscano, line 41: even if it's a very narrow corner case, I'd catch all the exceptions of `ConfigParser` and ignore them, to avoid stacktracing on an invalid `rhsm.conf` or in case the `server` section or the `proxy_hostname` key are missing

ptoscano (changes requested):
thanks for the improvements! there are a couple of changes needed to make this new code a bit safer

ptoscano (approved):
thanks!

PR #349, opened 2025-02-20
feat(ci): move downstream tests upstream

Lorquas:
As of last month, testing-farm supports [repo-configuration](https://docs.testing-farm.io/Testing%20Farm/0.1/rfd/rfd2-testing-farm-repo-config.html) and secrets, this unblocks us from running our integration tests fully upstream in gitlab. This will not only remove our dependency on jenkins for pull requests, but also allow us to stop using beaker. Beaker is the source of most of our test problems.

---

* Card ID: CCT-1196


Lorquas:
We should also consider if we want to run the full suite on more than RHEL-10-Nightly (full list of available composes [here](http://storage.tft.osci.redhat.com/composes-production.html)

zpetrace (approved):
LGTM, I have gone through the failed tests and none seems to be due to this change and are not related.

PR #350, opened 2025-02-24
chore(test): Update path of files collected in archive

m-horky:
The format of collected archive is being cleaned up, with most of the
files directly in the archive root being removed or moved.

File tags.json is present twice: at /tags.json and /data/tags.json. It's
been like that for a long time, we should be able to move the path of it
to always read /data/tags.json.

By replacing --output-file with --output-dir, we can omit manipulations
with the tar file.

---

- `el9` (all of RHEL 9)
- `el8` (all of RHEL 8)

* Card ID: CCT-1198
* Card ID: RHINENG-7581

### DIFF
--- a/integration-tests/test_client_options.py
+++ b/integration-tests/test_client_options.py
@@ -115,7 +115,7 @@ def test_group(insights_client, tmp_path):
     with tarfile.open(archive_location, "r") as tar:
         dir_name = tar.getnames()[0]
         for m in tar.getmembers():
-            if m.name == f"{dir_name}/tags.json":
+            if m.name == f"{dir_name}/data/tags.json":
                 tag_file = tar.extractfile(m)
                 tag_file_content = json.load(tag_file)
                 break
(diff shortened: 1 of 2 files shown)


integration-tests/test_client_options.py
```diff
@@ -104,21 +103,14 @@ def test_group(insights_client, tmp_path):
...
     )
 
-    with tarfile.open(archive_location, "r") as tar:
-        dir_name = tar.getnames()[0]
-        for m in tar.getmembers():
-            if m.name == f"{dir_name}/tags.json":
-                tag_file = tar.extractfile(m)
-                tag_file_content = json.load(tag_file)
-                break
+    with open(f"{tmp_path}/data/tags.json", "r") as f:
```
ptoscano, line 112: since this is a `pathlib.Path` now, let's use that way:
```python
with (tmp_path / "data" / "tags.json").open() as f:
   ...
```
alternatively, you can load the file at once:
```python
  ... = json.loads(that_path.read_text())
```

ptoscano (changes requested):
LGTM, one small thing to improve/simplify.

ptoscano (approved):
thanks!

PR #352, opened 2025-02-25
[el8] chore(test): Update path of files collected in archive

m-horky:
The format of collected archive is being cleaned up, with most of the
files directly in the archive root being removed or moved.

File tags.json is present twice: at /tags.json and /data/tags.json. It's
been like that for a long time, we should be able to move the path of it
to always read /data/tags.json.

By replacing --output-file with --output-dir, we can omit manipulations
with the tar file.

---

This pull request is a backport of: https://github.com/RedHatInsights/insights-client/pull/350

* Card ID: CCT-1198
* Card ID: RHINENG-7581

### DIFF
--- a/integration-tests/test_client_options.py
+++ b/integration-tests/test_client_options.py
@@ -8,7 +8,6 @@
 
 import json
 import os
-import tarfile
 import pytest
 import conftest
 import glob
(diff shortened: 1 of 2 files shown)


ptoscano (approved):
thanks!

PR #353, opened 2025-02-25
chore: move constants import earlier/global

ptoscano:
Import the 'constants' module earlier in a global context: this way more things can be added to it later.

There should be no behaviour changes.

This will help for #341 and #317.


ptoscano:
> Please open backports to el9 and el8 as well.

ACK, will do. I didn't note in the PR description because I wasn't sure it was needed.

PR #355, opened 2025-02-25
[el8] chore: move constants import earlier/global

ptoscano:
Import the 'constants' module earlier in a global context: this way more things can be added to it later.

There should be no behaviour changes.

This pull request is a backport of: #353


m-horky (approved):
LGTM

PR #356, opened 2025-02-26
docs: Remove reference to argument for --diagnosis parameter

pkoprda:
* Card ID: CCT-226

Since the argument for the --diagnosis parameter was removed from insights-core code, it is no longer relevant to keep the reference to it in the documentation.

---

- [x] `el9` (all of RHEL 9)
- [x] `el8` (all of RHEL 8)
- [x] `el7` (all of RHEL 7)


m-horky (approved):
LGTM

PR #359, opened 2025-02-26
[el9] docs: Remove reference to argument for --diagnosis parameter

pkoprda:
* Card ID: CCT-226

Since the argument for the --diagnosis parameter was removed from insights-core code, it is no longer relevant to keep the reference to it in the documentation.

---

This pull request is a backport of: #356


m-horky (approved):
LGTM

PR #363, opened 2025-03-04
test: Change diagnosis expected output when unregistered

pkoprda:
* Card ID: CCT-1195

When the host is not registered to insights-client, the output for diagnosis argument should not print out 404 error message. This change updates the failing integration test.

---

- `el9` (all of RHEL 9)
- `el8` (all of RHEL 8)

Since [insights-core#4366](https://github.com/RedHatInsights/insights-core/pull/4366) was merged, this PR is now ready for review.

### DIFF
--- a/integration-tests/test_client_options.py
+++ b/integration-tests/test_client_options.py
@@ -9,6 +9,7 @@
 import json
 import os
 import pytest
+from pytest_client_tools.util import Version
 import conftest
 import glob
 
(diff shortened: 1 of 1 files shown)


integration-tests/test_client_options.py
```diff
@@ -233,16 +233,17 @@ def test_client_diagnosis(insights_client):
...
         2. The client is registered
         3. The command retrieves diagnostic data and the output contains
             machine id
         4. The machine ID in the diagnostic data matches the system's machine id
     """
     # Running diagnosis on unregistered system returns appropriate error message
     diagnosis_result = insights_client.run("--diagnosis", check=False)
-    assert "Unable to get diagnosis data: 404" in diagnosis_result.stdout
+    assert diagnosis_result.returncode == 1
+    assert "Could not get diagnosis data." in diagnosis_result.stdout
```
m-horky, line 246: We probably need to add `if` for the Core version just like we have on other places. This would fail for the RPM Core.

m-horky (changes requested):

m-horky (approved):
LGTM

zpetrace (approved):
LGTM, thanks for the change:)

PR #364, opened 2025-03-05
fix(test): Pass test_phase_error_100 even if no egg exists no the host

m-horky:
* Card ID: CCT-1220

On systems where insights-client hasn't been run yet (e.g. pure container which only has the repository and test dependencies installed), the test fails because it never gets to the mocked section.

File existence and GPG validation in 'run_phase()' are only done if 'config.gpg' is set to True; mocking the attribute to read False skips these checks.

---

- `el9` (all of RHEL 9)
- `el8` (all of RHEL 8)
- `el7` (all of RHEL 7)

### DIFF
--- a/src/insights_client/tests/test_client.py
+++ b/src/insights_client/tests/test_client.py
@@ -1,11 +1,11 @@
-from unittest.mock import patch
+from unittest import mock
 import insights_client
 import pytest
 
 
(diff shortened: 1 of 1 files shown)


ptoscano (approved):
thanks!

PR #367, opened 2025-03-06
fix: display insights-client version when no egg is found

grunwmar:
Instead of showing only an error message, it should always display the client version while indicating  that the core version is not found.

* Card ID: CCT-393

---


zpetrace (approved):
The failures are due to the testware, not the change, so I think we can merge this backport:)

zpetrace (approved):
The failures are due to the testware, not the change, so I think we can merge this backport:)

pkoprda (approved):
LGTM

PR #372, opened 2025-03-07
[el9] test: Change diagnosis expected output when unregistered

pkoprda:
* Card ID: CCT-1195

When the host is not registered to insights-client, the output for diagnosis argument should not print out 404 error message. This change updates the failing integration test.

---

This pull request is a backport of: #363

### DIFF
--- a/integration-tests/test_client_options.py
+++ b/integration-tests/test_client_options.py
@@ -12,6 +12,7 @@
 import json
 import os
 import pytest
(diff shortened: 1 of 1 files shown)


zpetrace (approved):
Failures are due beaker system not being reserved but it's a backport so I think we can merge this:)

m-horky (approved):
ACK

PR #374, opened 2025-03-08
fix(test): Ensure that tests are part of evidence traceability matrix.

zpetrace:
Based on CCT-452 I have added the neccessary docstrings to the test classes and edited the testimony.yaml and custom betelgeuse config to accomodate these changes.
---

Also I am not sure, but I think we should backport it to all of our branches? So...

- `el9` (all of RHEL 9)
- `el8` (all of RHEL 8)
- `el7` (all of RHEL 7)

* Card ID: CCT-452


m-horky (approved):
Assuming we need Polarion for not just RHEL 10, but older versions, we do need to backport. In `el7` we don't have tests here on GitHub, so just into `el9` and `el8`.

pkoprda (approved):
LGTM

PR #376, opened 2025-03-10
[el8] test: Unregistration should be always explicit

pkoprda:
* Card ID: CCT-1122

Previously, certain commands (e.g., --status) could unintentionally trigger system unregistration if the host was missing from the Inventory. This change adds the integration test which verifies that system unregistration does not happen implicitly.

---

This pull request is a backport of: #345


zpetrace (approved):
Failures are not connected to the change done, it's also a backport so I think we can merge:)

m-horky (approved):
ACK

PR #378, opened 2025-03-12
feat(test): Adding tests for compliance option

zpetrace:
---
I have added 2 tests to test the --compliance and --compliance-policies  options. I also had to add some requirements as they were not on the system originally but compliance depends on them.

- `el9` (all of RHEL 9)
- `el8` (all of RHEL 8)

* Card ID: CCT-1229

### DIFF
--- a/integration-tests/test_compliance.py
+++ b/integration-tests/test_compliance.py
@@ -1,0 +1,93 @@
+"""
+:casecomponent: insights-client
+:requirement: RHSS-291297
+:polarion-project-id: RHELSS
+:polarion-include-skipped: false
+:polarion-lookup-method: id
+:subsystemteam: sst_csi_client_tools
+:caseautomation: Automated
+:upstream: Yes
+"""
+
+import conftest
+import pytest
+import distro
+
+pytestmark = [
+    pytest.mark.usefixtures("register_subman"),
+    pytest.mark.skipif(
+        distro.id() != "rhel",
+        reason="Skipping tests because OS ID is not RHEL",
+        allow_module_level=True,
+    ),
+]
+
+
+def test_compliance_option(insights_client):
+    """
+    :id: caa8b3e1-9347-494c-a1f5-1fa670136834
+    :title: Test compliance option
+    :reference: https://issues.redhat.com/browse/CCT-1229
+    :description:
+        This test verifies that running the --compliance will not result in a failure
+    :tags: Tier 1
+    :steps:
+        1. Run insights-client with --compliance option
+        2. Register insights-client
+        3. Run insights-client with --compliance option
+    :expectedresults:
+        1. Command will fail with instructions for user to register
+        2. System is successfully registered
+        3. The output of the command either informs user that system is not associated
+            with any policies or the report will be successfully uploaded
+    """
+    compliance_before_registration = insights_client.run("--compliance", check=False)
+    assert compliance_before_registration.returncode == 1
+    assert (
+        "This host has not been registered. Use --register to register this host"
+        in compliance_before_registration.stdout
+    )
+
+    insights_client.register()
+    assert conftest.loop_until(lambda: insights_client.is_registered)
+
+    compliance_after_registration = insights_client.run("--compliance", check=False)
+    if compliance_after_registration.returncode == 1:
+        assert (
+            "System is not associated with any policies."
+            in compliance_after_registration.stdout
+        )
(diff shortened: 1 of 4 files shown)


integration-tests/test_compliance.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+def test_compliance_policies_option(insights_client):
+    """
+    :id: ad3a2073-3a2e-485e-bc7b-fede2111a06a
+    :title: Test compliance-policies option
+    :reference: https://issues.redhat.com/browse/CCT-1229
+    :description:
+        This test verifies that running the --compliance-policies
+            will not result in a failure
```
pkoprda, line 64: Remove the indentation from the 2nd line of description
```suggestion
    :description:
        This test verifies that running the --compliance-policies
        will not result in a failure
```

integration-tests/test_compliance.py
```diff
@@ -0,0 +1,82 @@
...
+    :tags: Tier 1
+    :steps:
+        1. Register insights-client
+        2. Run insights-client with --compliance-policies option
+    :expectedresults:
+        1. System is successfully registered
+        2. The output of the command informs user that
+            system ystem is not assignable to any policy
```
pkoprda, line 72: `ystem` is probably not expected here :)

integration-tests/test_compliance.py
```diff
@@ -0,0 +1,82 @@
...
+    """
+    insights_client.register()
+    assert conftest.loop_until(lambda: insights_client.is_registered)
+
+    compliance_policies = insights_client.run("--compliance-policies", check=False)
+    assert "System is not assignable to any policy." in compliance_policies.stdout
```
pkoprda, line 78: Maybe I would add here some check before running `--compliance-policies` that there is no compliance policy available. I tried to run this test locally and it found some compliance-policy and therefore failed on ` assert "System is not assignable to any policy." in compliance_policies.stdout`

pkoprda (approved):
I think these tests can be run on centos or rhel systems, on Fedora systems it may fail. So add `pytest.skip` if it is run on fedora system. Also what about `--compliance-assign` and `--compliance-unassign` options? Can you add to test which would test these options please?

pkoprda (changes requested):
I accidentally checked `Approve` instead of `Request changes` for the above comment

zpetrace:
@pkoprda  so I have added the skip if not rhel exception for both of the tests but to the other functionality..we are not supposed to be (from what I understood) testing the whole functionality of compliance, these tests are supposed to tell us if we were to introduce some regression that something doesn't work. The functionality of compliance will be tested by compliance/core team.

integration-tests/test_compliance.py
```diff
@@ -0,0 +1,92 @@
...
+    compliance_policies = insights_client.run("--compliance-policies", check=False)
+    if compliance_policies.returncode == 1:
+        assert "System is not assignable to any policy." in compliance_policies.stdout
+        assert (
+            "Create supported policy using the Compliance web UI."
+            in compliance_policies.stdout
+        )
+    else:
+        # Checking only if some policy is available - ID will be available if yes
+        assert "ID" in compliance_policies.stdout
```
pkoprda, line 92: I would rather see this comment in the `:expectedresults:` in the function docstring.
Also rephrase it a little. Something like: `Verify that at least one policy is available - if so, ID must be present`

pkoprda (changes requested):
Looks better, the tests were passing locally for me. I just left some nitpicks.
Also please rebase

integration-tests/test_compliance.py
```diff
@@ -0,0 +1,93 @@
...
+        )
+    else:
+        assert "Successfully uploaded report" in compliance_after_registration.stdout
+
+
+def test_compliance_policies_option(insights_client):
+    """
+    :id: ad3a2073-3a2e-485e-bc7b-fede2111a06a
+    :title: Test compliance-policies option
+    :reference: https://issues.redhat.com/browse/CCT-1229:>
```
pkoprda, line 68:
```suggestion
    :reference: https://issues.redhat.com/browse/CCT-1229
```

zpetrace: whoops

pkoprda (approved):
LGTM, thank you @zpetrace!

m-horky (approved):
ACK

PR #381, opened 2025-03-12
[el8] test: Adding tests for compliance option

zpetrace:
I have added 2 tests to test the --compliance and --compliance-policies options. I also had to add some requirements as they were not on the system originally but compliance depends on them.

CCT Card: CCT-1229

---

This pull request is a backport of: https://github.com/RedHatInsights/insights-client/pull/378

### DIFF
--- a/systemtest/tests/integration/test.sh
+++ b/systemtest/tests/integration/test.sh
@@ -5,16 +5,20 @@ set -x
 # get to project root
 cd ../../../
 
-# Check for GitHub pull request ID and install build if needed.
-# This is for the downstream PR jobs.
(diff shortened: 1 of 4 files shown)


systemtest/tests/integration/test.sh
m-horky, line 1: Can you please split this off into a separate commit, and reference the hash of the original change done in main? This isn't in scope of CCT-1229.

m-horky (changes requested):
Looks good, except for mixing in the bootc changes.

zpetrace:
I have made the wrong cherry-pick it seems, I will close this and create a new and hopefully a correct backport PR.

PR #382, opened 2025-03-21
[el8] fix(test): Adding debug options to the test_common_specs

zpetrace:
As part of the work done for CCT-1107 I have added the debugging options for this test to be able to catch error more efficiently in case of this test failing. We will now be able to also see why the test was actually failing and not just the error from the spec JSON. 

As a minor change I have also tried to reduce the http_timeout to make this test functional.

---

- `el9` (all of RHEL 9)

 Card ID: CCT-1107

### DIFF
--- a/integration-tests/test_common_specs.py
+++ b/integration-tests/test_common_specs.py
@@ -12,6 +12,8 @@
 import json
 import os
 import pytest
+import subprocess
+import logging
 
 pytestmark = pytest.mark.usefixtures("register_subman")
 
@@ -82,5 +84,30 @@ def test_common_specs(insights_client, tmp_path):
         # (unless we are in container and the spec is known to not work in containers)
         if in_container and spec in privileged_specs:
             continue
-        assert not data["errors"], f"'{spec}' contains errors: {data['errors']} "
-        assert data["results"] is not None, f"'{spec}' does not contain results"
+
+        try:
+            assert not data["errors"], f"'{spec}' contains errors: {data['errors']} "
+            assert data["results"] is not None, f"'{spec}' does not contain results"
+        except AssertionError as e:
+            logging.error(f"Spec '{spec}' failed with an error '{str(e)}'")
+
+            # Try to extract and re-run the command from the spec JSON
+            cmd = data["results"]["object"]["cmd"]
+            if cmd:
+                logging.debug(f"Attempting to re-run the cmd from the spec: '{cmd}'")
+                result = subprocess.run(
+                    cmd, shell=True, capture_output=True, text=True, timeout=10
+                )
+
+                logging.debug(f"Command return code: {result.returncode}")
+                logging.debug(f"Command STDOUT code: {result.stdout.strip()}")
+                logging.debug(f"Command STDERR code: {result.stderr.strip()}")
+
+                if result.returncode != 0:
+                    raise AssertionError(
+                        f"Command '{cmd}' failed with a return code \
+                                         of {result.returncode}"
+                    )
+            else:
+                logging.debug(f"No command found in spec JSON for {spec}")
+            raise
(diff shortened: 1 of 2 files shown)


integration-tests/test_connection.py
```diff
@@ -96,7 +96,7 @@ def test_http_timeout(insights_client):
         3. The command fails with a return code of 1
         4. The output mentions timeout value
     """
-    insights_client.config.http_timeout = 0.1
+    insights_client.config.http_timeout = 0.001
```
pkoprda, line 99: Do this change also on the [line 108](https://github.com/RedHatInsights/insights-client/blob/be23d5ea887eb6704ff2820411caf2f2a88bf42d/integration-tests/test_connection.py#L108) in this file.

integration-tests/test_common_specs.py
```diff
@@ -81,5 +82,24 @@ def test_common_specs(insights_client, tmp_path):
         # (unless we are in container and the spec is known to not work in containers)
         if in_container and spec in privileged_specs:
             continue
-        assert not data["errors"], f"'{spec}' contains errors: {data['errors']} "
-        assert data["results"] is not None, f"'{spec}' does not contain results"
+
+        try:
+            assert not data["errors"], f"'{spec}' contains errors: {data['errors']}"
+        except AssertionError:
+            print(f"\n[DEBUG] Spec '{spec}' failed  with error: {data['errors']}")
```
pkoprda, line 89: Can we use [logging module](https://docs.python.org/3/howto/logging.html) instead of printing out this message and messages bellow?

integration-tests/test_common_specs.py
```diff
@@ -81,5 +82,24 @@ def test_common_specs(insights_client, tmp_path):
...
+            cmd = data["results"]["object"]["cmd"]
+            if cmd:
+                print(f"[DEBUG] Attempting to re-run the cmd from the spec: {cmd}")
+                result = subprocess.run(
+                    cmd, shell=True, capture_output=True, text=True, timeout=10
+                )
+
+                print(f"[DEBUG] Command return code: {result.returncode}")
+                print(f"[DEBUG] Command STDOUT code: {result.stdout}")
+                print(f"[DEBUG] Command STDERR code: {result.stderr}")
```
pkoprda, line 101: Shouldn't be here check for `assert result.returncode == 0`?

pkoprda (changes requested):

integration-tests/test_common_specs.py
zpetrace: I would really like to see it printed to the console if it fails for faster debugging and reporting of issues in case of some.

pkoprda: Yeah, but I think it would not be visible in the testing farm, because we run the tests like this:
```
pytest --log-level debug --junit-xml=./junit.xml -v integration-tests
```
and these print messages would be visible only if the test fail.

Or otherwise rewrite the code a little, because it looks confusing that the command passes but then there is a `raise` statement which fails the test.

pkoprda (changes requested):

integration-tests/test_common_specs.py
pkoprda: Yeah, but I think it would not be visible in the testing farm, because we run the tests like this:
```
pytest --log-level debug --junit-xml=./junit.xml -v integration-tests
```
and these print messages would be visible only if the test fail.

Or otherwise rewrite the code a little, because it looks confusing that the command passes but then there is a `raise` statement which fails the test.

zpetrace:
I have made the changes and made it a logging instead of print, pls check @pkoprda :)

zpetrace:
it seems that it fixed the http_timeout
`integration-tests/test_connection.py::test_http_timeout PASSED           [ 32%]`

and test_common_specs also seems like it's working with the changes so if it breaks next time we should know why
`integration-tests/test_common_specs.py::test_common_specs PASSED         [ 29%]`

pkoprda (changes requested):
Looks good, just a nitpick :)

PR #383, opened 2025-03-21
feat(test): Adding pytest markers for tiers

zpetrace:
As a part of the work for CCT-1247 I have added the pytest tier marks to the tests to be able to run tests from integration-tests based on the marks. I have registered the mark in the pytest.ini also.

---

- `el9` (all of RHEL 9)
- `el8` (all of RHEL 8)
- `el7` (all of RHEL 7)

* Card ID: RHEL-xxxx
Card ID: CCT-1247


pkoprda (changes requested):
Looks good, just update the commit message, so it will have reference to Jira card CCT-1247.

pkoprda (approved):
LGTM, thank you @zpetrace! The failing tests are unrelated to this change, we can merge this

PR #387, opened 2025-03-25
fix(test): Adding docstrings and pytest tier marks to test_verifier.py

zpetrace:
I have forgot to add the latest changes - adding the traceability docstrings and pytest tier marks into the playbook-verifier test. In this PR I'm adding these changes to test_verifier.py as an effort for CCT-1247.

---

- `el8` (all of RHEL 8)

For RHEL9 I will add these changes into the open PR so there will be no need to backporting this to the RHEL9.

Card ID: CCT-1247

### DIFF
--- a/integration-tests/playbook_verifier/test_verifier.py
+++ b/integration-tests/playbook_verifier/test_verifier.py
@@ -1,7 +1,17 @@
+"""
+:casecomponent: insights-client
+:requirement: RHSS-291297
+:polarion-project-id: RHELSS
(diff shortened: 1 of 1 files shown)


pkoprda (changes requested):
Looks good, just do a simple rewording of the description.

Also can you split the commit into two? One should include the addition of traceability docstring and the other one should include pytest tie mark and changed docstring of `test_official_playbook` function.

pkoprda (approved):
LGTM, thanks!

PR #388, opened 2025-03-25
fix(test): Extending duration of test runs

zpetrace:
Some of the test run failed on the timeout, so I created this PR extending the duration of the test runs to 3 hours instead of 2 to prevent failures on timeout.

---

- `el9` (all of RHEL 9)
- `el8` (all of RHEL 8)


ptoscano (approved):
The few failures are unrelated to this timeout change, and will need to be debugged separately.

PR #394, opened 2025-04-01
[el8] fix(test): Adding docstrings and pytest tier marks to test_verifier.py

zpetrace:
Adding the traceability docstrings and pytest tier marks into the playbook-verifier test. In this PR I'm adding these changes to test_verifier.py as an effort for CCT-1247.

cherry-picked from: dd4855dd850ea6ff9e8aba759fdbf5edbdfa9179 and 3bbce1e440f706b67e4ae068c680e9debd5c1529

---
This pull request is a backport of: https://github.com/RedHatInsights/insights-client/pull/387/

### DIFF
--- a/integration-tests/playbook_verifier/test_verifier.py
+++ b/integration-tests/playbook_verifier/test_verifier.py
@@ -1,7 +1,17 @@
+"""
+:casecomponent: insights-client
+:requirement: RHSS-291297
+:polarion-project-id: RHELSS
+:polarion-include-skipped: false
+:polarion-lookup-method: id
+:subsystemteam: sst_csi_client_tools
+:caseautomation: Automated
+:upstream: Yes
+"""
+
 import pathlib
 import subprocess
 import sys
-
 import pytest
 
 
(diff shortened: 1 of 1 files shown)


integration-tests/playbook_verifier/test_verifier.py
```diff
@@ -19,15 +30,34 @@
...
     """
+    if (
+        sys.version_info >= (3, 12)
+        and insights_client.core_version < Version(3, 5, 2)
+        and filename == "bugs.yml"
+    ):
+        pytest.xfail(
+            f"Core {insights_client.core_version} suffers from "
+            "CCT-1065, CCT-1101, CCT-1102."
+        )
```
pkoprda, line 59: I think we don't have to do this in `el8` branch. The test file `bugs.yml` was only aimed for RHEL10 (see CCT-1070) and also Python3.12 is not default for RHEL8.

pkoprda (changes requested):
I think the first commit should only include the traceability docstring for Polarion and docstring of `test_official_playbook` function. The second commit should have the addition of pytest tier for this function.
Also please avoid starting the commit message with `[el8]`. Since this is the `el8` branch, adding that prefix is unnecessary.

pkoprda (approved):
LGTM, let's just way for the pipeline to finish, then we can merge this

PR #397, opened 2025-04-01
Drop unused runtime requirement on setuptools (distutils)

hroncok:
---

### DIFF
--- a/insights-client.spec
+++ b/insights-client.spec
@@ -25,7 +25,6 @@ Requires: pciutils
 Requires: python3-requests >= 2.6
 Requires: python3-PyYAML
 Requires: python3-six
-Requires: python3dist(setuptools)
 Requires: coreutils
 Requires: insights-core >= 3.6.7
 
--- a/src/insights_client/tests/requirements.txt
+++ b/src/insights_client/tests/requirements.txt
@@ -1,5 +1,2 @@
 pytest
-# Setuptools is required because we still rely on 'distutils' which have been dropped in Python 3.12.
-# setuptools package ships it vendorized as a top-level package, so for a time being we can rely on it.
-setuptools
 -r requirements-core.txt


m-horky:
Hi. There is runtime dependency through `distutils` which setuptools vendorizes, since upstream Python dropped it. Both in `src/insights_client/__init__.py` and in Insights Core's `insights/client/client.py`.

m-horky:
Long-term, yes, but it hasn't been on our priority list. Are there plans to get rid of this component in fc42+/c11s/el11?

hroncok:
I want to move setuptools to CRB for c11s. At this point I am looking for trivial runtime-dependnecy removals (and I thought this is one). This one isn't trivial, but looks easy enough. I'll update this PR.

hroncok:
I've pushed my attempt here. If you need me to, I can open a RHEL card (what is it?), but I'd rather not :)

hroncok:
Note that comparing tuples with inst and string in them can lead to `TypeError: '<' not supported between instances of 'str' and 'int'` but that problem pre-existed in distutils:

```pycon
>>> from distutils.version import LooseVersion
>>> LooseVersion("1a1") < LooseVersion("1.0")
  File "<stdin>", line 1, in <module>
  File "/usr/lib64/python3.6/distutils/version.py", line 52, in __lt__
    c = self._cmp(other)
  File "/usr/lib64/python3.6/distutils/version.py", line 337, in _cmp
    if self.version < other.version:
TypeError: '<' not supported between instances of 'str' and 'int'

```

m-horky (approved):
Sure, this will probably work. I'll let @zpetrace do the QE verification pass.

zpetrace:
Hi, sorry for the delay on this. The QE tests passed for this changes so I think we are good to go and merge this:) @m-horky

m-horky:
Cool. Before we merge this, we'll need to do the same in [Insights Core](https://github.com/RedHatInsights/insights-core/). It's released async on top of insights-client, and would start failing as soon as this was released somewhere.

There's a downstream tracker RHINENG-6913 -- should I try to get it prioritized? We need to block this PR on resolving it there as well.

hroncok:
This is a priority for me, but it's not urgent.

hroncok:
For the record, I don't understand why [Insights Core](https://github.com/RedHatInsights/insights-core/) needs to ship the same change at the same time. It's not like this code exposed the no longer used distutils class as an API. It only uses it to compare versions (and the results of such comparisons are still the same).

m-horky:
Tracking update: this is slowly being worked on in background in downstream trackers. RHINENG-6913 has been finished, CCT-357 and CCT-1528 are pending.

hroncok:
Cool. See my https://github.com/RedHatInsights/insights-core/pull/4500 as well.

PR #399, opened 2025-04-03
docs: --status checks the presence of .registered file

pkoprda:
* Card ID: CCT-869

Since the status command checks only the presence of .registered file without calling Insights API, documentation must be updated accordingly.

---

- `el9` (all of RHEL 9)
- `el8` (all of RHEL 8)

Keeping this PR as a draft until [insights-core#4388](https://github.com/RedHatInsights/insights-core/pull/4388) will be merged.


zpetrace (approved):
The changes look good and the verification passed so IMO we can merge:)

PR #400, opened 2025-04-03
[el9] docs: --status checks the presence of .registered file

pkoprda:
* Card ID: CCT-869

Since the status command checks only the presence of .registered file without calling Insights API, documentation must be updated accordingly.

(cherry picked from commit f9044ac00aeae0eca0cec47646986f2a1ec65e3f)

---
This pull request is a backport of: #399


zpetrace (approved):
The changes look good and the verification passed so IMO we can merge:)

PR #403, opened 2025-04-04
fix(test): Adding debug options to the test_common_specs

zpetrace:
As part of the work done for CCT-1238 I have added the debugging options for this test to be able to catch error more efficiently in case of this test failing. We will now be able to also see why the test was actually failing and not just the error from the spec JSON. As a minor change I have also tried to reduce the http_timeout to make this test functional.

cherry-picked from e4f09a963ae2a45063edf77be694634b479e1666

---
This pull request is a backport of: https://github.com/RedHatInsights/insights-client/pull/382

* Card ID: CCT-1238

### DIFF
--- a/integration-tests/test_common_specs.py
+++ b/integration-tests/test_common_specs.py
@@ -12,6 +12,8 @@
 import json
 import os
 import pytest
+import subprocess
+import logging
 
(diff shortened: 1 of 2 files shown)


pkoprda (approved):
LGTM

PR #405, opened 2025-04-10
[el8] fix(test): Fixing test_tags

zpetrace:
test_tags was failing due to not having enough time for the changes to show in inventory. I have added the sleep function to wait 60 seconds before trying to get the inventory data back which then passed for me locally. This is a part of an effort for CCT-1107.

---

- `el9` (all of RHEL 9)
- `master`

Card ID: CCT-1107


pkoprda (approved):
LGTM, thank you!

PR #410, opened 2025-04-30
[el9] feat: optional SELinux policy for insights-core

m-horky:
insights-core will get soon an own SELinux policy, and thus insights-client must properly run insights-core using its own policy.

To do that:
- add a new meson option to specify the type of the SELinux policy for insights-core; an empty string disables the switch, i.e. keep the current behaviour (insights-core running in the same context of insights-client)
- use the libselinux Python binding in case a SELinux policy for insights-core is specified
- detect whether SELinux is enabled, and disable the switch if not
- in case a switch has to happen, then use the existing SELinux context, and apply the different type for insights-core

Card: CCT-984

(cherry picked from commit 169eb504000f45367bb9ea661d9b8dc1e718b08b)

---

This pull request is a backport of: https://github.com/RedHatInsights/insights-client/pull/317


pkoprda (approved):
LGTM from dev side

zpetrace (approved):
Based on the verification I'd say it looks good, the comment is in the corresponding CCT card:)

PR #411, opened 2025-05-02
[el8] fix(test): Fix the rest of RHEL8 tests

zpetrace:
Updated the test_no_obfuscation_on_package_version to avoid attempting
to read directories from the archive, which was causing an AttributeError.
The test now checks only regular files and supports recursively verifying
files within directories like modules.d/. This ensures all relevant package files
are validated without errors. I also fixed test_check_results by changing the order
of the actions to give it a little time to reflect the ssh change later. The
test_display_name_option now skips if the error is SSL related as that is
a server issue and not test issue. This PR also fixes the registered only locally
test.

---

Card ID: CCT-1107

### DIFF
--- a/integration-tests/test_client_options.py
+++ b/integration-tests/test_client_options.py
@@ -287,23 +287,23 @@ def test_check_show_results(insights_client):
     :reference:
     :tags: Tier 1
     :steps:
-        1. Register insights-client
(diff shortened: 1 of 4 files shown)


pkoprda (commented):
This looks good, can you just modify the commit message so it would match the changes that you did?

pkoprda (approved):
Looks good, thank you, we are green now :green_apple:

PR #415, opened 2025-05-15
fix(test): Fixing betelgeuse docstrings

zpetrace:
While uploading to Polarion I have figured that some of the docstrings we are using are wrong - I have changed the sst group and when the test have parameters I have added :parametrized:. I have added this ield also to testimony.yml so that we are checking for this field also. Another part of effort for CCT-1090.

---

- `el9` (all of RHEL 9)
- `el8` (all of RHEL 8)

Card ID: CCT-1090

## Summary by Sourcery

Update test metadata and schema to correct subsystem team identifiers and support parameterized tests documentation

Enhancements:
- Standardize the :subsystemteam: tag from 'sst_csi_client_tools' to 'rhel-sst-csi-client-tools' across all integration tests
- Add a new ':parametrized:' docstring field set to 'yes' for tests with parameters
- Extend testimony.yml to include the optional 'Parametrized' field and update SubSystemTeam choices accordingly

### DIFF
--- a/integration-tests/playbook_verifier/test_verifier.py
+++ b/integration-tests/playbook_verifier/test_verifier.py
@@ -4,7 +4,7 @@
 :polarion-project-id: RHELSS
 :polarion-include-skipped: false
 :polarion-lookup-method: id
-:subsystemteam: sst_csi_client_tools
+:subsystemteam: rhel-sst-csi-client-tools
 :caseautomation: Automated
 :upstream: Yes
 """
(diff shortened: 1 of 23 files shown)


pkoprda (changes requested):
Looks good, can you just change commit description, there is a typo:
```
(...) I have added this ield also to testimony.yml so that we are (...)
```
`ield` should probably be `field`.

pkoprda (approved):
LGTM

PR #417, opened 2025-05-20
fix(scripts): Do not pull egg on 4xx/5xx

m-horky:
## Summary by Sourcery

Bug Fixes:
- Add --fail flag to curl commands in 01-upgrade-egg.sh to abort on HTTP 4xx/5xx errors and avoid pulling invalid files


pkoprda (approved):
LGTM

PR #416, opened 2025-05-20
chore: Release insights-client-v3.10.1

m-horky:
- Updates Core to 3.5.13

(Trying to open a MR from not-a-fork, since https://github.com/RedHatInsights/insights-client/pull/412 is failing)

## Summary by Sourcery

Prepare the 3.10.1 release by updating the project version, adjusting the egg retrieval script to a new endpoint with redirect support, refreshing the signature file, and cleaning up unused files.

Enhancements:
- Update upgrade script to use the console.redhat.com static release API and follow redirects

Build:
- Bump project version to 3.10.1 in meson.build

Chores:
- Replace RPM egg PGP signature for the new release
- Remove obsolete EGG_VERSION file

PR #418, opened 2025-05-20
[el9] chore: Release insights-client-v3.9.0

m-horky:
- Updates Core to 3.5.13

This pull request is a backport of: #416 #417 

Deprecates PR #413


zpetrace:
Only the expected failures showed up so I think we are good to go and merge this:)

PR #421, opened 2025-05-22
fix(test): Installing RPM builds for downstream Gating Jobs.

Archana-PandeyM:
This change affects downstream gating jobs only.
Currently jenkins uses tmt --last login option to install these builds during prepare phase, due to an issue #3716 in tmt our gate jobs are failing. With this change we will remove dependency on --last login and install the brew build directly.

---

## Summary by Sourcery

Fix downstream Jenkins gating jobs by bypassing tmt’s --last login and installing RPM builds directly from the TEST_RPMS environment variable.

Bug Fixes:
- Remove reliance on tmt’s --last login to install builds, preventing gating job failures.

Tests:
- Update integration test script to install RPMs via dnf when TEST_RPMS is defined.

### DIFF
--- a/systemtest/tests/integration/test.sh
+++ b/systemtest/tests/integration/test.sh
@@ -20,6 +20,13 @@ if ! command -v bootc >/dev/null || bootc status | grep -q 'type: null'; then
     scap-security-guide openscap-scanner openscap bzip2-devel
 fi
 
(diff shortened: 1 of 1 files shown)


Lorquas (approved):
Looks good to me, has this been tested in a run?

PR #423, opened 2025-05-22
fix(test): Fixing the rest of integration-tests for RHEL10

zpetrace:
As a final effort for the CCT-1238 I tried fixing the last failing test. I have fixed test_display_name_disable_autoconfig_and_autoupdate based on the changes we have made in RHEL8 and 9 where the test worked after these changes.

---

Card ID: CCT-1238

## Summary by Sourcery

Tests:
- Skip test_display_name_disable_autoconfig_and_autoupdate upon SSL certificate verification failures instead of erroring out

### DIFF
--- a/integration-tests/test_display_name_option.py
+++ b/integration-tests/test_display_name_option.py
@@ -16,6 +16,7 @@ import logging
 import json
 import random
 import conftest
+import subprocess
 
 from constants import HOST_DETAILS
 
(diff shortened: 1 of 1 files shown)


pkoprda (approved):
This looks good, we can merge this :rocket: 
In the future, we should investigate why the tests are failing for Fedora

PR #427, opened 2025-06-03
fix(test): Extending duration of the tests runs

zpetrace:
Some of the architectures like ppc64le or s390x are running on emulated systems - they are running on x86_64 with e.g. s390x emulated. This slows the execution of the tests which then usually fail on the timeout. Until that is solved and we have other arches available without having to emulate them we need to extend the duration of the test runs.

---

- `el9` (all of RHEL 9)
- `el8` (all of RHEL 8)

### DIFF
--- a/systemtest/tests/integration/main.fmf
+++ b/systemtest/tests/integration/main.fmf
@@ -1,3 +1,3 @@
 summary: Runs tmt tests
 test: ./test.sh
-duration: 3h
+duration: 5h


pkoprda (approved):
LGTM, we can merge this. Please create also another PR with the fix for these failing integration tests:
```
FAILED integration-tests/test_obfuscation.py::test_ip_obfuscation
FAILED integration-tests/test_obfuscation.py::test_hostname_obfuscation
FAILED integration-tests/test_registration.py::test_register_auth_proxy
```

PR #430, opened 2025-06-06
chore: Stop shipping .exp.sed file

pkoprda:
* Card ID: CCT-1204

Since data obfuscation is done by Core now, the .exp.sed file is no longer needed to be present in this repository.

---

- `el9` (all of RHEL 9)


zpetrace (approved):
The obfuscation failing tests are being fixed in the latest core egg release and for test_check_show_results there is a PR created (https://github.com/RedHatInsights/insights-client/pull/432). The playbooks failing in Fedora will need to be worked on in the future too. Therefore as of now it looks good from a QE side.

m-horky (approved):
Patch LGTM

PR #432, opened 2025-06-06
fix(test): Fixed test_check_show_results

zpetrace:
In order for insights-client to reliably check the correct results from Advisor we needed to change the order of action in the test. The permission of the file is changed before registering to insights so that there is enough time for the issue to be reflected in Advisor.

---

- `el9` (all of RHEL 9)
- `el8` (all of RHEL 8)

## Summary by Sourcery

Tests:
- Move os.chmod in test_check_show_results to before insights_client.register and remove redundant chmod and client.run call


pkoprda (changes requested):
Looks good, just adjust docstring for this test (the order of steps and expected results)

PR #437, opened 2025-06-11
test: Adding tests for insights-client-ros

zpetrace:
I have added multiple tests for insights-client-ros. I am testing all the basic functionality of ROS - these tests test that we can download the subpackage and activate the pmlogger. Then we test if the latest metrics exists once the subpaclkage is started. Then they test basic functionality of insights while ROS is installed and active and lastly that we can uninstal the subpackage and the pmlogger is inactive again. This is an effor based on the CCT-1402.

---

- `el9` (all of RHEL 9)
- `el8` (all of RHEL 8)

* Card ID: CCT-1402

### DIFF
--- a/integration-tests/test_ros.py
+++ b/integration-tests/test_ros.py
@@ -1,0 +1,189 @@
+"""
+:component: insights-client-ros
+:requirement: RHSS-291297
+:polarion-project-id: RHELSS
+:polarion-include-skipped: false
+:polarion-lookup-method: id
+:poolteam: rhel-sst-csi-client-tools
+:caseautomation: Automated
+:upstream: Yes
+"""
+
+import pytest
+import subprocess
+import conftest
+import shutil
+from constants import CONFIG_FILE
+from pathlib import Path
+
+pytestmark = pytest.mark.usefixtures("register_subman")
+
+PACKAGE = "insights-client-ros"
+SERVICE = "pmlogger"
+
+
+@pytest.mark.tier1
+def test_ros_install():
+    """
+    :id: 3fc19957-be97-429d-817a-610e6c69dd9d
+    :title: Verify insights-client-ros can be installed
+    :description:
+        Ensure that the insights-client-ros can be successfully installed
+    :tags: Tier 1
+    :steps:
+        1. Try to install insights-client-ros
+        2. Check that ros_collect is set to true in insights-client.conf
+    :expectedresults:
+        1. Subpackage insights-client-ros is installed
+        2. Field ros_collect is set to true in insights-client.conf
+    """
+    install = subprocess.run(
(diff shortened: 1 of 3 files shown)


integration-tests/test_ros.py
```diff
@@ -0,0 +1,186 @@
...
+    :id: 3fc19957-be97-429d-817a-610e6c69dd9d
+    :title: Verify insights-client-ros can be installed
+    :description:
+        Ensure that the insights-client-ros can be successfully installed
+    :tags: Tier 1
+    :steps:
+        1. Try to install insights-client-ros
+        2. Check that ros_collect is set to true in insights-client.conf
+    :expectedresults:
+        1. Insights-client-ros is installed.
```
pkoprda, line 36: IMO lowercase better here
```suggestion
        1. insights-client-ros is installed.
```

zpetrace: It is better when the steps and expected results start with a capital letter so I have changed it a bit so that the name can be lowecased.

integration-tests/test_ros.py
```diff
@@ -0,0 +1,187 @@
...
+    :title: Verify insights-client-ros can be installed
+    :description:
+        Ensure that the insights-client-ros can be successfully installed
+    :tags: Tier 1
+    :steps:
+        1. Try to install insights-client-ros
+        2. Check that ros_collect is set to true in insights-client.conf
+    :expectedresults:
+        1. Subpackage insights-client-ros is installed.
+        2. ros_collect is set to true in insights-client.conf
```
pkoprda, line 38: Just a nitpick, to be consistent, remove the ending dot in the 1st expected results. Also `ros_collect` should start with a capital letter according to [your comment](https://github.com/RedHatInsights/insights-client/pull/437#discussion_r2145086104):
```suggestion
        1. Subpackage insights-client-ros is installed
        2. ros_collect is set to true in insights-client.conf
```

integration-tests/test_ros.py
```diff
@@ -0,0 +1,187 @@
...
+import subprocess
+import conftest
+import shutil
+from constants import CONFIG_FILE
+from pathlib import Path
+
+pytestmark = pytest.mark.usefixtures("register_subman")
+
+PACKAGE = "insights-client-ros"
+SERVICE = "pmlogger"
```
pkoprda, line 22: In your tests you sometimes use this defined constant `SERVICE`, however in `test_cleanup_pmlogger_and_ros` you use `pmlogger`. I would suggest to use everywhere either this defined constant or just use string `"pmlogger"`

pkoprda (changes requested):

integration-tests/test_ros.py
zpetrace: Yeah, fixed the inconsistencies.

PR #438, opened 2025-06-13
[el9] fix(test): Fixed test_check_show_results

zpetrace:
In order for insights-client to reliably check the correct results from Advisor we needed to change the order of action in the test. The permission of the file is changed before registering to insights so that there is enough time for the issue to be reflected in Advisor.

(cherry picked from commit 92538eb0a705f29eb6800aa2aab17cde300b85df)

---

This pull request is a backport of: https://github.com/RedHatInsights/insights-client/pull/432


pkoprda (approved):
LGTM, thanks

PR #441, opened 2025-06-17
[el9] fix(test): Fixing some architectural differences in tests

zpetrace:
The test_tags and test_common_specs are currently not running on other architectures other than x86_64. This will need to be fixed some more in the future with better assert. For that we'll need systems not emulated but the exact architecture. Until then, as it is being worked on, we'll skip the tests for now.

(cherry picked from commit eafc474ace772afd264a59958aa5fc4b9b3717a7)

---

This pull request is a backport of: https://github.com/RedHatInsights/insights-client/pull/440

### DIFF
--- a/integration-tests/test_common_specs.py
+++ b/integration-tests/test_common_specs.py
@@ -18,6 +18,10 @@ import logging
 pytestmark = pytest.mark.usefixtures("register_subman")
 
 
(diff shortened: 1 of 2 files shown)


pkoprda (approved):
LGTM, let's just wait for the pipeline to finish, then we can merge this :)

PR #444, opened 2025-06-19
[el8] test: Adding tests for insights-client-ros

zpetrace:
I have added multiple tests for insights-client-ros. I am testing all the basic functionality of ROS - these tests test that we can download the subpackage and activate the pmlogger. Then we test if the latest metrics exists once the subpaclkage is started. Then they test basic functionality of insights while ROS is installed and active and lastly that we can uninstal the subpackage and the pmlogger is inactive again. This is an effor based on the CCT-1402.

(cherry picked from commit cfcd2287931da077b01a50b395d3be649e0d3302)

---

This pull request is a backport of: https://github.com/RedHatInsights/insights-client/pull/437


pkoprda (changes requested):
The tests are failing probably because you used the parameter `capture_output` in  `subprocess.run()` which is available since [Python3.7+](https://docs.python.org/3.13/library/subprocess.html#subprocess.run). You should use `stdout=subprocess.PIPE, stderr=subprocess.PIPE` instead

PR #449, opened 2025-07-03
fix(test): Fix Satellite failing tests

zpetrace:
Some tests were continuously failing when run against Satellite. The test_set_ansible_host_info was according to related CCT card not fixed on Satellite 6.15 and therefore should be skipped in that version. The test_compliance_policies_option was failing due to API communication troubles so in that case it will be also skipped. Based on some communications the test_display_name is not applicable on Satellite and therefore skipping it entirely if it run against it. This is the final part of CCT-1387.

---

- `el9` (all of RHEL 9)
- `el8` (all of RHEL 8)

* Card ID: CCT-1387

### DIFF
--- a/integration-tests/test_client_options.py
+++ b/integration-tests/test_client_options.py
@@ -22,7 +22,7 @@ ARCHIVE_CACHE_DIRECTORY = "/var/cache/insights-client"
 
 
 @pytest.mark.tier1
-def test_set_ansible_host_info(insights_client):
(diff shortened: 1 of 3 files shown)


zpetrace:
I am opening the PR for a review as the Satellite job is green and PASSING https://jenkins-csb-rhsm-prod.dno.corp.redhat.com/job/insights-client-satellite/63/

PR #452, opened 2025-07-11
[el8] fix(test): Fix Satellite failing tests

zpetrace:
Some tests were continuously failing when run against Satellite. The test_set_ansible_host_info was according to related CCT card not fixed on Satellite 6.15 and therefore should be skipped in that version. The test_compliance_policies_option was failing due to API communication troubles so in that case it will be also skipped. Based on some communications the test_display_name is not applicable on Satellite and therefore skipping it entirely if it run against it. This is the final part of CCT-1387.

(cherry picked from commit 0046638c6262a6ef6275ea9fce28718a243d8432)

---


pkoprda (approved):
LGTM, thanks

PR #453, opened 2025-07-15
moving gating rpms install from jenkins to test suite

Lorquas:
As part of testing-farm migration we need to do more test setup in this project and not Jenkins.  This exact setup is already being used by rhc.   Will need to be back-ported to all testing branches.


zpetrace (approved):
Thanks, LGTM, I have run a gating job with the changes, if it comes through I will merge this. Backports will probably be needed to both el9 and el8 @Lorquas.

PR #456, opened 2025-07-24
fix(test): Fix insights-client details file test

zpetrace:
Based on the change in CCT-1082 the insights-client details file should not get recreated after the --check-results option. I have therefore edited the test to reflect that change so the test does not fail.

---

- `el9` (all of RHEL 9)
- `el8` (all of RHEL 8)

* Card ID: CCT-1082


pkoprda (approved):
LGTM

PR #461, opened 2025-07-30
feat(test): Add selinux_context=None for user-invoked functionality tests

zpetrace:
Accommodate changes to InsightsClient.run() which now uses runcon by default to execute insights-client with insights_client_t SELinux context, matching production systemd behavior.

Updated user-invoked functionality tests to use selinux_context=None:
- test_version.py: --version command (1 call)
- test_client_options.py: --support, --validate, and unregistered --diagnosis (3 calls)
- test_connection.py: all --test-connection calls (5 calls)
- test_status.py: all --status calls (3 calls)
- test_motd.py: --status calls for MOTD testing (2 calls)
- test_collection.py: --verbose for timeout configuration testing (1 call)

Production/service functionality tests (registration, upload, compliance, data collection) retain default SELinux context to match systemd behavior.

This ensures tests properly differentiate between user-invoked commands and production service operations while maintaining comprehensive coverage of both scenarios.

Message Created by Cursor (calude-4-sonnet)
---

- `el9` (all of RHEL 9)


zpetrace:
The tests are expected to fail until https://github.com/RedHatInsights/pytest-client-tools/pull/28 is merged as well. There are necessary changes to allow for this parameter to be added to the tests.

m-horky (approved):
ACK

PR #464, opened 2025-07-30
[el8] fix(test): Fix Betelgeuse field mappings and invalid values

zpetrace:
- Update custom_betelgeuse_config.py to use casecomponent/subsystemteam instead of component/poolteam
- Fix test docstrings to use correct Polarion field names
- Resolve duplicate test case UUID in test_upload.py
- Fix invalid enumeration value insights-client-ros in test_ros.py"

I have also removed testimony.yml configuration and GitHub workflow as they are incompatible with our Betelgeuse-based test documentation format used for Polarion integration.

Testimony expects docstrings with 'assert', 'feature', and 'test' tokens, while Betelgeuse uses Polarion-specific fields like 'id', 'description', 'casecomponent', etc. Since our Polarion uploads are now working correctly with Betelgeuse, we standardize on the Betelgeuse format and remove the conflicting testimony validation.

(cherry picked from commit 7a2f7fbdf0f4160e54662c3922903023e739c838)

---


pkoprda (changes requested):
Looks good, could you just update the commit message a little? This branch does not have `testimony.yml` file in `.github/workflow` directory

PR #465, opened 2025-07-31
chore: Disable egg rotation

m-horky:
With Core available in site-packages, provided by RPM package insights-core, we can disable the egg rotation and verification.

With insights-core delivered as an RPM, we can drop the egg management machinery.

This may have broken some minor stuff (integration testing run by tmt, meson builds of the test directory), that will have to be verified and/or fixed later

While the RPM-shipped Core behaves mostly the same as the CDN Core, there are behavioral differences between the two.

Phase 2 drops the 'update' phase, where an egg is fetched and GPG-verified. In the end of 'collect_and_upload' phase, v2 drops egg rotation code, because there are no eggs to rotate.

---

- `rhel-9-main` (RHEL 9.8+)

* Card ID: CCT-1321

---

This PR probably breaks QE scripts in `systemtest/`, which would need to be updated to fetch and/or build RPM Core.


README.md
```diff
@@ -42,19 +42,7 @@ Follow these instructions to prepare your system for development.
...
-
-6. To build an insights-core egg from source, run `build_client_egg.sh` from the insights-core repo.
-
-   ```shell
-   $ cd ../insights-core
-   $ bash build_client_egg.sh
-   $ # File `insights.zip` gets created in the current directory
-   $ cd ../insights-client
-   $ # To use the zip file as an egg, pass `EGG=../insights-core/insights.zip`
+   $ sudo PYTHONPATH=./src:../insights-core python src/insights_client/__init__.py --help
```
pkoprda, line 45: Previous steps used `python3`, so maybe it would be better to be consistent with that

README.md
```diff
@@ -66,7 +54,7 @@ Follow these instructions to prepare your system for development.
 - Read [TESTING.md](TESTING.md) for more information.
 
 
-## Architecture Summary
+## Legacy Architecture Summary
```
pkoprda, line 57: Wouldn't it be better to put this somewhere below, since it is legacy? Or add before this some architecture summary which is not legacy?

Also `Phases` section will become legacy too, since there will be no `Update` phase

pkoprda (changes requested):
Mostly looks fine, I've tested it on Fedora42 and RHEL10 (Python3.12+) with various options, didn't look to break something. However I left some comments.

pkoprda (approved):

m-horky:
> so the RHEL pipeline ran and there are too many tests failing with this

That will happen until Core gets into there. Do we want to block this PR until then?

PR #466, opened 2025-08-01
fix(test): Fix unreliable systemd timer test

zpetrace:
The test was producing false positives/negatives due to complex
journal parsing and external API dependencies.

Changes:
- Replace journal monitoring with log file existence check
- Remove external API dependency causing timing issues
- Use 3-minute timer interval with 5-second delay
- Add reliable wait logic and proper cleanup
- Create constant for log file path and use helper functions consistently

This is more reliable as log file creation is direct proof of
execution without external timing dependencies.

---

- `el9` (all of RHEL 9)
- `el8` (all of RHEL 8)

* Card ID: CCT-1557


pkoprda (changes requested):
The code seems a bit unnecessary complex, I left some comments with the improvement.

Also could you include the reference to the Jira ticket in the commit message and modify a bit commit message so it would not be that long? Thanks!

PR #473, opened 2025-08-13
fix(test): Remove fields now provided by Jenkins pipeline

zpetrace:
These fields are now injected via the Jenkins pipeline, so we don't need to define them in the integration tests and related config. This removes duplication and avoids drift between CI and the test suite.

---

- `el9` (all of RHEL 9)
- `el8` (all of RHEL 8)


Aftermath (approved):
LGTM.

PR #480, opened 2025-08-15
[el8] fix(test): Fix unreliable systemd timer test

zpetrace:
Simplify systemd timer test to use log file detection

The test_data_upload_systemd_timer test was producing false positives and false negatives due to complex journal parsing and external API dependencies.

Changes made:
- Replace complex journal log monitoring with simple log file existence check
- Remove external_inventory API dependency that caused timing issues
- Use direct log file creation at /var/log/insights-client/insights-client.log as proof of execution
- Simplify timer configuration to 3-minute interval with 5-second delay
- Implement reliable wait logic: timer interval + buffer time
- Add proper cleanup of both timer configuration and log files
- Remove unused imports and fix linter errors

This approach is much more reliable because:
- Log file creation is direct proof that insights-client executed and uploaded data
- No dependency on external API timing or propagation delays
- Simpler timing model that just waits for the timer interval

(cherry picked from commit dbed32e19ecdafc9cbb2051c95d182aeb4a464d1)

---

This pull request is a backport of: https://github.com/RedHatInsights/insights-client/pull/466

* Card ID: CCT-1557

### DIFF
--- a/integration-tests/test_client_systemd.py
+++ b/integration-tests/test_client_systemd.py
@@ -7,75 +7,176 @@
 """
 
 import conftest
-from time import sleep
 import pytest
 import subprocess
 from pathlib import Path
+import logging
 
+logger = logging.getLogger(__name__)
 pytestmark = pytest.mark.usefixtures("register_subman")
 
 
+def _run_systemctl_command(cmd, description="systemctl operation"):
+    """Helper function to run systemctl commands with proper error handling."""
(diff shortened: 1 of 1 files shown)


pkoprda (changes requested):
Please drop `text` parameter in `subprocess.run()`, that parameter is available since [Python3.7+](https://docs.python.org/3.13/library/subprocess.html#subprocess.run). Use `universal_newlines` parameter instead

PR #488, opened 2025-08-21
feat(readme): update README with new instructions

Peto-RH:
* Card ID: CCT-1062

Added a section about log management. Include what directories does the client use. Added a section about which services the client connects to.

---


m-horky (changes requested):
Hi, I left a few comments there. Overall, it is not bad, but addressing them might make it clearer for the target audience.

PR #489, opened 2025-08-26
fix(test): Make hostname test work in container environments

pkoprda:
* Card ID: CCT-994

Remove container skip condition and add container-specific logic that:
- Modifies hostname files in archive instead of system hostname
- Uploads modified archive using --payload and --content-type
- Verifies archive modification and upload success

Co-Authored-By: Claude (claude-sonnet-4)

---

- `el9` (all of RHEL 9)
- `el8` (all of RHEL 8)


pkoprda:
> Unless we test it elsewhere already, I'd drop the code changing hostnames and only go with the modified archive.

I kept the original test as was before (I just dropped unnecessary sleep from the test, see the reason in [PR description](https://github.com/RedHatInsights/insights-client/pull/489#issue-3356311300)) and I added a new test `test_file_workflow_archive_modification_and_upload` which is testing modification of the archive and then uploading to Ingress. However, there are already some tests for that in [test_upload.py](https://github.com/RedHatInsights/insights-client/blob/master/integration-tests/test_upload.py), so I am not sure if this new test is necessary.

m-horky (approved):
ACKing since the tests are passing, to get it off the table. If the need to change this appears again, we'll update it.

PR #491, opened 2025-08-28
[el8] fix(test): Fix unreliable systemd timer test

zpetrace:
The test was producing false positives/negatives due to complex journal parsing and external API dependencies.

Changes:
- Replace journal monitoring with log file existence check
- Remove external API dependency causing timing issues
- Use 3-minute timer interval with 5-second delay
- Add reliable wait logic and proper cleanup
- Create constant for log file path and use helper functions consistently

This is more reliable as log file creation is direct proof of execution without external timing dependencies. This should solve CCT-1557.

(cherry picked from commit 3f5e89e00d229f552521af8e9b6ef0b0e747bb4e)

---
This pull request is a backport of: #466

* Card ID: CCT-1557

### DIFF
--- a/integration-tests/constants.py
+++ b/integration-tests/constants.py
@@ -5,3 +5,4 @@ REGISTERED_FILE: str = "/etc/insights-client/.registered"
 UNREGISTERED_FILE: str = "/etc/insights-client/.unregistered"
 MACHINE_ID_FILE: str = "/etc/insights-client/machine-id"
 TAGS_FILE = pathlib.Path("/etc/insights-client/tags.yaml")
+INSIGHTS_CLIENT_LOG_FILE = "/var/log/insights-client/insights-client.log"
--- a/integration-tests/test_client_systemd.py
+++ b/integration-tests/test_client_systemd.py
@@ -7,75 +7,158 @@
 """
 
 import conftest
-from time import sleep
 import pytest
 import subprocess
 from pathlib import Path
+import logging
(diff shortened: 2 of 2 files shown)


zpetrace:
As opposed to the master and el9 changes I have Replaced `capture_output=True` with `stdout=subprocess.PIPE, stderr=subprocess.PIPE` to work with corresponding Python version of this branch. I have also put more `try..except` blocks as `Path.unlink(missing_ok=True)` couldn't be used here as well.

integration-tests/constants.py
```diff
@@ -5,3 +5,5 @@
 UNREGISTERED_FILE: str = "/etc/insights-client/.unregistered"
 MACHINE_ID_FILE: str = "/etc/insights-client/machine-id"
 TAGS_FILE = pathlib.Path("/etc/insights-client/tags.yaml")
+CONFIG_FILE = "/etc/insights-client/insights-client.conf"
```
pkoprda, line 8: Is this used anywhere? I think this shouldn't be part of this commit

pkoprda (changes requested):

integration-tests/constants.py
zpetrace: whoops. yeah, that got added as a part of the cherry-pick and I did not catch that, will remove it.

pkoprda (approved):
LGTM, thanks!

PR #492, opened 2025-08-28
fix(test): Expand architecture support for the earlier skipped tests

zpetrace:
For test_common_specs I have removed x86_64-only restrictions and added targeted IBM arch handling. I enabled ARM64 support while skipping problematic specs (lspci/dmidecode) on IBM Power/Z systems.
Finally I fixed inventory synchronization timing issues with better inventory polling rather that letting the system sleep for some time in test_tags.

---

- `el9` (all of RHEL 9)
- `el8` (all of RHEL 8)

* Card ID: CCT-1440


pkoprda (approved):
LGTM

PR #495, opened 2025-08-29
feat: Add obfuscation_list option to the config file

pkoprda:
* Card ID: CCT-1576

To support the new configuration for IPv6 and MAC address obfuscation, this commit adds obfuscation_list option and removes the old obfuscate options in the configuration file. Also this commit updates manpage for the configuration file to reflect this change.

---

- `el9` (all of RHEL 9)
- `el8` (all of RHEL 8)

### DIFF
--- a/data/insights-client.conf
+++ b/data/insights-client.conf
@@ -16,11 +16,9 @@
 # Automatically update the dynamic configuration
 #auto_update=True
 
-# Obfuscate IP addresses
-#obfuscate=False
-
-# Obfuscate hostname. Requires obfuscate=True.
-#obfuscate_hostname=False
+# Specify which data types should be obfuscated in the data collection
+# Supported options: ipv4, ipv6, hostname, mac (comma-separated list)
+#obfuscation_list=
 
 # Display name for registration
 #display_name=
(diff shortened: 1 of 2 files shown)


m-horky (changes requested):
Patch itself looks fine. Tests look good as well, that's a good sign.

I'd like to see integration tests created for this new key/value combo. One parametrized test could be enough:

```py
@pytest.mark.parametrize(
    "input,output",
    [
    "ipv4", ("<obfuscated-v4>", "<v6>", "<hostname>", "<mac>"),
    "ipv4,ipv6", (...),
    "ipv4, ipv6", (...),
    "ipv5", (...),
    ...
    ]
)
def test_obfuscation_list(insights_client):
    ...
```

No need to test all combinations, just that each of the possible obfuscation types is applied when requested, and that bad values are handled reasonably.

m-horky (changes requested):
The test is too complicated and will take too long. It seems it would work, but we can significantly simplify it.

m-horky:
Let's wait for RHEL CI. If it passes, we can go with merging it. Fedora jobs failing on Playbook verifier is a know issue that can be ignored for this.

PR #504, opened 2025-09-08
[el8] fix: Fix file workflow test for container environments

pkoprda:
* Card ID: CCT-994

The test for updating info in the archive was failing in containers, which are not in charge of their own hostnames. This commit resolves the test failures by simulating hostname updates through archive file modification.

---
This pull request is a backport of: #489


m-horky (approved):
ACKing since the tests are passing, to get it off the table. If the need to change this appears again, we'll update it.

PR #520, opened 2025-09-24
chore: Remove delay from status registered test

pkoprda:
Remove unnecessary 5-second delay. The test already uses loop_until to wait for registration, making the additional delay redundant.

---

- `rhel-9-main` (RHEL >= 9.8)
- `rhel-9-egg` (RHEL <= 9.7)
- `rhel-8-egg` (RHEL 8)


m-horky (approved):
ACK

PR #529, opened 2025-09-25
[rhel-9-main] chore: Disable egg rotation

m-horky:
* Card ID: CCT-1321

With Core available in site-packages, provided by RPM package insights-core, we can disable

- the egg rotation and verification,
- egg management machinery.

This patch also

- Deletes all egg files that may be present on the system.
- Simplifies MOTD handling which used to rely on a presence of an egg file.

Co-Authored-By: Claude Code (claude-sonnet-4)

(cherry picked from commit 4d1d76ec99c87eb82080874cff73332e734b4ece)

This pull request is a backport of: #465


m-horky:
Rebased to match post-review changes in the reference PR.

PR #532, opened 2025-10-17
feat(ci): using tmt to install test deps for image mode, removing unu…

Lorquas:
This moves some of the test setup earlier in the test process, which now works for image mode testing.

- Moves rpm test dependencies to the tmt prepare step
- Removes downstream PR code since all that happens here on github now

What branches do I need to backport this to?

can't be merged until: https://github.com/RedHatInsights/pytest-client-tools/pull/38

This can be tested with `testing-farm request --compose RHEL-10.2-image-mode`


m-horky (approved):
ACK, I'll let Zdenek do the review as well.

m-horky:
> What branches do I need to backport this to?

Ideally all effected, which means `rhel-10-egg`, `rhel-9-main`, `rhel-9-egg`.

The `copr-setup.sh` will eventually need to go in old branches (`rhel-8-egg`) as well, but we can handle it on our own I think.

m-horky:
After internal communication, merging as-is. Let's revisit if we need further changes. The CI failures are not related to this patch.

PR #533, opened 2025-10-22
refactor: moved loop_until to pytest-client-tools

Lorquas:
moved loop_until to pytest-client-tools since it seems to be useful for all cct pytests.

Can't be mreged until https://github.com/RedHatInsights/pytest-client-tools/pull/39


jstavel (approved):
I have reviewed the code:
- there is no old way of using loop_until
- no conftest.loop_until

PR #539, opened 2025-11-10
feat(test): Add test for insights-client --register in unconfined_t SELinux context

zpetrace:
Implement test_register_unconfined_t_no_context_change that verifies insights-client --register executes correctly when running as root in unconfined_t SELinux context with SELinux enforcing mode enabled.

The test:

    - Skips if SELinux is disabled
    - Uses _selinux_mode() context manager to automatically set SELinux to enforcing mode and restore original mode on exit (no manual teardown needed)
    - Verifies execution context is unconfined_t by reading /proc/self/attr/current directly (kernel interface, avoids subprocess overhead and syscall interference)
    - Runs insights-client --register and verifies successful registration
    - Uses SELinuxDenialsChecker context manager to automatically track time period for audit log queries
    - Checks audit logs for process execution contexts (SYSCALL execve events) using ausearch to verify that insights-client and insights-core processes remained in unconfined_t context, even when no AVC denials occur
    - Checks for SELinux AVC denials using ausearch
    - Verifies that no processes transitioned to confined contexts (insights_client_t or insights_core_t)

Key improvements:

    - Context managers ensure automatic cleanup of SELinux mode and time tracking
    - Process context verification from audit logs works even when no AVC denials occur (the expected case), addressing the limitation of only checking denials
    - Direct kernel interface (/proc/self/attr/current) instead of subprocess for more reliable context reading
    - Simplified code structure with helper functions, context managers, and classes properly organized
    - Relies solely on ausearch for audit log queries (no fallback to direct file reading)

Commit message co-created with Cursor.

* Card ID: CCT-1717

---

- rhel-9-main (RHEL >= 9.8)

### DIFF
--- a/integration-tests/test_selinux.py
+++ b/integration-tests/test_selinux.py
@@ -1,0 +1,287 @@
+"""
+:casecomponent: insights-client
+:requirement: RHSS-291816
+:subsystemteam: rhel-sst-csi-client-tools
+:caseautomation: Automated
+:upstream: Yes
+"""
+
+import re
+import subprocess
+import time
+import pytest
+from contextlib import contextmanager
+from pytest_client_tools.util import loop_until
+from datetime import datetime
+
+pytestmark = pytest.mark.usefixtures("register_subman")
+
+
+# Skip entire test file if SELinux is disabled
+def _get_selinux_mode():
+    """Get current SELinux mode (Enforcing, Permissive, or Disabled)."""
+    try:
+        result = subprocess.run(
+            ["getenforce"], capture_output=True, text=True, check=True
+        )
+        return result.stdout.strip()
+    except (subprocess.CalledProcessError, FileNotFoundError):
+        return "Disabled"
+
+
+# Check if SELinux is disabled and skip all tests in this module
+if _get_selinux_mode() == "Disabled":
+    pytest.skip("SELinux is disabled on this system", allow_module_level=True)
+
+
+def _get_current_context():
+    """Get current SELinux context by reading /proc/self/attr/current.
+    This uses the kernel interface directly.
+    """
+    with open("/proc/self/attr/current", "r") as f:
+        context = f.read().strip()
+        # Remove null bytes that may be present in the file
+        return context.replace("\x00", "")
+
+
+def _format_audit_time(timestamp):
+    """Convert timestamp to audit log format."""
+    if isinstance(timestamp, (int, float)):
+        return datetime.fromtimestamp(timestamp).strftime("%Y/%m/%d %H:%M:%S")
+    return timestamp or "today"
+
+
+def _check_denials_with_ausearch(start_time, end_time=None):
+    # Check for denials using ausearch command.
+    cmd = [
+        "ausearch",
+        "--message",
+        "AVC",
+        "--comm",
+        "insights-client",
+        "--start",
+        _format_audit_time(start_time),
+    ]
+    if end_time:
+        cmd.extend(["--end", _format_audit_time(end_time)])
+    result = subprocess.run(cmd, capture_output=True, text=True, timeout=5, check=False)
+    return result.stdout.strip() or None
+
+
+def _parse_execve_event_block(event_lines):
+    """
+    Parse a block of audit lines (one event) to extract process info.
+    Correlates SYSCALL (context) with EXECVE (full arguments).
+    """
+    # Join lines to make searching easier, or iterate
+    full_text = "\n".join(event_lines)
+
+    # 1. Extract contexts from SYSCALL
+    # scontext = subject context (parent process that executed)
+    scontext_match = re.search(r"(?:subj|scontext)=([^\s]+)", full_text)
+    if not scontext_match:
+        return None
+    scontext = scontext_match.group(1)
+
+    # tcontext = target context (file being executed)
+    # newcontext = actual running context if setexeccon was used
+    newcontext_match = re.search(r"newcontext=([^\s]+)", full_text)
+    tcontext_match = re.search(r"tcontext=([^\s]+)", full_text)
+
+    # Running context is newcontext if setexeccon was used, otherwise tcontext
+    if newcontext_match:
+        running_context = newcontext_match.group(1)
+    elif tcontext_match:
+        running_context = tcontext_match.group(1)
+    else:
+        # Fallback to scontext if neither found (shouldn't happen normally)
+        running_context = scontext
+
+    # 2. Extract Command/Arguments from EXECVE
+    if "type=EXECVE" not in full_text:
+        return None
+
+    # Extract all arguments to reconstruct the command roughly
+    args = re.findall(r'a\d+=(?:"([^"]+)"|([^\s]+))', full_text)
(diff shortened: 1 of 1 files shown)


integration-tests/test_selinux_unconfined.py
```diff
@@ -0,0 +1,192 @@
...
+        end_idx = line.find(":", start_idx)
+        if end_idx > start_idx:
+            return float(line[start_idx:end_idx])
+    except (ValueError, IndexError):
+        pass
+    return None
+
+
+def _check_denials_in_audit_log(process_name, start_time):
+    """Check for denials by reading audit.log directly."""
```
pkoprda: ditto

integration-tests/test_selinux_unconfined.py
```diff
@@ -0,0 +1,192 @@
...
+    except FileNotFoundError:
+        pytest.skip(reason="id command not available - cannot check SELinux context")
+    except subprocess.CalledProcessError as e:
+        pytest.skip(reason=f"Failed to get SELinux context: {e}")
+
+
+def _check_denials_with_ausearch(process_name, start_time):
+    """Check for denials using ausearch command."""
+    from datetime import datetime
+
```
pkoprda: Put this import at the beginning of this file

pkoprda (changes requested):
Thanks, I provided some comments with the improvement.

Would it be possible to export the helper functions to some other file? I think you can use these functions when you will write other tests related to new selinux policy.

Also I believe that this change (according to Jira card) should be backported only to `rhel-9-main` branch, in other branches insights-core is not selinux confined.

integration-tests/test_selinux.py
```diff
@@ -0,0 +1,183 @@
...
+        2. Register system using subscription-manager (already done by fixture)
+        3. Switch system to SELinux enforcing mode
+        4. Verify current context is unconfined_t
+        5. Run insights-client --register as root
+        6. Check that insights-core code was executed in unconfined_t context
+        7. Check that there was no SELinux denial during execution
+        8. Restore original SELinux mode
+    :expectedresults:
+        1. Current SELinux mode is saved
+        2. System is registered (handled by register_subman fixture)
```
pkoprda: Ditto

integration-tests/test_selinux.py
```diff
@@ -0,0 +1,183 @@
...
+        3. SELinux is set to enforcing mode
+        4. Current context is confirmed as unconfined_t
+        5. insights-client --register completes successfully
+        6. insights-core runs in unconfined_t context (no transition to confined)
+        7. No SELinux denials are found in audit logs
+        8. Original SELinux mode is restored
+    """
```
pkoprda: I don't think this should be as expected result. There is no assert to check that the original selinux mode is restored. So I would remove this expected result.

pkoprda (changes requested):

integration-tests/test_selinux.py
```diff
@@ -0,0 +1,212 @@
...
+                for i in ["This host is registered", "Registered"]
+            ):
+                return True
+            return False
+
+        assert loop_until(check_registered_with_context)
+        assert result.returncode == 0, f"Registration failed: {result.returncode}"
+
+        # Check no SELinux denials occurred
+        denials = _check_selinux_denials(start_time=start_time)
```
pholica: In case of no AVCs (valid and expected state), the test is unable to check that the transition to `insights_core_t` happened and it remained in the `unconfined_t` context, so different approach needs to be chosen to test this.

integration-tests/test_selinux.py
```diff
@@ -0,0 +1,212 @@
...
+                        f"Process transitioned to confined context {scontext} "
+                        f"instead of staying in unconfined_t. This indicates that "
+                        f"insights-client transitioned to insights_client_t when it "
+                        f"should have remained in unconfined_t. Denials:\n{denials}"
+                    )
+        assert not denials, f"SELinux denials found:\n{denials}"
+
+    finally:
+        # Teardown: Restore original SELinux mode
+        if original_mode != "Disabled":
```
pholica: This is probably rather a suggestion for future improvement that does not need to be addressed now, but it would make sense to change the `_set_selinux_mode` to a context manager which would during the `__exit__` restore the original state.

integration-tests/test_selinux.py
```diff
@@ -0,0 +1,287 @@
...
+        assert "unconfined_t" in context, f"Expected unconfined_t, got: {context}"
+
+        with SELinuxDenialsChecker() as checker:
+            result = insights_client.run("--register", selinux_context=context)
+            assert result.returncode == 0, f"Registration failed: {result.returncode}"
+
+            # Verify registration
+            def check_registered():
+                status = insights_client.run(
+                    "--status", check=False, selinux_context=context
```
pholica: The context does not have to be passed to achieve the same context for the new process. In that case, the `selinux_context` argument can be set just to `None`

integration-tests/test_selinux.py
```diff
@@ -0,0 +1,287 @@
...
+    target_markers = ["insights-client", "insights_client", "insights-core", "run.py"]
+    if not any(marker in cmd_string for marker in target_markers):
+        return None
+    # Comm is usually the base command (first arg or explicit comm field)
+    comm_match = re.search(r'comm="([^"]+)"', full_text)
+    comm = (
+        comm_match.group(1) if comm_match else (cmd_args[0] if cmd_args else "unknown")
+    )
+    # Return format: (comm_name, source_context/unused, running_context)
+    return (comm, running_context, running_context)
```
pholica: This looks fishy, the second and third items in the tuple are the same.

integration-tests/test_selinux.py
```diff
@@ -0,0 +1,287 @@
...
+    """
+    if _get_selinux_mode() == "Disabled":
+        pytest.skip(reason="SELinux is disabled on this system")
+
+    with _selinux_mode("Enforcing"):
+        context = _get_current_context()
+        assert "unconfined_t" in context, f"Expected unconfined_t, got: {context}"
+
+        with SELinuxDenialsChecker() as checker:
+            result = insights_client.run("--register", selinux_context=context)
```
pholica: The context does not have to be passed to achieve the same context for the new process. In that case, the `selinux_context` argument can be set just to `None`

integration-tests/test_selinux.py
```diff
@@ -0,0 +1,287 @@
...
+                )
+                return status.returncode == 0 and any(
+                    i in status.stdout
+                    for i in ["This host is registered", "Registered"]
+                )
+
+            assert loop_until(check_registered)
+
+        # Verify process contexts from audit logs
+        for proc_name, scontext, running_context in checker.get_process_contexts():
```
pholica: The logic below looks quite complicated and it's hard to assess whether the right thing is checked. In general, I don't have problem with the `any`, but my head started to spin when I started to think about this code :sweat_smile: Right now I however don't have suggestion for better way other than yet another abstraction layer on top of the audit log. Having that said, let's stick with this implementation if it works and I may introduce changes later if I have some better idea that proves to be working. What do you think?

PR #507, opened 2025-09-12
chore: Remove meson build system

Peto-RH:
### **Removed meson build system entirely** 
This includes:
  - Deleting meson.build files
  - Deleting meson_options.txt
  - Modifying .spec file to install systemd files without meson
  - Deleting many .in files because they are no longer needed
    - hardcoding meson macros that were in .in files
  - Removing some unused constants from constants.py.in
  - Updating build dependencies in .spec

---
* Card ID: CCT-1567
---
#### results of `diff`:
note: I downloaded the other RPM with meson from #533 
Since the ticket CCT-1760 removes these 2 options I:
- did not do `diff` between old RPM with option checkin enabled and new RPM with option checkin.
- did not do `diff` between old RPM with option auto_registration enabled and new RPM with option auto_registration.
- comparing RPM without ros. Other thatn file contents I also compared "header" info, like the Name, Version, Release, Summary, License, %post and %preun scripts side-by-side, Requires: lines
``` 
~/insights-client$ diff -r RPM-no_meson-insights-client-3.10.2-0.el10.noarch.rpm RPM-with_meson-insights-client-3.10.2-0.el10.noarch.rpm/
diff -r RPM-no_meson-insights-client-3.10.2-0.el10.noarch.rpm/usr/bin/insights-client RPM-with_meson-insights-client-3.10.2-0.el10.noarch.rpm/usr/bin/insights-client
7c7
< sys.path.insert(1, "/usr/lib/python3.12/site-packages")
---
> sys.path.insert(1, "/usr/lib/python3.12/site-packages/")
diff -r RPM-no_meson-insights-client-3.10.2-0.el10.noarch.rpm/usr/lib/python3.12/site-packages/insights_client/constants.py RPM-with_meson-insights-client-3.10.2-0.el10.noarch.rpm/usr/lib/python3.12/site-packages/insights_client/constants.py
6a7,15
> PREFIX = "/usr"
> BINDIR = "/usr/bin"
> SBINDIR = "/usr/sbin"
> LIBEXECDIR = "/usr/libexec"
> DATAROOTDIR = "/usr/share"
> DATADIR = "/usr/share"
> SYSCONFDIR = "/etc"
> LOCALSTATEDIR = "/var"
> DOCDIR = "/usr/share/doc/insights-client"
[...]

PR #540, opened 2025-11-10
feat: drop distro dependency

Peto-RH:
* Card ID: CCT-1619

The distro package is a dependency for the insights-client integration test suite. To reduce the need for third-party dependencies, we decided to drop this package.

---

- `rhel-9-main` (RHEL >= 9.8)
- `rhel-9-egg` (RHEL <= 9.7)
- `rhel-8-egg` (RHEL 8)

### Notes for the reviewer:
I reviewed the test farm, and the only test potentially affected by this change (test_compliance.py) fails with the same error as reported in issue #507 . This suggests the failure is unrelated to my changes.

While I aimed to keep the implementation stylistically close to the original distro package, I chose not to use the shlex module for parsing /etc/os-release

### DIFF
--- a/integration-tests/requirements.txt
+++ b/integration-tests/requirements.txt
@@ -1,4 +1,3 @@
 git+https://github.com/RedHatInsights/pytest-client-tools@main
 pyyaml
 pytest-subtests
-distro
--- a/integration-tests/test_compliance.py
+++ b/integration-tests/test_compliance.py
@@ -6,14 +6,29 @@
 :upstream: Yes
 """
 
+import os
 import pytest
-import distro
 from pytest_client_tools.util import loop_until
+
+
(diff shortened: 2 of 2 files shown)


integration-tests/utils.py
pkoprda: You don't have to create a new file, `get_distro_id()` can be in `integration-tests/conftest.py` file.

integration-tests/utils.py
```diff
@@ -0,0 +1,47 @@
+import os
+
+_ALLOWED_OS_IDS = {
```
pkoprda: We just need rhel OS, no need to create a list of allowed OS IDs

pkoprda (changes requested):

PR #553, opened 2025-11-25
[CCT-1719] Test checking that client transitions to core SELinux context

pholica:
* Card ID: CCT-1719

---

- `rhel-9-main` (RHEL >= 9.8)

### DIFF
--- a/integration-tests/test_selinux.py
+++ b/integration-tests/test_selinux.py
@@ -14,6 +14,8 @@ from contextlib import contextmanager
 from pytest_client_tools.util import loop_until
 from datetime import datetime
 
+from constants import REGISTERED_FILE
+
 pytestmark = pytest.mark.usefixtures("register_subman")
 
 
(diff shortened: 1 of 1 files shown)


integration-tests/test_selinux.py
zpetrace: The test looks good, and from the results in the checks before and now it seems like it's working and failing/passing as expected. Thanks!

pholica:
I've moved the PR to draft as it became quite bloated and I should better split it into multiple PRs focusing on different areas as the test code itself (I believe) proved to be correct.

zpetrace (dismissed):
LGTM, thanks!

zpetrace (approved):
The failing test `integration-tests/test_collection.py::test_output_dir_creates_archive_for_directory` was seen in another PR and that suggest again that this is a issue we should be solving some time later but is not connected to this PR. Other than that the test looks good so I'll merge the PR.

PR #569, opened 2025-12-03
fix(test): Fix flaky tests that use --check-results

pholica:
There seems to be a race condition in tests that use --check-result which requires system to appear both in inventory and in advisor. Otherwise the tests don't behave as expected. This PR adds active waiting for both in tests that seem to be affected by this.

* Card ID: CCT-1634

---

- `rhel-9-main` (RHEL >= 9.8)
- `rhel-9-egg` (RHEL <= 9.7)
- `rhel-8-egg` (RHEL 8)

### DIFF
--- a/integration-tests/test_checkin.py
+++ b/integration-tests/test_checkin.py
@@ -41,8 +41,9 @@ def test_ultralight_checkin(insights_client, test_config):
         4. The updated timestamps were retrieved and recorded
         5. Both updated timestamps will be greater than before check-in
     """
-    insights_client.register()
-    assert loop_until(lambda: insights_client.is_registered)
+    assert insights_client.register(wait_for_registered=True)
+    assert insights_client.wait_for_inventory()  # required by --check-results
+    assert insights_client.wait_for_advisor()  # required by --check-results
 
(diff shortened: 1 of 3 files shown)


integration-tests/test_checkin.py
```diff
@@ -41,8 +41,9 @@ def test_ultralight_checkin(insights_client, test_config):
         4. The updated timestamps were retrieved and recorded
         5. Both updated timestamps will be greater than before check-in
     """
-    insights_client.register()
-    assert loop_until(lambda: insights_client.is_registered)
+    assert insights_client.register(wait_for_registered=True)
+    assert insights_client.wait_for_inventory()  # required by --check-result
+    assert insights_client.wait_for_advisor()  # required by --check-result
```
pkoprda: This should be `--check-results` (check manpage of insights-client)
```suggestion
    assert insights_client.wait_for_inventory()  # required by --check-results
    assert insights_client.wait_for_advisor()  # required by --check-results
```
Same for the changes bellow

pkoprda (commented):
Just a small nitpick in this PR (non-blocking though)

pkoprda (approved):
LGTM

PR #571, opened 2025-12-03
Address SELinux incompatibility of testing

pholica:
Some tests were not prepared for SELinux testing. Mostly, they are testing command-line/tool related functionality rather than the "service" mode of insights-client - being invoked by system timer or by other services such as rhccc. Those test cases need to explicitly call insights-client as a user would do not setting SELinux context.

* Card ID: CCT-1867

---

- `rhel-9-main` (RHEL >= 9.8)

### DIFF
--- a/integration-tests/test_client_options.py
+++ b/integration-tests/test_client_options.py
@@ -110,7 +110,12 @@ def test_group(insights_client, tmp_path):
     group_name = "testing-group"
 
     # Running insights-client in offline mode to generate archive
-    insights_client.run("--offline", f"--group={group_name}", f"--output-dir={tmp_path}")
+    insights_client.run(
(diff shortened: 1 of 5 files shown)


pholica:
Ok, so I tried `FAILED integration-tests/test_collection.py::test_output_dir_with_not_empty_directory` and it fails regardless the SELinux, so there's something else wrong there and the `test_collection.py::test_output_dir_with_not_empty_directory` is not failing on my test system no matter if SELinux is in permissive or enforcing mode (let me re-run the tests to see if this was some other flakiness/infra issue other than SELinux).
When it comes to the CentOS Stream tests, I see them more as nice to have and with lesser priority right now so I'm not going to spend time on it now.

PR #573, opened 2025-12-04
feat: building RPMs with ros subpackages

Peto-RH:
Modified .spec file to always build additional RPM with ros subpackage.

* Card ID: CCT-1872

---

- `rhel-9-main` (RHEL >= 9.8)
- `rhel-9-egg` (RHEL <= 9.7)
- `rhel-8-egg`


pkoprda (approved):
Approved, thanks!

PR #576, opened 2025-12-09
ci: Change max line length to 100

pholica:
This change is in agreement with current key contributors of the project.

---

- `rhel-9-main` (RHEL >= 9.8)
- `rhel-9-egg` (RHEL <= 9.7)
- `rhel-8-egg` (RHEL 8)


zpetrace (approved):
LGTM, thanks!

PR #581, opened 2025-12-09
fix(test): quote pytest filter

pholica:
This caused issues when the PYTEST_FILTER contains whitespace characters as the next "word" is passed as additional argument to pytest and not as a part of the filter expression.

---
- `rhel-9-main` (RHEL >= 9.8)
- `rhel-9-egg` (RHEL <= 9.7)
- `rhel-8-egg` (RHEL 8)


zpetrace (approved):
LGTM, thanks. I tried to start a run with the changes on Jenkins and it seems to be working fine, will merge soon!

PR #601, opened 2025-12-19
fix: Remove redhattools.pub.gpg file and all references

zpetrace:
The redhattools.pub.gpg file was used to store the GPG public key for verification of egg files. Since insights-core is now shipped as an RPM package, this file is no longer needed.

Changes:
- Deleted data/redhattools.pub.gpg file
- Removed file installation entry from insights-client.spec
- Removed file from data/meson.build build configuration
- Removed symlink instruction from README.md developer setup

* Card ID: CCT-1812

---

- `rhel-9-main` (RHEL >= 9.8)


pkoprda (changes requested):

pkoprda (approved):
LGTM

PR #602, opened 2025-12-22
fix: Failure to remove missing rpm.egg files

Peto-RH:
* Removed redundant rm commands from the %post script. These commands were deleting files that the RPM update handles automatically (because the old RPM owned those files), resulting in unnecessary file removal warnings during updates.

* Card ID: CCT-1802

- `rhel-9-main` (RHEL >= 9.8)


pkoprda (approved):
LGTM, verification comment is in the relevant CCT card.

PR #603, opened 2026-01-02
test: Check that egg related content is not present on the system

pkoprda:
* Card ID: CCT-1858

To make sure that egg related content is no longer used and is not brought in to the system by the insights-client, the universal check was needed for the tests. This commit implements check that the directory /var/lib/insights is empty and files related to egg-based insights-core are not present on the system.

---

- `rhel-9-main` (RHEL >= 9.8)


zpetrace (approved):
LGTM, thanks! The failed test in the CentOS check is not related to the change so I think we should be good to merge.

PR #604, opened 2026-01-05
[rhel-9-main] fix: Remove redhattools.pub.gpg file and all references

zpetrace:
The redhattools.pub.gpg file was used to store the GPG public key for verification of egg files. Since insights-core is now shipped as an RPM package, this file is no longer needed. This is a work for CCT-1812.

Changes:
- Deleted data/redhattools.pub.gpg file
- Removed file installation entry from insights-client.spec
- Removed file from data/meson.build build configuration
- Removed symlink instruction from README.md developer setup

(cherry picked from commit c08049bba77597a6bc6df4207d86b1712951473a)

---


pkoprda (approved):
LGTM, the failing tests are unrelated to this change.

PR #609, opened 2026-01-07
ci: Crash when failing to install requested packages

pholica:

- `rhel-9-main` (RHEL >= 9.8)
- `rhel-9-egg` (RHEL <= 9.7)
- `rhel-8-egg` (RHEL 8)


Peto-RH (changes requested):

pholica:
Thanks for the review!

PR #612, opened 2026-01-14
fix(test): Verify no egg-based content before every test

pkoprda:
The fixture for checking that there is no egg-based content on the system was incorrect (it ran only once at the very beginning of the entire test run). This change fixes it so the fixture runs before every test.

---

- `rhel-9-main` (RHEL >= 9.8)


m-horky (dismissed):
ACK

m-horky (dismissed):
OK

PR #630, opened 2026-01-28
[rhel-8-egg] fix(ci) Print versions of tested packages

pkoprda:
This information is not available anywhere especially when executed in TF. Print the versions immediately before the testing starts (at moment when the state of the installed packages should be stable).

(cherry picked from commit 33a3bc06f587211ad602f5c275a2d48d9e4f60e1)

---

This pull request is a backport of: #619


Peto-RH (approved):
LGTM

PR #633, opened 2026-02-02
chore: removed obsolete systemd services

Peto-RH:
* Removed auto_registration and checkin systemd services because they are no longer used by insights client.

* Card ID: CCT-1760

---

- `rhel-9-main` (RHEL >= 9.8)


pkoprda (approved):
LGTM

PR #654, opened 2026-03-17
chore: bump black from 24.3.0

m-horky:
Addresses https://github.com/RedHatInsights/insights-client/pull/652 

We are apparently not very good with managing Black versions. The simplest way to ensure we are kept up to date is to stop pinning the version.

insights-client codebase is small enough, we shouldn't be hitting any weird formatting issues as new minor versions are releases.


pkoprda (approved):
LGTM, thanks!

PR #657, opened 2026-03-27
feat(ci): add image-mode/bootc test support via TMT native builder

Lorquas:
## Summary

Add image-mode (bootc) test infrastructure for insights-client by leveraging
TMT 1.68+'s native bootc builder support. TMT automatically detects image-mode
systems, wraps shell prepare scripts into Containerfile RUN commands, builds the
image with podman, runs `bootc switch`, and manages the reboot -- no custom
bootc logic needed in our CI scripts.

- Add `systemtest/install-test-deps.sh` as a shell prepare step so TMT can wrap
  it into a Containerfile for bootc builds (replaces the inline `install` prepare
  which TMT cannot wrap)
- Refactor `systemtest/plans/main.fmf` to use `how: shell` prepare instead of
  `how: install` (required for TMT bootc builder compatibility)
- Add `is_bootc()` guard in `test.sh` to skip `dnf install` and `guest-setup.sh`
  on image-mode systems (deps are baked into the image)
- Enhance `guest-setup.sh` with copr fallback for environments where
  `insights-client` is not in the default repos
- Clean up error handling and settings download logic in `test.sh`

Verified on `RHEL-10.2-image-mode` compose via Testing Farm: prepare phase
completes cleanly, no reboot loops.

```
testing-farm request \
  --compose RHEL-10.2-image-mode \
  --git-url https://github.com/RedHatInsights/insights-client \
  --git-ref jmolet/imagemode \
  -e SETTINGS_URL=http://auto-services.usersys.redhat.com/automation-properties/settings.toml \
  -e ENV_FOR_DYNACONF='prod' \
```
Ref: CCT-2148

### DIFF
--- a/integration-tests/selinux.py
+++ b/integration-tests/selinux.py
@@ -148,21 +148,29 @@ class SELinuxAVCChecker:
         self.avc_skiplist.append(condition)
         return condition
 
+    _SEPARATOR = "=" * 63
+
     def get_avcs(self, skiplisted=True):
-        lines = (
-            subprocess.run(self.aureport_command, stdout=subprocess.PIPE)
-            .stdout.decode()
-            .splitlines()
-        )
-        assert lines.pop(0) == ""
-        assert lines.pop(0) == "AVC Report"
-        assert lines.pop(0) == "==============================================================="
-        keys = lines.pop(0).split()
-        assert lines.pop(0) == "==============================================================="
-        if lines[0] == "<no events of interest were found>":
+        output = subprocess.run(self.aureport_command, stdout=subprocess.PIPE).stdout.decode()
+        lines = [line for line in output.splitlines() if line.strip()]
+
+        if not lines or "<no events of interest were found>" in output:
             return
-        for line in lines:
-            if not line:  # skip empty lines
+
+        # Find the column-header line: it sits between two separator rows.
+        keys = None
+        data_start = None
+        for i, line in enumerate(lines):
+            if line.startswith("===") and i + 2 < len(lines) and lines[i + 2].startswith("==="):
+                keys = lines[i + 1].split()
+                data_start = i + 3
+                break
+
+        if keys is None or data_start is None:
+            return
+
+        for line in lines[data_start:]:
+            if line.startswith("===") or not line.strip():
                 continue
             entry = AuditLogEntry(keys, line.split())
             if skiplisted and any(condition(entry) for condition in self.avc_skiplist):
--- a/systemtest/tests/integration/test.sh
+++ b/systemtest/tests/integration/test.sh
@@ -1,37 +1,31 @@
 #!/bin/bash
-set -u
-set -x
+set -ux
 
 # get to project root
 cd ../../../
 
-# Check for bootc/image-mode deployments which should not run dnf
-if ! command -v bootc >/dev/null || bootc status | grep -q 'System is not deployed via bootc'; then
+is_bootc() {
+  command -v bootc > /dev/null &&
+    ! bootc status --format=humanreadable | grep -q 'System is not deployed via bootc'
+}
+
+if ! is_bootc; then
   # TEST_RPMS is set in jenkins jobs after parsing CI Messages in gating Jobs.
   # If TEST_RPMS is set then install the RPM builds for gating.
   if [[ -v TEST_RPMS ]]; then
     echo "Installing RPMs: ${TEST_RPMS}"
-    dnf -y install --allowerasing ${TEST_RPMS} || { echo "REQUESTED insights-client PACKAGE IS NOT INSTALLED"; exit 2; }
+    dnf -y install --allowerasing ${TEST_RPMS}
   fi
 
   # Simulate the packit setup on downstream builds.
   # This is for ad-hoc and compose testing.
-  rpm -q insights-client > /dev/null || ./systemtest/guest-setup.sh
-
-  # In most cases these should already be installed by tmt, see systemtest/plans/main.fmf
-  dnf --setopt install_weak_deps=False install -y \
-    podman git-core python3-pip python3-pytest logrotate bzip2 zip \
-    scap-security-guide openscap-scanner openscap bzip2-devel
+  rpm -q insights-client || ./systemtest/guest-setup.sh
 fi
 
-# If SETTINGS_URL is set (most likely in .testing-farm.yaml), download the settings
-# file from the provided URL. Back up any existing settings.toml before downloading.
-if [[ -v SETTINGS_URL ]]; then
+# Override settings if provided and available.
+if [ -n "${SETTINGS_URL+x}" ] && curl -I "$SETTINGS_URL" > /dev/null 2>&1; then
   [ -f ./settings.toml ] && mv ./settings.toml ./settings.toml.bak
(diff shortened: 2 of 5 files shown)


Lorquas:
## Image-mode Jenkins verification — before/after

Ran the `insights-client/imagemode` Jenkins job against both `main` and this branch on `RHEL-10.2-image-mode` (x86_64) to verify the changes.

| | **#99 — `main` (control)** | **#102 — this branch** |
|---|---|---|
| Test cases | 118 | 115 |
| Passed | 47 | **112** |
| Failed | 3 | 3 |
| **Errors** | **112** | **0** |
| Skipped | 3 | 2 |

The 112 errors on `main` are all from the `check_avcs` SELinux fixture crashing on teardown. The `aureport` output parser used rigid line-position assertions (`lines.pop(0) == "AVC Report"`) that broke on RHEL 10.2's changed `aureport` output format. This cascaded into every test's teardown, masking tests that would otherwise pass.

This branch eliminates all 112 errors by:
- Adding image-mode support via TMT's native bootc builder (`how: shell` prepare step)
- Rewriting the `aureport` parser in `selinux.py` to dynamically locate the separator-header-separator pattern instead of asserting exact line positions

Additionally verified the `aureport` fix on **RHEL 9.8** image-mode ([#103](https://jenkins-csb-rhsm-prod.dno.corp.redhat.com/job/insights-client/job/imagemode/103/) — SUCCESS) and **RHEL 10.2** ([#104](https://jenkins-csb-rhsm-prod.dno.corp.redhat.com/job/insights-client/job/imagemode/104/) — SUCCESS) with a single-test run.

The 3 remaining failures are pre-existing and unrelated to this PR:
- `test_output_file_with_relative_path` — path handling in Testing Farm workspace
- `test_pmlogger_running_and_metrics_exist` — pmlogger unavailable in bootc
- `test_selinux_core_context` — SELinux policy expectation mismatch

Note: test count differs slightly (118 vs 115) because `test_collection` has different parametrization between stage/prod environments.

**Jenkins links:**
- Control (`main`): [insights-client/imagemode #99](https://jenkins-csb-rhsm-prod.dno.corp.redhat.com/job/insights-client/job/imagemode/99/)
- This branch: [insights-client/imagemode #102](https://jenkins-csb-rhsm-prod.dno.corp.redhat.com/job/insights-client/job/imagemode/102/)
- RHEL 9.8 smoke: [insights-client/imagemode #103](https://jenkins-csb-rhsm-prod.dno.corp.redhat.com/job/insights-client/job/imagemode/103/)
- RHEL 10.2 smoke: [insights-client/imagemode #104](https://jenkins-csb-rhsm-prod.dno.corp.redhat.com/job/insights-client/job/imagemode/104/)

**Testing Farm artifacts:**
- Control: https://artifacts.osci.redhat.com/testing-farm/10525a20-5de1-4853-aa6b-a15a3bc7ecf2
- This branch: https://artifacts.osci.redhat.com/testing-farm/a872f158-ef05-4758-bf10-2e363cb92d9d

integration-tests/selinux.py
```diff
@@ -148,21 +148,29 @@ def skip_all_avcs(self):
...
+        keys = None
+        data_start = None
+        for i, line in enumerate(lines):
+            if line.startswith("===") and i + 2 < len(lines) and lines[i + 2].startswith("==="):
+                keys = lines[i + 1].split()
+                data_start = i + 3
+                break
+
+        if keys is None or data_start is None:
+            return
```
pholica, line 170: Thanks for checking this, however I have some concerns. Was Jenkins job 101 the one were you see issues with the `checker.get_avcs`? If so, the assert that fails is the second after the first empty line, so there is actually an empty line:
```
        # Strip leading blank lines (RHEL 10+ aureport omits the leading newline)
        while lines and lines[0] == "":
            lines.pop(0)
>       assert lines.pop(0) == "AVC Report"
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E       AssertionError
```
The reason why the code is so strict is to ensure that the capturing works correctly. With the new code proposed in this PR, it would silently continue if there was something else printed claiming that there was no AVC instead of ending with error (as the code is not compatible with the produced output and cannot reliably provide information about the AVCs).
This suspicion is also supported by the `test_selinux_core_context` failing as that test is expecting an AVC being raised.
Also, it doesn't sound like it's 10.2+ change as the AVC checking code behaves correctly on 10.2 systems that do not run in image mode.

Can you please provide sample of the `aureport` output that the code was hitting?

Thanks

pholica (changes requested):

systemtest/tests/integration/test.sh
```diff
@@ -1,37 +1,31 @@
...
+    ! bootc status --format=humanreadable | grep -q 'System is not deployed via bootc'
+}
+
+if ! is_bootc; then
   # TEST_RPMS is set in jenkins jobs after parsing CI Messages in gating Jobs.
   # If TEST_RPMS is set then install the RPM builds for gating.
   if [[ -v TEST_RPMS ]]; then
     echo "Installing RPMs: ${TEST_RPMS}"
-    dnf -y install --allowerasing ${TEST_RPMS} || { echo "REQUESTED insights-client PACKAGE IS NOT INSTALLED"; exit 2; }
+    dnf -y install --allowerasing ${TEST_RPMS}
```
zpetrace, line 17: The previous logic explicitly failed when TEST_RPMS install failed but as it was removed, failed installs no longer stop the test flow from what I understood. I would maybe want to have it there so we are not waiting for the test run to finish for an hour and half only to realize RPMs install failed at the beginning.

zpetrace (changes requested):

systemtest/tests/integration/test.sh
```diff
@@ -40,15 +34,14 @@ python3 -m venv venv
 
 pip install -r integration-tests/requirements.txt
 
-# Print versions of packages that are actively being tested
-rpm -q insights-client insights-core selinux-policy || true
-
```
m-horky, line 45: I don't think this should have been removed.

m-horky (commented):
Thanks for the PR. It seems it is reverting a lot of "downstream" changes we have made to make testing work reliably.

Given the context of CCT-2060, and specific bugs like RHEL-135034, I'll be closing it for now. We'll be handling it on our own; at runtime, insights-client should behave the same as under package mode, introducing bootc runtime testing increases maintenance scope while not addressing the real pain points. The "green CI results" in this case isn't a reliable metric (again, see CCT-2060).

integration-tests/selinux.py
Lorquas: @pholica good point about the error gobbling.  I implemented this as a  workaround for a issue with imagemode breaking the output of aurreport and I thought it was a line spacing issue but trying to reproduce it again the issue seems deeper.  I will remove this change from the PR and show you my findings outside of this.

Lorquas:
@m-horky @zpetrace I've improved this PR with the requested changes.  I've moved the aureport stuff out as i think that now warrants a separate investigation.  The other suggestions were me getting too overzealous with the delete button and I just never pushed the fixes back before review.   I know this PR is closed, but I will leave the branch around as a reference for when you want to get image-mode runtime tests running.

PR #658, opened 2026-03-30
fix: Remove .etag files during upgrade process

pkoprda:
* Card ID: RHEL-145160

Since insights-core is already shipped as an RPM, .etag files (which were created by insights-core egg) should not be present in /etc/insights-client.

---

- `rhel-9-main` (RHEL >= 9.8)


m-horky (approved):
ACK

PR #659, opened 2026-03-30
[rhel-9-main] fix: Remove .etag files during upgrade process

pkoprda:
* Card ID: RHEL-145158

Since insights-core is already shipped as an RPM, .etag files (which were created by insights-core egg) should not be present in /etc/insights-client.

(cherry picked from commit b8e138d32d52d9bfe20f14fc496a0378d5d66298)

---

This pull request is a backport of: #658


m-horky (approved):
ACK

PR #660, opened 2026-04-09
fix(test): report universal checks as FAIL

Peto-RH:
* CardID: CCT-1999

Pytest reports fixture setup and teardown failures as ERROR by default.
Add a pytest hook that detects pytest.fail() raised from universal check
fixtures and reports them as test FAIL instead.

Universal check fixtures are identified by fixture function names
starting with check_.

---

- `rhel-9-main` (RHEL >= 9.8)
- `rhel-9-egg` (RHEL <= 9.7)

### DIFF
--- a/integration-tests/conftest.py
+++ b/integration-tests/conftest.py
@@ -12,6 +12,25 @@ from pytest_client_tools.util import loop_until
 logger = logging.getLogger(__name__)
 
 
+@pytest.hookimpl(hookwrapper=True)
+def pytest_runtest_makereport(item, call):
+    """
+    Reclassify check_* fixture "teardown"/"setup" failures as FAIL (instead of ERROR) by
+    setting the test execution phase to "call" instead of "setup"/"teardown".
+    """
+    outcome = yield
+    rep = outcome.get_result()
+    if call.when == "call" or not call.excinfo:
+        return
+    if not isinstance(call.excinfo.value, pytest.fail.Exception):
+        return
+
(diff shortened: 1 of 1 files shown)


Peto-RH:
Note for the reviewer: Last 2 commits will be removed. They are there just to showcase that it reports FAIL

zpetrace:
I think overall the PR looks good, removal of the last 2 commits should also resolve the black&flake8 check that is failing.

integration-tests/conftest.py
```diff
@@ -12,6 +12,24 @@
 logger = logging.getLogger(__name__)
 
 
+@pytest.hookimpl(hookwrapper=True)
+def pytest_runtest_makereport(item, call):
+    outcome = yield
+    rep = outcome.get_result()
+
+    # Reclassify check_* fixture teardown/setup failures as FAIL (instead of ERROR) by
+    # setting the test execution phase to "call" instead of "setup"/"teardown".
```
pkoprda: Can be this comment be part of docstring for this method?

pkoprda (changes requested):
Thanks for the PR, can be this also backported to `-egg` branches?

Peto-RH:
> Thanks for the PR, can be this also backported to `-egg` branches?

Yes, good catch. It it is only not relevant for the rhel-8-egg branch.

pkoprda (approved):
Thanks, approved

PR #666, opened 2026-05-25
[rhel-10-egg] fix(test): Printing EGG version before tests

zpetrace:
As we do not need insights-core version on RHEL version where EGGs are being used I have removed the insights-core from the rpm command and added insights-client version printing to show the egg version. This change will help mainly with testing the EGG from insights-core team.

---

- `rhel-9-egg` (RHEL <= 9.7)
- `rhel-8-egg` (RHEL 8)


pkoprda (approved):
ACK

PR #672, opened 2026-07-13
fix(test): Handle aarch64 AVC syscall

Peto-RH:
The known CCT-2009 getattr denial is reported as fstat on x86_64 but as newfstat on aarch64. Skip both syscall variants for the same known SELinux AVC pattern.

* Card ID: CCT-2555

---

- `rhel-9-main` (RHEL >= 9.8)


zpetrace (changes requested):
The change looks good, please just fix the black & flake8 and let's wait for the test results after that.

Peto-RH:
The same test fails also in this PR
https://github.com/RedHatInsights/insights-client/pull/669/checks

zpetrace (approved):
As the same test fails on [#669](https://github.com/RedHatInsights/insights-client/pull/669) it points to an existing RHEL 10 Nightly integration-test/SELinux issue, not a regression caused by this PR. This LGTM.

PR #678, opened 2026-07-20
[rhel-8-egg] fix: Update build process to match new Core schema

pkoprda:
- .fallback.json needs to be removed from the meson.build file and .spec file
- uploader.v2.json is no longer necessary, legacy collection has been removed from insights-core-3.5.1 up.

---

This pull request is partially backport of: #335


scripts/01-upgrade-egg.sh
```diff
@@ -18,7 +18,5 @@ done
 
 TAG=$(cat EGG_VERSION)
```
stephenbourke: Will the script not fail with the non-zero exit (set -e) considering EGG_VERSION has been removed?

```
TAG=$(cat EGG_VERSION)
cat: EGG_VERSION: No such file or directory
echo $?
1
```

pkoprda: Yeah, I forgot about this, fixed, thanks!

pkoprda: I also updated `scripts/README.md` file

stephenbourke (approved):
LGTM

PR #681, opened 2026-07-22
[rhel-10-egg] chore: remove unused EGG_VERSION reference from egg upgrade script

pkoprda:
The `EGG_VERSION` file is no longer used; the egg is downloaded directly from console.redhat.com. Update README accordingly.


zpetrace (approved):
LGTM, thanks!

PR #682, opened 2026-07-22
[rhel-9-egg] docs: update egg download description in scripts README

pkoprda:
Egg upgrade script doesn't use `EGG_VERSION` file and `uploader.json` anymore, README file should be updated accordingly.


zpetrace (approved):
LGTM!

PR #683, opened 2026-07-23
[rhel-8-egg] chore: Drop ros subpackage from spec file

pkoprda:
The insights-client-ros subpackage is not available on RHEL 8, so it doesn't make sense to have the ros subpackage definition, post/postun scripts, and files section in the spec file.


zpetrace (approved):
LGTM, makes sense, thanks!

PR #684, opened 2026-07-27
[rhel-10-egg] test: assert --version core matches runtime egg

zpetrace:
Add a regression test for RHEL-123603 so --version cannot report an older rpm/stable core while other commands use newest.egg.

* Card ID: RHEL-123603

---

- `rhel-9-egg` (RHEL <= 9.7)
- `rhel-8-egg` (RHEL 8)

### DIFF
--- a/integration-tests/test_version.py
+++ b/integration-tests/test_version.py
@@ -6,7 +6,11 @@
 :upstream: Yes
 """
 
+import re
 import pytest
+
+from pytest_client_tools.util import Version
+from constants import INSIGHTS_CLIENT_LOG_FILE
 
 
 @pytest.mark.tier1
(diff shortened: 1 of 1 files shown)


integration-tests/test_version.py
```diff
@@ -28,3 +32,36 @@ def test_version(insights_client):
...
+@pytest.mark.usefixtures("register_subman")
+@pytest.mark.tier1
+def test_version_core_matches_runtime_egg(insights_client):
+    """
+    :id: 5fa66d8b-0304-450d-9ad7-c3170fe4bd07
+    :title: Test --version Core matches the egg used by other commands
+    :description:
+        This test verifies that `insights-client --version` reports the same
+        Core version as the egg used by other commands such as `--status`,
+        including when `/var/lib/insights/newest.egg` is present.
```
pkoprda: > such as `--status`

This test uses just `--status`. But what about other commands? Can we add at least `--checkin`?

zpetrace: Sure, added it as an option.

pkoprda (approved):
LGTM

PR #685, opened 2026-07-28
[rhel-9-egg] test: assert --version Core matches runtime egg

zpetrace:
Add a regression test for RHEL-123603 so --version cannot report an older rpm/stable core while other commands use newest.egg.

(cherry picked from commit 88dd770456bddeb0376fd37b8d84f9b3c85e573d)

---


pkoprda (approved):
LGTM

PR #687, opened 2026-07-28
[rhel-10-egg] chore: clean up logrotate file after uninstallation

pkoprda:
During uninstallation process of insights-client in downstream, the logrotate file for insights-client is removed. Upstream spec file should do the same thing.

---

- `rhel-9-egg` (RHEL <= 9.7)
- `rhel-8-egg` (RHEL 8)


zpetrace (approved):
LGTM

PR #691, opened 2026-07-29
fix: Mark runtime directories as %ghost in spec file

pkoprda:
* Card ID: RHEL-140570
* Card ID: RHEL-187922

Directories managed by tmpfiles.d are marked as %ghost so RPM does not warn about missing files when the package is removed after the directories have already been cleaned up.

---

- `rhel-9-main` (RHEL >= 9.8)

### DIFF
--- a/insights-client.spec
+++ b/insights-client.spec
@@ -79,6 +79,8 @@ make install-files \
 %systemd_post %{name}.timer
 %systemd_post %{name}-boot.service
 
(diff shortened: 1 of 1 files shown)


pkoprda:
Failing test `test_register_with_ros` is a known issue in insights-core, see RHEL-180919.

stephenbourke (approved):
LGTM

PR #692, opened 2026-08-04
[rhel-9-main] fix: Mark runtime directories as %ghost in spec file

pkoprda:
* Card ID: RHEL-135034

Directories managed by tmpfiles.d are marked as %ghost so RPM does not warn about missing files when the package is removed after the directories have already been cleaned up.

(cherry picked from commit cae148616bb647f11547b4b32c026305d70ac19b)

---

This pull request is a backport of: #691


stephenbourke (approved):
LGTM. Failing tests have nothing to do with this change.

